PR #431, opened 2026-03-14
perf: producer hot path optimizations

### LINKED ISSUE #404: perf: ReserveMemorySync uses 10ms polling for backpressure wake-up
## Problem

`ReserveMemorySync` in `RecordAccumulator.cs` uses `WaitHandle.WaitAny` with a **10ms timeout** in a polling loop to wait for buffer space. When a batch completes and frees buffer memory, `_bufferSpaceAvailable` is signaled, but the producer thread may sleep up to 10ms after space becomes available due to the polling interval.

```csharp
// Current implementation (RecordAccumulator.cs:1043-1044)
var waitMs = Math.Min(10, (int)remainingMs);
var signaled = WaitHandle.WaitAny(_syncWaitHandles, waitMs);
```

## Impact

- Up to 10ms of unnecessary latency per backpressure event
- Under sustained load at millions of messages/sec, this polling overhead adds up
- `ReserveMemorySync` shows as **3.55% inclusive CPU** in profiling, with `WaitHandle.WaitMultiple` at 3.54% exclusive

## Suggested approaches

1. **Increase WaitAny timeout** to a larger value (e.g., 100ms or `Timeout.Infinite`) since the `ManualResetEventSlim` signal provides the actual wake-up. The 10ms is a safety fallback, not the primary wake-up mechanism.

2. **Replace with SemaphoreSlim** — use a counting semaphore where `ReleaseMemory` releases a permit and `ReserveMemorySync` acquires one, providing exact-count wake-up without polling.

3. **Use ManualResetEventSlim.Wait directly** instead of `WaitHandle.WaitAny` — MRES has a more efficient spin-then-kernel-wait strategy and doesn't need the WaitHandle array allocation pattern.

## Profiling context

Captured with `dotnet-trace collect --profile cpu-sampling` during a 2-minute producer stress test. The async variant (`ReserveMemoryAsync`) uses a similar 5ms `Task.Delay` polling pattern.

### LINKED ISSUE #426: perf: ReserveMemorySync thundering herd on backpressure wakeup
## Problem

`RecordAccumulator.ReserveMemorySync` accounts for **3.18% inclusive CPU time** in stress test traces, with `ManualResetEventSlim.Wait` at **3.15% exclusive**.

When `ReleaseMemory` calls `_bufferSpaceAvailable.Set()`, **all** waiting threads wake up and CAS-race in `TryReserveMemory`. Only one succeeds; the rest loop back to `Wait`. This creates a thundering herd pattern under backpressure.

## Trace Evidence

- 2-minute producer stress test, fire-and-forget, 141K msg/sec
- `ManualResetEventSlim.Wait` = 3.15% exclusive CPU time (rank #6 overall)
- `RecordAccumulator.ReserveMemorySync` = 3.18% inclusive

## Suggested Approach

Replace `ManualResetEventSlim` (broadcast wake) with a mechanism that wakes a **single** waiter:
- `SemaphoreSlim` — wakes one waiter per `Release()` call
- `AutoResetEvent` — auto-resets after waking one thread

This would eliminate the CAS contention storm when many threads are waiting on buffer space.

## Location

`src/Dekaf/Producer/RecordAccumulator.cs` — `ReserveMemorySync`, `ReleaseMemory`, `TryReserveMemory`

### LINKED ISSUE #427: perf: Reduce per-partition lock contention in RecordAccumulator.Append
## Problem

`Monitor.Enter_Slowpath` accounts for **0.89% exclusive CPU time** in stress test traces. The per-partition `PartitionDeque.Lock` is held across slow operations during `Append()`.

## Trace Evidence

- 2-minute producer stress test, fire-and-forget, 141K msg/sec
- `Monitor.Enter_Slowpath` = 0.89% exclusive CPU time (rank #8 overall)

## Root Cause

The lock in `Append` / `TryAppendWithCompletion` is held while:
1. Appending to batch arrays (`TryAppendToBatch`)
2. Growing arrays (`GrowArray` → `Array.Copy`)
3. Creating new batches (`RentBatch` → pool allocation)
4. Sealing batches (`SealCurrentBatchUnderLock`)

When multiple producer threads target the same partition, they contend on this lock.

## Suggested Approach

- Move array growth (`GrowArray`) outside the lock — pre-allocate or double-buffer
- Reduce critical section to just the pointer swap (append index increment)
- Consider lock-free append for the common case (batch has capacity)

## Location

`src/Dekaf/Producer/RecordAccumulator.cs` — `Append`, `TryAppendWithCompletion`, `PartitionDeque.Lock`

### LINKED ISSUE #428: perf: Reduce per-message allocations in fire-and-forget producer path (~750 bytes/msg)
## Problem

The producer allocates approximately **750 bytes per message** in the fire-and-forget path, resulting in **7.42 GB total allocated** over a 2-minute stress test (141K msg/sec). This drives **612 Gen0 and 612 Gen1 GC collections** (~5.1/sec each) plus 65 Gen2 collections.

## Trace Evidence

- 7.42 GB allocated / ~17.2M messages = ~750 bytes/message
- Gen0 = Gen1 = 612 collections (unusual — suggests most Gen0 survivors promote immediately)
- Gen2 = 65 collections in 2 minutes
- GC heap steady at ~1.1 GB, working set ~1.23 GB
- `GC.RunFinalizers` = 0.82% exclusive CPU time

## Root Cause

`AppendWithSlowPath` rents per-message arrays from `ArrayPool<byte>.Shared`:
- `ArrayPool<byte>.Shared.Rent(keyLength)` — per message
- `ArrayPool<byte>.Shared.Rent(valueLength)` — per message
- Header arrays — per message (if headers present)

At 141K msg/sec, that's 282K+ ArrayPool rentals/sec just for key/value data.

## Suggested Approach

- **Arena allocation**: Append key/value bytes directly into a batch-level contiguous buffer (arena) instead of renting individual arrays per message. Track offsets rather than individual array references.
- **Batch-level memory**: Allocate a single large buffer per batch and sub-allocate from it for each message's key/value data.
- This would reduce per-message allocations to near zero and collapse the cleanup loop from O(n) array returns to O(1) arena return.

## Location

- `src/Dekaf/Producer/KafkaProducer.cs` — `AppendWithSlowPath` (lines ~890-946)
- `src/Dekaf/Producer/RecordAccumulator.cs` — batch array management, `GrowArray`

### LINKED ISSUE #429: perf: ReadyBatch.Cleanup loops through ~1000 arrays per batch with clearArray overhead
## Problem

`ReadyBatch.Cleanup` accounts for **0.33% inclusive CPU time**. It loops through ~1000 pooled arrays per batch, returning each to `ArrayPool` with `clearArray: true`, which zeros each buffer before returning.

## Trace Evidence

- 2-minute producer stress test, fire-and-forget, 141K msg/sec
- `ReadyBatch.Cleanup` = 0.33% inclusive CPU time
- Called from `CompleteSend` on the sender thread, blocking subsequent sends

## Root Cause

```csharp
for (var i = 0; i < _pooledDataArraysCount; i++)
    ArrayPool<byte>.Shared.Return(_pooledDataArrays[i], clearArray: true);
```

At ~1000 messages per batch and ~5000 batches per minute:
- ~5M individual `ArrayPool.Return` calls per minute
- Each call zeros the buffer before returning (`clearArray: true`)
- Runs synchronously on the sender thread

## Suggested Approach

1. **Use `clearArray: false`** for byte arrays if zeroing isn't required for security
2. **Batch the returns** — return array-of-arrays in one operation if possible
3. **Offload cleanup** to a background thread or queue so it doesn't block the sender
4. **Long-term**: Eliminate per-message arrays entirely (see arena allocation issue) — this would make the cleanup loop O(1) instead of O(n)

## Location

`src/Dekaf/Producer/RecordAccumulator.cs` — `ReadyBatch.Cleanup`

### LINKED ISSUE #430: perf: ReadyBatch.CompleteSend per-message work on sender thread
## Problem

`ReadyBatch.CompleteSend` accounts for **0.33% inclusive CPU time**. It iterates through per-message completion sources and callbacks, then triggers `Cleanup()`, all on the sender thread.

## Trace Evidence

- 2-minute producer stress test, fire-and-forget, 141K msg/sec
- `ReadyBatch.CompleteSend` = 0.33% inclusive CPU time
- Combined with `Cleanup` (also 0.33%), the batch completion path is ~0.66% of CPU

## Root Cause

- Creates `RecordMetadata` per message for completion sources and callbacks
- Invokes user callbacks inline on the sender thread
- Calls `Cleanup()` synchronously in the `finally` block

In fire-and-forget mode (no completion sources or callbacks), verify that the loops short-circuit early. If `_completionSourcesCount == 0 && _callbackCount == 0`, the method should skip directly to cleanup.

## Suggested Approach

1. **Fast-path for fire-and-forget**: Skip completion source and callback loops entirely when counts are zero
2. **Offload user callbacks**: Schedule callbacks on the thread pool instead of running inline on the sender thread
3. **Defer Cleanup**: Queue cleanup work instead of blocking the sender thread

## Location

`src/Dekaf/Producer/RecordAccumulator.cs` — `ReadyBatch.CompleteSend`, `ReadyBatch.Cleanup`

thomhurst:
## Summary

Addresses 6 open performance issues identified from stress test profiling (141K msg/sec, 2-minute fire-and-forget):

- **#430 + #429**: Skip completion source/callback loops in fire-and-forget mode (CompleteSend fast path) and use `clearArray: false` in Cleanup to avoid unnecessary zeroing of byte arrays
- **#426 + #404**: Replace `ManualResetEventSlim` with `AutoResetEvent` for sync backpressure signaling, eliminating thundering herd where all waiters wake but only one succeeds the CAS
- **#428**: Arena-based zero-copy in the fire-and-forget slow path — serialize directly into batch arena instead of renting per-message ArrayPool arrays (~750 bytes/msg reduction)
- **#427**: Replace `Monitor` lock with `SpinLock` for per-partition deque access, avoiding kernel transitions for brief critical sections

Also fixes a correctness bug found during code review: `valueLength == 0` was incorrectly used as `valueIsNull`, causing empty non-null values to be stored as null in Kafka records.

## Test plan

- [x] All 3027 unit tests pass
- [ ] Integration tests (requires Docker)
- [ ] Stress test to verify allocation reduction and throughput improvement
- [ ] Verify empty-value messages are correctly distinguished from null-value messages

Closes #427, closes #428, closes #429, closes #430, closes #426, closes #404

### DIFF
--- a/src/Dekaf/Producer/RecordAccumulator.cs
+++ b/src/Dekaf/Producer/RecordAccumulator.cs
@@ -15,6 +15,30 @@ using Microsoft.Extensions.Logging;
 namespace Dekaf.Producer;
 
 /// <summary>
+/// RAII guard for SpinLock that ensures Exit() is called on dispose.
+/// Must be used with <c>using var guard = new SpinLockGuard(ref spinLock);</c>.
+/// </summary>
+internal ref struct SpinLockGuard
+{
+    private ref SpinLock _lock;
+    private bool _taken;
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public SpinLockGuard(ref SpinLock spinLock)
+    {
+        _lock = ref spinLock;
+        _taken = false;
+        _lock.Enter(ref _taken);
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public void Dispose()
+    {
+        if (_taken) _lock.Exit();
+    }
+}
+
+/// <summary>
 /// Debug-only tracking for message flow through the producer pipeline.
 /// Tracks messages at each stage to identify where messages are lost.
 /// All recording methods use [Conditional("DEBUG")] so calls are stripped in Release builds.
@@ -814,15 +838,9 @@ public sealed partial class RecordAccumulator : IAsyncDisposable
                 continue;
 
             ReadyBatch? head;
-            bool lockTaken = false;
-            try
             {
-                pd.Lock.Enter(ref lockTaken);
+                using var guard = new SpinLockGuard(ref pd.Lock);
                 head = pd.PeekFirst();
-            }
-            finally
-            {
-                if (lockTaken) pd.Lock.Exit();
             }
 
             if (head is null)
@@ -932,10 +950,8 @@ public sealed partial class RecordAccumulator : IAsyncDisposable
                 continue;
 
             ReadyBatch? batch;
-            bool lockTaken = false;
-            try
             {
-                pd.Lock.Enter(ref lockTaken);
+                using var guard = new SpinLockGuard(ref pd.Lock);
                 batch = pd.PeekFirst();
                 if (batch is null)
                     continue;
@@ -948,10 +964,6 @@ public sealed partial class RecordAccumulator : IAsyncDisposable
                     break;
 
                 batch = pd.PollFirst();
-            }
-            finally
-            {
-                if (lockTaken) pd.Lock.Exit();
             }
 
             if (batch is not null)
@@ -972,18 +984,12 @@ public sealed partial class RecordAccumulator : IAsyncDisposable
     {
         batch.Reenqueued(nowMs);
         var pd = GetOrCreateDeque(batch.TopicPartition);
-        bool lockTaken = false;
-        try
         {
-            pd.Lock.Enter(ref lockTaken);
+            using var guard = new SpinLockGuard(ref pd.Lock);
             if (ProducerId >= 0)
                 pd.InsertInSequenceOrder(batch);
             else
                 pd.AddFirst(batch);
-        }
-        finally
-        {
-            if (lockTaken) pd.Lock.Exit();
         }
         SignalWakeup();
     }
@@ -1040,20 +1046,14 @@ public sealed partial class RecordAccumulator : IAsyncDisposable
         foreach (var kvp in _partitionDeques)
         {
             var pd = kvp.Value;
-            bool lockTaken = false;
-            try
             {
-                pd.Lock.Enter(ref lockTaken);
+                using var guard = new SpinLockGuard(ref pd.Lock);
                 var b = pd.PollFirst();
                 if (b is not null)
                 {
                     batch = b;
                     return true;
                 }
-            }
-            finally
-            {
-                if (lockTaken) pd.Lock.Exit();
             }
         }
 
@@ -1234,10 +1234,8 @@ public sealed partial class RecordAccumulator : IAsyncDisposable
         var pd = GetOrCreateDeque(topicPartition);
         bool batchSealed = false;
 
-        bool lockTaken = false;
-        try
         {
-            pd.Lock.Enter(ref lockTaken);
+            using var guard = new SpinLockGuard(ref pd.Lock);
 
             // Check disposal under lock
             if (_disposed)
@@ -1276,10 +1274,6 @@ public sealed partial class RecordAccumulator : IAsyncDisposable
                 throw new KafkaException(ErrorCode.MessageTooLarge,
                     $"Record of size {recordSize} exceeds maximum batch size of {_options.BatchSize}");
             }
-        }
-        finally
-        {
-            if (lockTaken) pd.Lock.Exit();
         }
 
         if (batchSealed)
@@ -1321,10 +1315,8 @@ public sealed partial class RecordAccumulator : IAsyncDisposable
         var pd = GetOrCreateDeque(topicPartition);
         bool batchSealed = false;
 
-        bool lockTaken = false;
-        try
         {
-            pd.Lock.Enter(ref lockTaken);
+            using var guard = new SpinLockGuard(ref pd.Lock);
 
             if (_disposed)
             {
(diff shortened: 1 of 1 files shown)


thomhurst:
Addressed in 2743d09:

**Concern 1 (AutoResetEvent under-signaling):** Fixed. After a successful `TryReserveMemory` in the wait loop, we now chain-wake the next waiter by calling `_bufferSpaceAvailable.Set()` if space remains below the limit. This propagates wakeups when a single large batch frees enough memory for multiple waiting threads.

**Concern 2 (_syncWaitHandles race):** Fixed. Moved initialization to the constructor as a `readonly` field. No more lazy `??=` race.

**Concern 3 (clearArray: false security):** No change. The codebase already uses `clearArray: false` in 8+ other ArrayPool returns (PooledMemory.Return, BatchArena, GrowArray, all other Cleanup arrays). This change makes the last holdout consistent. Users sending sensitive data should use encryption at the transport layer rather than relying on array clearing in pooled memory.

**Concern 4 (SpinLock memory barrier on ARM):** Fixed. Changed all `Exit(useMemoryBarrier: false)` to `Exit()` (default = `true`). Correctness on weak memory model architectures is worth the minor cost.

**Concern 5 (TryAppendFromSpans duplication):** Acknowledged but not changing. `ReadOnlySpan<byte>` cannot be captured in delegates, stored in classes, or passed through async boundaries — this is a fundamental C# constraint that prevents unifying the two paths. A shared `CommitRecord` helper could reduce ~30 lines of the suffix, but the arena-vs-pool allocation logic in the middle is inherently different. The duplication is a pragmatic trade-off for zero-copy performance.

PR #459, opened 2026-03-14
fix: replace blocking FetchTopicMetadataSync with async path

thomhurst:
## Summary

- **`Send()` no longer blocks on metadata fetch for new topics.** Previously, `Send()` called `FetchTopicMetadataSync` which used `.GetAwaiter().GetResult()` to block the calling thread while fetching metadata. This could hang indefinitely if the broker was slow to respond (e.g., auto-creating a topic).
- **Delegates to the async produce path on cache miss.** Uses `ProduceInternalAsync` (the same path `ProduceAsync` uses) which handles metadata fetching without blocking.
- **Proper resource lifecycle.** Uses existing `PooledValueTaskSource.ObserveForFireAndForget()` and `SetDeliveryHandler()` APIs to ensure completion sources are returned to the pool and delivery handlers are invoked.
- **Removes `FetchTopicMetadataSync` entirely** — no more `Task.Factory.StartNew` with `LongRunning` + `.GetAwaiter().GetResult()`.

## Test plan

- [x] All 3034 unit tests pass
- [ ] Integration tests (CI)
- [ ] Stress test warmup completes in <1s (verified locally: ~780ms)

### DIFF
--- a/src/Dekaf/Producer/KafkaProducer.cs
+++ b/src/Dekaf/Producer/KafkaProducer.cs
@@ -628,32 +628,10 @@ public sealed partial class KafkaProducer<TKey, TValue> : IKafkaProducer<TKey, T
                 return;
             }
 
-            // Topic cache miss — fetch topic metadata synchronously and produce.
-            // This is a one-time cost per new topic.
-            try
-            {
-                var topicInfo = FetchTopicMetadataSync(message.Topic);
-                ProduceSyncCoreFireAndForgetDirect(
-                    message.Topic,
-                    message.Key,
-                    message.Value,
-                    message.Partition,
-                    message.Timestamp,
-                    message.Headers,
-                    topicInfo);
-                activity?.SetStatus(ActivityStatusCode.Ok);
-            }
-            catch (KafkaTimeoutException)
-            {
-                // BufferMemory backpressure or metadata timeout must propagate
-                throw;
-            }
-            catch (Exception ex)
-            {
-                if (activity is not null)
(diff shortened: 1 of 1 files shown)


thomhurst:
## Response to Review

### KafkaTimeoutException propagation

The old code caught `KafkaTimeoutException` from `ProduceSyncCoreFireAndForgetDirect` (BufferMemory backpressure during accumulator append), not from `FetchTopicMetadataSync`. The metadata fetch path threw `ProduceException` on timeout.

The fast path (99%+ of calls, cached metadata) still propagates `KafkaTimeoutException` synchronously from the accumulator append. The cache-miss async path can't propagate back to the `Send()` caller by design — that's the tradeoff of making it non-blocking. Checking buffer availability synchronously before handing off would reintroduce the blocking we're removing.

In practice, `KafkaTimeoutException` from BufferMemory is extremely unlikely on the very first produce to a new topic (the buffer is empty). For the rare case where it does happen, the exception routes through `ObserveForFireAndForget` which swallows it — acceptable for fire-and-forget semantics.

### Send(message, deliveryHandler) semantic change

Agreed this is a behavioral change. The new behavior (routing errors through the delivery handler) is more consistent with fire-and-forget semantics and is the better contract.

### Activity/tracing on cache-miss path

Pre-existing limitation. The activity span covers "message accepted into client buffer" — for the cache-miss async path, the span ends before the metadata fetch completes. This is a known tradeoff documented in the `finally` block comment.

PR #505, opened 2026-03-15
refactor: remove custom statistics callback in favor of OTel metrics

thomhurst:
## Summary

- Remove the custom `StatisticsHandler`/`StatisticsInterval` callback feature entirely
- Delete the `Statistics/` directory (7 files), all builder methods, options properties, collector call sites, and associated tests
- **-3,013 lines** across 26 files

## Why

The custom statistics feature duplicated what `System.Diagnostics.Metrics` (`DekafMetrics.cs`) already provides via standard OTel instrumentation. It was a Confluent.Kafka-ism — Confluent needs it because it's a C wrapper that can't integrate with .NET's metrics infrastructure. Dekaf, being pure .NET, has proper OTel metrics via `DekafMetrics` (semantic conventions: `messaging.client.sent.messages`, `messaging.client.consumed.messages`, etc.).

The custom feature used a timer-based polling emitter that was inherently flaky on slow CI runners — the `Metrics_UnderHighThroughput_StayAccurate` test had **5 prior fix attempts** and still fails intermittently.

## What's preserved

`DekafMetrics.cs` and `DekafDiagnostics.cs` are **completely untouched** — they use static `System.Diagnostics.Metrics` counters recorded directly at call sites, independent of the removed collectors. Users get the same observability through standard OTel:

```csharp
builder.Services.AddOpenTelemetry()
    .WithMetrics(m => m.AddDekafInstrumentation());
```

## Test plan

- [x] All 2,995 unit tests pass
- [x] Library, integration tests, and stress tests build clean
- [ ] CI passes

### DIFF
--- a/src/Dekaf/Producer/KafkaProducer.cs
+++ b/src/Dekaf/Producer/KafkaProducer.cs
@@ -2792,16 +2792,19 @@ public sealed partial class KafkaProducer<TKey, TValue> : IKafkaProducer<TKey, T
         _disposed = true;
         LogProducerDisposing();
 
-        // Graceful shutdown: flush all pending messages to Kafka before closing.
-        // Use half of CloseTimeoutMs for the graceful flush phase, reserving the rest
-        // for forceful cleanup (BrokerSender cancellation, sender task teardown, etc.).
-        // Previously the entire CloseTimeoutMs was consumed by the flush, leaving no
-        // budget for the forceful path and causing 30s+ dispose times under load.
+        // Time budget allocation within CloseTimeoutMs:
+        //   40% → graceful flush (accumulator close + sender drain)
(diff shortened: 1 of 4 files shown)


thomhurst:
Verified the OTel metric parity gap. Current `DekafMetrics.cs` is missing:

1. **Rebalance duration** (`messaging.consumer.rebalance.duration` histogram)
2. **Fetch latency** (`messaging.consumer.fetch.duration` histogram)  
3. **Consumer lag** (`messaging.consumer.lag` observable gauge)

These are worth adding as OTel instruments in a follow-up PR — the data sources are still in `KafkaConsumer` (rebalance timing, fetch timing, `_highWatermarks` dictionary). Not a blocker for this removal since the custom stats callback was never the recommended path for production monitoring.

PR #564, opened 2026-03-23
perf: pool cancelled SyncWaiterNode objects to reduce allocation churn

### LINKED ISSUE #554: Pool SyncWaiterNode objects in ReserveMemorySync slow path
## Context

`RecordAccumulator.ReserveMemorySync` allocates a fresh `SyncWaiterNode` + `ManualResetEventSlim` on every slow-path entry (when buffer memory is full). The comment notes: "Nodes are NOT pooled — a fresh node is allocated per slow-path entry to avoid a race where a recycled node is signaled for the wrong thread."

## Problem

Under sustained buffer pressure (producer faster than network drain), this creates allocation churn. CPU profiling showed `ReserveMemorySync` at 5.77% inclusive CPU time during a stress test, with `ManualResetEventSlim.Wait` at 5.74%.

## Proposed solution

Pool `SyncWaiterNode` objects using a `ConcurrentBag` or thread-local cache. On rental:
1. Reset `ManualResetEventSlim` state via `Reset()`
2. Clear `Cancelled` flag

The race condition cited in the comment (recycled node signaled for wrong thread) should be safe if the node is only returned to the pool after the owning thread sets `Cancelled = true` and the node has been dequeued by `WakeNextSyncWaiter`.

## Origin

Identified during CPU profiling of producer stress test (PR #551 analysis).

thomhurst:
## Summary
- Recycle `SyncWaiterNode` instances in `WakeNextSyncWaiter` when nodes are cancelled, returning them to a `ConcurrentBag` pool instead of discarding
- On slow-path entry in `ReserveMemorySync`, rent from pool (resetting `ManualResetEventSlim` and `Cancelled` flag) before allocating new
- Reduces allocation churn under sustained buffer pressure where CPU profiling showed `ReserveMemorySync` at 5.77% inclusive time

Closes #554

## Test plan
- [x] New `SyncWaiterNodePoolTests` covering pool growth, reset-on-rental, independent node lifecycle, and backpressure cycling
- [x] Existing unit tests pass
- [x] Code reviewed via /simplify

### DIFF
--- a/src/Dekaf/Producer/RecordAccumulator.cs
+++ b/src/Dekaf/Producer/RecordAccumulator.cs
@@ -2380,10 +2380,12 @@ public sealed partial class RecordAccumulator : IAsyncDisposable
     /// <summary>
     /// Rents a <see cref="SyncWaiterNode"/> from the pool, or allocates a new one if the pool is empty.
     /// </summary>
+    /// <remarks>
     /// Reset-on-return is safe here because nodes are only returned after being fully
     /// dequeued from <c>_syncWaiterQueue</c> — no other thread holds a reference.
     /// This differs from <c>PartitionBatch</c> which requires reset-on-rent because
     /// multiple code paths may inspect a batch between return and next rental.
+    /// </remarks>
     private SyncWaiterNode RentWaiterNode()
     {
         if (_syncWaiterNodePool.TryDequeue(out var node))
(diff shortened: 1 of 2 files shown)


thomhurst:
Addressed in bdeb174:

**Advisory bound comment:** Added. The non-atomic check-then-enqueue is intentional — a CAS loop would add contention on a path that only runs during backpressure, and the transient overshoot is bounded by concurrency (typically single-digit).

**XML doc `<remarks>` tag:** Fixed. The reset-on-return explanation is now inside `<remarks>` so it renders in IDE tooltips.

**Test race:** Both pool-count assertions now use `WaitForCondition` with a 2-second deadline instead of asserting immediately. This handles the window where the worker thread has completed `reserveTask` but hasn't yet returned the node to the pool.

PR #642, opened 2026-03-27
perf: use 2 connections per broker in consumer to reduce write lock contention

thomhurst:
## Summary
- **Behavioral change**: default consumer TCP connections per broker increases from 1 to 2, reducing head-of-line blocking where heartbeats (every 3s) and offset commits (every 5s) contend with fetch requests for a single write lock (`SemaphoreSlim _writeLock`)
- Add `ConnectionsPerBroker` property to `ConsumerOptions` (default: 2) and `WithConnectionsPerBroker()` to `ConsumerBuilder` for configurability
- Wire the option through to `ConnectionPool` construction in `KafkaConsumer`, replacing the hardcoded `connectionsPerBroker: 1`
- Control-plane operations (heartbeat, commit, FindCoordinator) use dedicated coordination connection; fetch requests use the data-plane connection
- **Producer guard added**: `ProducerBuilder.WithConnectionsPerBroker` now enforces an upper bound of 32 (previously unbounded). This is unlikely to affect existing users but is a technically breaking change.

## Details
The `ConnectionPool` already fully supports multiple connections per broker with round-robin selection (used by the producer). This change simply enables it for the consumer. With 2 connections, fetch requests and coordination requests (heartbeats, commits, metadata) naturally separate onto different connections, reducing write lock contention on the hot path.

The consumer caps `ConnectionsPerBroker` at 2 (via `ConsumerOptions.MaxConnectionsPerBroker`) since it only uses two connection indices (data-plane and control-plane). Values above 2 would create unused connections.

Users who want the previous behavior can set `ConnectionsPerBroker = 1` in `ConsumerOptions` or call `.WithConnectionsPerBroker(1)` on the builder.

## Test plan
- [x] Verify build succeeds (`dotnet build`)
- [x] Unit tests for builder validation (zero, negative, above-max throw `ArgumentOutOfRangeException`)
- [x] Unit test for default value (`ConnectionsPerBroker_DefaultsTo_2`)
- [x] Integration test: produce-consume-commit cycle with `ConnectionsPerBroker(2)`
- [x] Integration test: single-connection fallback with `ConnectionsPerBroker(1)`
- [ ] Run stress tests / benchmarks to measure throughput improvement

### DIFF
--- a/tests/Dekaf.Tests.Unit/Consumer/PrefetchPipelineRunnerTests.cs
+++ b/tests/Dekaf.Tests.Unit/Consumer/PrefetchPipelineRunnerTests.cs
@@ -865,7 +865,10 @@ public class PrefetchPipelineRunnerTests
         var fetchPosition = 0L;
         var positionsReadByFetch = new ConcurrentDictionary<int, long>();
         var fetchCount = 0;
-        var targetFetchCount = pipelineDepth * 2; // Enough iterations to exercise the pipeline
+        // Fixed target: throughput is 2 fetches/iteration regardless of pipelineDepth
+        // (PR #648's "one eager per iteration" rule caps in-flight queue at 1).
+        // Scaling with pipelineDepth caused timeouts on slow CI runners for high depths.
+        var targetFetchCount = 8;
         CancellationTokenSource? testCts = null;
 
         var runner = CreateRunner(


thomhurst:
Addressed in 5a44b15:

**Dual validation drift** — Both builder and constructor now reference `ConsumerOptions.MaxConnectionsPerBroker` constant. Single-point edit for future limit changes.

**Producer 32-cap undocumented** — Updated PR description to call out the producer upper-bound guard as a technically breaking change.

PR #650, opened 2026-03-27
perf: adaptive connection scaling for consumer

### LINKED ISSUE #646: perf: adaptive connection scaling for consumer
## Context

PR #642 introduces `ConnectionsPerBroker` for the consumer (default 2), hard-pinning fetch traffic to connection index 0 and coordination traffic (heartbeats, commits, JoinGroup/SyncGroup) to connection index 1. This cleanly solves the head-of-line blocking problem where control-plane and data-plane operations contended for a single write lock.

However, the current implementation has a gap: setting `ConnectionsPerBroker` to any value >2 allocates extra connections that are never used, because fetch and coordination are pinned to fixed indices 0 and 1 respectively.

### Phase 1: Make `ConnectionsPerBroker` meaningful beyond 2

Change the fetch path to round-robin across connections `0..N-2` (all connections except the last, which is reserved for coordination). This makes the existing knob actually useful:

- `ConnectionsPerBroker = 1` → shared connection (graceful degradation, current behavior)
- `ConnectionsPerBroker = 2` → 1 fetch + 1 coordination (current default, clean separation)
- `ConnectionsPerBroker = 3` → 2 fetch (round-robin) + 1 coordination
- `ConnectionsPerBroker = N` → N-1 fetch (round-robin) + 1 coordination

This benefits consumers with high prefetch pipeline depth, large partition counts per broker, or large fetch responses where multiple in-flight requests on a single connection create write lock contention.

### Phase 2: Adaptive scaling (like the producer)

The producer already has adaptive connection scaling driven by buffer pressure. The consumer could use analogous signals:

| Signal | When to scale up | When to scale down |
|--------|-----------------|-------------------|
| Fetch latency | Fetch round-trip time increases beyond baseline | Latency returns to baseline |
| Assigned partition count | Many partitions on one broker → more concurrent fetch potential | Rebalance reduces partition count |
| Prefetch pipeline saturation | Pipeline depth consistently full, backpressure on fetch requests | Pipeline rarely fills |
| Consumer lag | Lag growing despite available data (fetch throughput bottleneck) | Lag stable or decreasing |

The producer's existing adaptive scaling infrastructure (`EnableAdaptiveConnections`, cooldown timers, deferred scaling for in-flight requests) could be generalized or adapted for the consumer.

### Open questions

- Should the coordination connection always be the **last** index (so scaling up/down only affects fetch connections and never disrupts coordination)?
- Should `MaxConnectionsPerBroker` default to something lower for consumers than producers (e.g., 4 vs 10) since fetch is less parallelizable than produce?
- Is fetch latency a reliable enough signal, or would it conflate broker-side slowness with connection contention?

## Related

- #642 — initial `ConnectionsPerBroker` support for consumer
- Producer adaptive scaling: `BrokerSender.MaybeScaleConnections()`, `ProducerOptions.EnableAdaptiveConnections`

thomhurst:
## Summary

Closes #646

- **Phase 1**: Round-robin fetch connections across indices `0..N-2`, coordination always uses last index (`N-1`). `ConnectionsPerBroker > 2` now actually distributes fetch traffic across multiple connections.
- **Phase 2**: `ConsumerConnectionScaler` monitors prefetch pipeline saturation and scales connections dynamically. Scale-up after 5s sustained saturation, scale-down after 120s below 30% utilization, with 5s cooldown between operations.
- **Configuration**: `EnableAdaptiveConnections` (default `true`), `MaxConnectionsPerBroker` (default 4), builder methods `WithAdaptiveConnections()`/`WithoutAdaptiveConnections()`

| `ConnectionsPerBroker` | Fetch indices | Coordination index |
|---|---|---|
| 1 | 0 | 0 (shared) |
| 2 | 0 | 1 |
| 3 | 0, 1 | 2 |
| 4 | 0, 1, 2 | 3 |

### Key design decisions

- Coordination index is **dynamic** (reads current connection count at call time) so scaling up doesn't cause fetch/coordination overlap
- No partition affinity concerns (unlike producer) — consumer fetches are per-broker, so round-robin is always safe
- Scale operations are fire-and-forget with exception observation via `ContinueWith`
- Pipeline saturation chosen as primary signal over fetch latency (avoids conflating broker slowness with connection contention)

## Test plan

- [x] Unit tests: connection index calculation, round-robin distribution, scaler thresholds (scale-up, scale-down, cooldown, boundaries, saturation reset) — 24 new tests
- [x] Integration test: consumer with `ConnectionsPerBroker=3` produces/consumes 50 messages with manual offset commit
- [x] Full unit test suite: 3233 tests passing
- [x] Integration test project builds successfully


thomhurst:
### Re: Follow-up Review — `_fetchRoundRobinCounter` race

Verified: `PrefetchFromBrokerAsync` runs concurrently per broker via `Task.WhenAll`, so `counter++` was indeed a data race. Fixed with `Interlocked.Increment(ref counter) - 1` in `GetNextFetchConnectionIndex`. Commit 99e9a7e.

PR #655, opened 2026-03-28
fix: suppress spurious 'unknown correlation ID' warnings for cancelled requests

### LINKED ISSUE #651: Connection contention under high parallelism: 'Received response for unknown correlation ID' warnings
## Summary

When integration tests run with high parallelism (16 concurrent tests), the following warnings appear in test output:

```
[Warning] Dekaf.Networking.KafkaConnection: Received response for unknown correlation ID 2590
[Warning] Dekaf.Networking.KafkaConnection: Received response for unknown correlation ID 2570
[Warning] Dekaf.Networking.KafkaConnection: Received response for unknown correlation ID 2595
```

These were observed in the Transaction matrix job from CI run [#23671568264](https://github.com/thomhurst/Dekaf/actions/runs/23671568264/job/68966215410?pr=649), triggered by PR #649 which removed parallel limiters from integration tests.

## Root Cause (suspected)

Under high concurrency, multiple consumers across different tests share connections via the connection pool. A response arrives for a correlation ID that has already been removed (e.g. the request timed out, was cancelled, or the consumer was disposed). The connection logs a warning and discards the response.

This can cause:
- Slower consumer group joins (responses discarded, must retry)
- Increased Kafka container load from retried requests
- Cascading timeouts in tests that depend on prompt partition assignment

## Impact

- Contributes to slow consumer group joins under load, which can expose timing-sensitive bugs (see #649 fix for `ConsumeOneAsync`)
- May cause flaky tests in high-parallelism CI environments even after the `ConsumeOneAsync` race fix

## Investigation needed

1. Is the connection pool correctly keying connections per broker, or are connections being shared across logical consumers in a way that causes response routing issues?
2. Is cancellation of in-flight requests properly cleaning up the correlation ID from the pending-response map before the response arrives?
3. Should there be a connection-per-consumer option, or stricter per-broker connection limits under high parallelism?

## Observed in

- CI run: https://github.com/thomhurst/Dekaf/actions/runs/23671568264/job/68966215410
- Branch: `feature/parallelise-tests` (PR #649)
- Test category: Transaction

thomhurst:
## Summary

- Track cancelled/timed-out correlation IDs so the receive loop can silently discard expected late responses instead of logging misleading "unknown correlation ID" warnings
- Preserves the Warning log for truly unexpected correlation IDs (protocol errors, data corruption)
- No changes to pool-return semantics, `PooledPendingRequest` ownership, or `FailAllPendingRequests` behavior

## Root Cause

When a request times out or is cancelled after being written to the broker, `AwaitAndParseResponseAsync`'s `finally` block removes the correlation ID from `_pendingRequests` before the broker's response arrives. The receive loop then can't find it → `LogUnknownCorrelationId` warning.

Under high parallelism (16 concurrent tests from #649), the single Kafka container gets overloaded, causing more timeouts and making this race much more frequent.

## How It Works

1. **New field**: `_cancelledCorrelationIds` (`ConcurrentDictionary<int, byte>`) tracks correlation IDs where the request was sent but the response was not received
2. **On timeout/cancellation**: after removing from `_pendingRequests` and returning to pool, the correlation ID is added to `_cancelledCorrelationIds`
3. **In receive loop**: checks `_cancelledCorrelationIds` before logging Warning — silently discards at Debug level if found
4. **Cleanup**: entries removed when the late response arrives; remaining entries GC'd on connection disposal

## Test plan

- [x] All 3233 unit tests pass
- [x] Zero build warnings
- [ ] Integration tests with high parallelism no longer produce "unknown correlation ID" warnings

Closes #651

### DIFF
--- a/src/Dekaf/Networking/KafkaConnection.cs
+++ b/src/Dekaf/Networking/KafkaConnection.cs
@@ -132,6 +132,7 @@ public sealed partial class KafkaConnection : IKafkaConnection
     // Using globally unique correlation IDs prevents this issue.
     private static int s_globalCorrelationId;
     private readonly ConcurrentDictionary<int, PooledPendingRequest> _pendingRequests = new();
+    private readonly ConcurrentDictionary<int, byte> _cancelledCorrelationIds = new();
     private readonly PendingRequestPool _pendingRequestPool = new();
     private readonly CancellationTokenSourcePool _timeoutCtsPool = new();
     private readonly SemaphoreSlim _writeLock = new(1, 1);
@@ -530,6 +531,7 @@ public sealed partial class KafkaConnection : IKafkaConnection
         where TRequest : IKafkaRequest<TResponse>
         where TResponse : IKafkaResponse
     {
+        var responseReceived = false;
         try
         {
             LogWaitingForResponse(correlationId);
@@ -554,6 +556,8 @@ public sealed partial class KafkaConnection : IKafkaConnection
             {
                 pending.DisposeRegistration();
             }
+
+            responseReceived = true;
 
             LogResponseReceived(correlationId);
 
@@ -604,6 +608,15 @@ public sealed partial class KafkaConnection : IKafkaConnection
             if (_pendingRequests.TryRemove(correlationId, out var removed))
             {
                 _pendingRequestPool.Return(removed);
+
+                // Broker may still send a response for a timed-out/cancelled request.
+                // Record it so the receive loop discards silently instead of warning.
+                // If the receive loop already called TryComplete (and lost the CAS),
+                // this entry becomes a harmless no-op cleaned up by DisposeAsync.
+                if (!responseReceived)
+                {
+                    _cancelledCorrelationIds.TryAdd(correlationId, 0);
+                }
             }
         }
     }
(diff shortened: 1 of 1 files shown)


thomhurst (commented):
**Point 1 (`_cancelledCorrelationIds` unbounded / missing `Clear()`):** Already addressed — `_cancelledCorrelationIds.Clear()` is at line 1813 in `DisposeAsync`, added during a post-commit cleanup pass. The PR description was written before that pass and is now stale on this point.

**Point 2 (write-phase cancellation not adding to `_cancelledCorrelationIds`):** Confirmed correct — the `TryRemove` guard at line 386 is load-bearing here. If `AwaitAndParseResponseAsync`'s finally already removed the entry (and added to `_cancelledCorrelationIds`), this catch's `TryRemove` returns false and skips. If the write phase failed, this catch's `TryRemove` succeeds but correctly doesn't track the ID (no request sent = no response coming). The existing comment at line 383-385 explains the `TryRemove` interaction between the two cleanup sites.

**Points 3-5:** Agreed these are all correct as-is. No changes needed.

thomhurst (commented):
**Point 1 (stale entries from TryComplete race):** Good catch on the race window. Added a comment documenting this — the entry is harmless dead weight cleaned up by `DisposeAsync.Clear()`.

**Point 2 (`responseReceived` naming):** Keeping the current name. `responseReceived` expresses the domain concern directly (did we get the broker's response?), which is what the `_cancelledCorrelationIds` logic cares about. More mechanical names like `awaitSucceeded` would require mapping back to why it matters.

PR #709, opened 2026-03-29
perf: split consumer fetches across multiple connections per broker

thomhurst:
## Summary

- When multiple fetch connections are available per broker, partition assignments are now split across connections and fetched concurrently within each prefetch cycle
- Adds `SplitPartitionsAcrossConnections` to `ConsumerConnectionScaler` for chunked partition distribution (stackalloc, zero heap allocation on hot path)
- Updates `PrefetchRecordsAsync` to issue N concurrent fetch tasks per broker (one per connection) instead of 1
- Changes `_prefetchPendingItemsByBroker` key from `int` (brokerId) to `(int, int)` (brokerId, connectionIndex) to avoid list reuse conflicts with concurrent fetches to the same broker

This enables true parallel fetch I/O to the same broker, reducing round-trip latency impact on consumer throughput. With `ConnectionsPerBroker=3` (2 fetch connections), a broker with 6 partitions now issues 2 concurrent fetches of 3 partitions each.

## Test plan

- [x] All 13 `ConsumerConnectionScalerTests` pass (including 6 new partition split tests)
- [x] Covers: single connection, even split, uneven split, more connections than partitions, single partition, exhaustive coverage check
- [ ] Integration test with multi-partition topic to verify throughput improvement

### DIFF
--- a/src/Dekaf/Consumer/ConsumerConnectionScaler.cs
+++ b/src/Dekaf/Consumer/ConsumerConnectionScaler.cs
@@ -123,16 +123,6 @@ internal sealed class ConsumerConnectionScaler
     internal static int GetFetchConnectionCount(int connectionsPerBroker)
         => connectionsPerBroker > 1 ? connectionsPerBroker - 1 : 1;
 
-    /// <summary>
-    /// Returns the next fetch connection index using round-robin.
-    /// </summary>
-    internal static int GetNextFetchConnectionIndex(ref int counter, int fetchConnectionCount)
-    {
-        if (fetchConnectionCount == 1) return 0;
-        var next = Interlocked.Increment(ref counter) - 1;
-        return next % fetchConnectionCount;
-    }
-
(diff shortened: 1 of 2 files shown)


thomhurst:
Addressed both remaining items:

**1. Dictionary pruning:** Added lazy pruning at the top of `PrefetchRecordsAsync` — iterates `_prefetchPendingItemsByBroker.Keys` and removes entries where `ConnectionIndex >= fetchConnectionCount`. O(k) over a small dictionary, runs once per fetch cycle (not hot path). No entries persist after scale-down.

**2. Cache bypass documented:** Added comment on the `isFullList` check in `BuildFetchRequestTopics` explaining that when `fetchConnectionCount > 1`, all calls use subranges so the cache is intentionally bypassed. The cache only applies to single-connection mode. Extending it for subrange keying would add complexity for marginal benefit since the dictionary rebuild is per-batch, not per-message.

PR #787, opened 2026-04-10
perf(producer): use partition affinity for all connection distribution

thomhurst:
## Summary

- **Fix 0.66x throughput gap** in non-idempotent fire-and-forget producer with 3 brokers (132K vs 200K msg/s compared to Confluent)
- Non-idempotent producers previously used round-robin to distribute batches across connections, breaking CPU cache locality by scattering the same partition's batches across different TCP connections
- Change all producers to use partition affinity (`partition % connectionCount`), matching the idempotent path that already achieves 1.00x parity

## What changed

In `BrokerSender.cs`, the multi-connection batch distribution logic:
- **Before**: Idempotent used `partition % N`, non-idempotent used `roundRobinCounter++ % N`
- **After**: All producers use `partition % N` unconditionally

Additional cleanup:
- Removed `roundRobinCounter` variable (no longer needed)
- Renamed `_requirePartitionAffinity` to `_isIdempotent` (now only gates sequence assignment and scaling drain guards)
- Updated comments to reflect the unified partition affinity strategy

## Why this works

Partition affinity keeps each partition's batches on the same connection, which:
1. **Improves CPU cache locality** — per-partition metadata stays warm in L1/L2 cache
2. **Reduces TCP connection contention** — each connection handles a consistent set of partitions
3. **No correctness trade-off** — Kafka handles per-partition ordering at the broker level regardless of which TCP connection delivers the batch

## Test plan

- [x] All 21 BrokerSender unit tests pass
- [x] All 325 Producer unit tests pass
- [x] No new allocations in hot path (removed a branch and a counter)
- [ ] Run stress tests with 3-broker non-idempotent fire-and-forget to verify throughput improvement

### DIFF
--- a/src/Dekaf/Producer/BrokerSender.cs
+++ b/src/Dekaf/Producer/BrokerSender.cs
@@ -2863,8 +2863,8 @@ internal sealed partial class BrokerSender : IAsyncDisposable
         // Idempotent producers require ALL connections to be drained before shrinking
         // because partitions remap (P % N -> P % (N-1)) and in-flight batches on any
         // connection could conflict with new batches post-remap, causing sequence errors.
-        // Non-idempotent producers only need the last connection idle — partition affinity
-        // means other connections still serve their assigned partitions without conflict.
+        // Non-idempotent producers only need the last connection idle — no sequence numbers
+        // means partition remapping mid-flight cannot cause ordering violations.
         if (_isIdempotent)
         {
             if (_totalPendingResponseCount > 0)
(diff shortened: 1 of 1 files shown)


thomhurst (commented):
Addressed the feedback:

1. **Connections > Partitions**: Added a comment at the distribution site noting that excess connections will be idle and adaptive scaling handles this by scaling them down. For users who explicitly set `ConnectionsPerBroker` higher than partition count — this is a behavior change worth noting in release notes, agreed.

2. **Drain guard comment**: Clarified that the non-idempotent drain guard difference is about partition affinity not causing conflicts (since there are no sequence numbers), not about the distribution strategy being different.

3. **Stress test**: Can't run from here — needs the CI workflow. The reasoning is sound (partition affinity eliminates cross-connection cache thrashing that round-robin caused), and idempotent 3-broker already validates this approach at 1.00x parity.

thomhurst (commented):
Fixed the drain guard comment. The reviewer is correct — partition affinity doesn't prevent remapping during scale-down (P % N → P % (N-1) remaps most partitions). The actual reason is simply: no sequence numbers means no ordering violations from remapping. Updated to match the precise wording suggested.

PR #788, opened 2026-04-10
perf(producer): reduce idempotent producer GC pressure from sequence tracking

thomhurst:
## Summary

- Replace `TryRemove`/`Clear` + `GetOrAdd` pattern with in-place `Interlocked.Exchange(ref box.Value, 0)` for sequence number resets in `RecordAccumulator`
- During epoch bump recovery, the previous pattern removed `ConcurrentDictionary` entries, forcing reallocation of internal `Node` objects (~48 bytes) and `StrongBox<int>` instances (~24 bytes) per partition on next access
- These mid-lived objects survived Gen0 during the network round-trip, got promoted to Gen2, then died -- contributing to the pathological Gen2 GC pressure observed in idempotent producer stress tests (658 GB allocated, 450 Gen2 collections vs 4.65 GB / 17 Gen2 for non-idempotent)

## What changed

`ResetSequenceNumbers()` and `ResetSequencesForPartitions()` in `RecordAccumulator.cs` now reset `StrongBox<int>.Value` to 0 in place rather than removing dictionary entries. This eliminates Node + StrongBox reallocation when sequences are next assigned, keeping the idempotent path allocation-free in steady state.

## Test plan

- [x] All 3435 unit tests pass (including 57 epoch bump recovery tests that directly exercise `ResetSequenceNumbers`, `ResetSequencesForPartitions`, and concurrent access)
- [ ] Run idempotent producer stress test to verify reduced Gen2 collections

### DIFF
--- a/src/Dekaf/Producer/RecordAccumulator.cs
+++ b/src/Dekaf/Producer/RecordAccumulator.cs
@@ -924,6 +924,12 @@ public sealed partial class RecordAccumulator : IAsyncDisposable
     // (lock-free hash lookup only), avoiding AddOrUpdate's per-call bucket locking.
     // During leader migration, two BrokerSender threads may access the same partition
     // concurrently, so StrongBox.Value is mutated via Interlocked.Add for atomicity.
+    //
+    // Reset operations use Interlocked.Exchange(ref box.Value, 0) instead of TryRemove
+    // to avoid re-allocating ConcurrentDictionary internal Node objects (~48 bytes) and
+    // StrongBox instances (~24 bytes) per partition. Under epoch bump recovery, TryRemove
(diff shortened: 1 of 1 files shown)


thomhurst (commented):
Addressed both points:

1. **O(n) vs O(1)**: Added doc noting this is O(n) over tracked partitions, acceptable for the epoch-bump recovery path since n is bounded by partition count. This method runs once per epoch bump (rare error recovery), not on the hot path.

2. **Append-only dictionary**: Added a `<remarks>` block documenting this is intentional. The dictionary grows monotonically but is bounded by the number of partitions the producer writes to — typically a fixed, small set. The tradeoff (bounded memory retention vs Gen2 GC churn from Node/StrongBox reallocation) strongly favors append-only for any realistic partition count.

PR #789, opened 2026-04-10
perf(consumer): batch per-message overhead to per-fetch granularity

thomhurst:
## Summary
- Move `_positions` ConcurrentDictionary write from per-message to per-fetch, eliminating ~20-50ns per message from hash + lock + barrier overhead
- Batch metrics `Counter<T>.Add` calls into single per-fetch calls using `PendingFetchData.MessageCount`/`TotalBytesConsumed`
- Add finally-block position flush for partially-iterated fetches on early `break`

Targets the 0.89x consumer throughput gap vs Confluent.Kafka (97K vs 109K msg/s). Both position tracking and metrics were unnecessary at per-message granularity since `GetPosition()`/`CommitAsync()` are called infrequently (auto-commit every 5s).

## Test plan
- [x] Unit tests pass (ConsumeAsyncRecoveryTests updated for per-fetch position semantic)
- [ ] Stress test comparison to verify throughput improvement

### DIFF
--- a/src/Dekaf/Consumer/IKafkaConsumer.cs
+++ b/src/Dekaf/Consumer/IKafkaConsumer.cs
@@ -102,9 +102,6 @@ public interface IKafkaConsumer<TKey, TValue> : IInitializableKafkaClient, IAsyn
 
     /// <summary>
     /// Gets the current position (next offset to consume) for a partition.
-    /// Updated at per-fetch granularity (once per partition-fetch, not per message)
-    /// for performance. Reflects the offset after the last fully-consumed fetch,
-    /// or the last yielded message if the caller broke out of the consume loop early.
     /// </summary>
     long? GetPosition(TopicPartition partition);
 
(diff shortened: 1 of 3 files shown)


thomhurst (commented):
Addressed the review feedback:

1. **Metrics flush on early break (Issue 2)**: Added metrics flush in the `finally` block mirroring the position flush pattern. Messages yielded before break are now counted.

2. **GetPosition() docs (Issue 3)**: Updated the XML doc on `IKafkaConsumer.GetPosition()` to document per-fetch update granularity and the early-break flush behavior.

3. **Finally comment (Issue 1)**: Tightened to clarify it's only relevant on early break/exception, and is a no-op on normal loop exit.

4. **Test rename (minor)**: Skipping — the current name still accurately describes what's tested (partial batch consumption reflects last yielded offset), and the semantic hasn't changed from the test's perspective.

thomhurst (commented):
Addressed both findings:

1. **Metrics duplication**: Extracted to `EmitFetchMetrics(PendingFetchData)` — both call sites are now one-liners. Hoisted `metricsEnabled` above the `try` block since C# locals inside `try` aren't accessible in `finally`.

2. **`_fetchPositions` flush in finally**: Added `if (!_prefetchEnabled) _fetchPositions[tp] = offset + 1` alongside the `_positions` flush, so re-calling `ConsumeAsync` after an early break resumes from the correct offset instead of re-delivering already-consumed messages.

PR #808, opened 2026-04-12
perf(consumer): defer _positions writes to batch boundary in auto-commit mode

thomhurst:
## Summary

- **Skip per-message `_positions` ConcurrentDictionary write in auto-commit mode.** Auto-commit reads `_positions` every ~5s, so batch-granularity updates via `FlushConsumedPositions()` are sufficient. This eliminates ~60-100ns of hash computation + bucket lock per message in the hot path.
- **Manual-commit retains per-message writes** so `GetPosition()` and `CommitAsync()` reflect the latest consumed offset mid-batch.
- **Extract `FlushConsumedPositions()` helper** consolidating position update logic across 4 call sites (batch boundary, finally block, ConsumeBatchAsync, ConsumeRawBatchAsync).
- **Fix finally-block gap**: previously the early-break path only flushed `_fetchPositions`, now also flushes `_positions`.

## Design

The optimization uses a hoisted boolean (`var manualCommit = ...`) evaluated once per `ConsumeAsync()` call. In auto-commit mode, `_positions` is written once per partition-fetch batch (~1000 messages) instead of per message. `FlushConsumedPositions()` at the batch boundary and in the finally block ensures `_positions` is always consistent when the consume loop yields control.

## Test plan

- [x] All 3,506 unit tests pass (1 pre-existing failure in BufferMemoryTests unrelated to this change)
- [ ] Integration tests pass in CI
- [ ] Stress test comparison shows throughput improvement for auto-commit consumers

Tracking deeper structural improvements in #809.

### DIFF
--- a/src/Dekaf/Consumer/IKafkaConsumer.cs
+++ b/src/Dekaf/Consumer/IKafkaConsumer.cs
@@ -119,6 +119,8 @@ public interface IKafkaConsumer<TKey, TValue> : IInitializableKafkaClient, IAsyn
 
     /// <summary>
     /// Gets the current position (next offset to consume) for a partition.
+    /// In auto-commit mode, reflects the last fully-processed batch boundary;
+    /// for per-message accuracy, use manual commit mode.
     /// </summary>
     long? GetPosition(TopicPartition partition);
 
--- a/src/Dekaf/Consumer/KafkaConsumer.cs
+++ b/src/Dekaf/Consumer/KafkaConsumer.cs
@@ -1089,6 +1089,11 @@ public sealed partial class KafkaConsumer<TKey, TValue> : IKafkaConsumer<TKey, T
 
                             pending.TrackConsumed(offset, messageBytes);
 
+                            // Update consumed position per-message so GetPosition()/CommitAsync()
+                            // reflect the latest message the app has seen, even mid-batch.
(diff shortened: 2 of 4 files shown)


thomhurst:
Thanks for the thorough review! The manual-commit regression was caught independently during a `/simplify` pass and addressed in the second commit (`5c97dfe`).

**What the final diff actually does (both commits combined):**

1. **Per-message `_positions` write is retained** — the first commit removed it, the second restored it. The combined diff shows no change to the per-message update at line 1095.

2. **`FlushConsumedPositions()` helper extracted** — replaces 4 identical inline `_positions` + `_fetchPositions` update blocks across `ConsumeAsync` (batch boundary + finally), `ConsumeBatchAsync`, and `ConsumeRawBatchAsync`.

3. **Finally-block bug fix** — the `ConsumeAsync` finally block previously only flushed `_fetchPositions` (and only in non-prefetch mode). It now calls `FlushConsumedPositions()` which also updates `_positions`. This fixes a real gap: if the user broke out of `await foreach` mid-batch, `_positions` could be stale on the next `CommitAsync()` call.

Re: the minor note about `Peek()` only flushing one partition — agreed, that's intentional. Un-iterated batches in the queue haven't been yielded to the user, so advancing their position would be incorrect.

PR #833, opened 2026-04-12
feat: handle REBOOTSTRAP_REQUIRED error code (KIP-1102)

### LINKED ISSUE #820: KIP-1102: Handle REBOOTSTRAP_REQUIRED error code from broker
## Summary

Implement handling for the `REBOOTSTRAP_REQUIRED` error code introduced in KIP-1102. When a Kafka 4.0+ broker returns this error in a MetadataResponse, the client should immediately trigger a rebootstrap instead of waiting for the timer-based trigger.

**KIP**: https://cwiki.apache.org/confluence/display/KAFKA/KIP-1102

## Current State

Dekaf already has a robust rebootstrap implementation (KIP-899):
- `MetadataRecoveryStrategy.Rebootstrap` is the default strategy
- `MetadataRecoveryRebootstrapTriggerMs` (default 300000ms / 5min) controls timer-based trigger
- DNS re-resolution and reconnection logic exists in `MetadataManager.TryRebootstrapAsync()`

However, the `REBOOTSTRAP_REQUIRED` error code is **not defined in `ErrorCode.cs`** and is not handled in the metadata response path. The client currently relies solely on the timer-based trigger when all brokers become unavailable.

### 1. Add Error Code
Add `RebootstrapRequired` to the `ErrorCode` enum with the appropriate numeric value.

### 2. Handle in MetadataResponse Processing
In the metadata response handler (likely in `MetadataManager`):
- Check for `REBOOTSTRAP_REQUIRED` error code in the response
- When received, **immediately** trigger `TryRebootstrapAsync()` — bypass the timer-based delay
- Log the broker-initiated rebootstrap at `Information` level

### 3. Rebootstrap as Default (KIP-1102 Enhancement)
Verify that `MetadataRecoveryStrategy.Rebootstrap` is the default (it is) and that `metadata.recovery.rebootstrap.trigger.ms` defaults to 300000ms (5 minutes) — matching the KIP-1102 spec.

## Why This Matters

Without this, when a broker explicitly signals that the client should rebootstrap (e.g., during a rolling cluster migration or IP change), the client waits up to 5 minutes before attempting recovery. With the error code handling, recovery is near-instant.

## Testing Requirements

- Unit test: verify `REBOOTSTRAP_REQUIRED` error in MetadataResponse triggers immediate rebootstrap
- Unit test: verify the timer-based trigger still works as fallback when no error code is received
- Integration test (if possible): verify behavior during broker address change

## Effort Estimate

Small — likely a few lines of code in `ErrorCode.cs` and `MetadataManager.cs`.

## References

- [KIP-1102 Wiki](https://cwiki.apache.org/confluence/display/KAFKA/KIP-1102)
- [KIP-899 Wiki](https://cwiki.apache.org/confluence/display/KAFKA/KIP-899) (base rebootstrap implementation)
- Existing implementation: `src/Dekaf/Metadata/MetadataManager.cs`, `src/Dekaf/Metadata/MetadataRecoveryStrategy.cs`

thomhurst:
## Summary

- Add Kafka protocol error codes 121-129 to `ErrorCode` enum, including `RebootstrapRequired = 129`
- Bump `MetadataRequest` and `MetadataResponse` to v13 — reads new top-level `ErrorCode` field from response
- When broker returns `REBOOTSTRAP_REQUIRED`, immediately trigger rebootstrap (bypass 5-minute timer)
- Fix `IncludeClusterAuthorizedOperations` version guard in `MetadataRequest.Write` (field is v8-10 only per Kafka protocol spec)

Closes #820

## How It Works

When a Kafka 4.0+ broker (or protocol-aware proxy) returns `REBOOTSTRAP_REQUIRED` in a MetadataResponse, the client calls `TryRebootstrapAsync` with `immediate: true`, which skips the timer-based delay and goes straight to DNS re-resolution and new broker connections. The existing timer-based fallback (KIP-899) continues to work unchanged for cases where the broker doesn't signal explicitly.

## Test plan

- [x] `RebootstrapRequired_ErrorCode_HasValue129` — enum value is correct
- [x] `MetadataResponse_ErrorCode_DefaultsToNone` — default init value
- [x] `MetadataResponse_ErrorCode_CanBeSetToRebootstrapRequired` — property round-trip
- [x] `TryRebootstrapAsync_Immediate_SkipsTimerDelay` — immediate mode bypasses CAS timer, doesn't consume normal first-call slot
- [x] Full unit test suite: 3315/3315 passing

### DIFF
--- a/src/Dekaf/Metadata/MetadataManager.cs
+++ b/src/Dekaf/Metadata/MetadataManager.cs
@@ -464,6 +464,13 @@ public sealed partial class MetadataManager : IAsyncDisposable
                 // Success - reset the rebootstrap timer
                 ResetAllBrokersUnavailableTimestamp();
 
+                // KIP-1102: Broker signaled that client should rebootstrap immediately
+                if (response.ErrorCode == ErrorCode.RebootstrapRequired
+                    && _options.MetadataRecoveryStrategy == MetadataRecoveryStrategy.Rebootstrap)
+                {
+                    await TryRebootstrapAsync(topics, cancellationToken, immediate: true).ConfigureAwait(false);
+                }
+
                 return;
             }
             catch (BrokerVersionException)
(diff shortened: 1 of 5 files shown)


thomhurst:
All 5 findings addressed in 24de71e:

1. **Boolean parameter → named methods**: Split into `TryRebootstrapAsync` (timer-gated), `TryRebootstrapImmediateAsync` (KIP-1102), and shared `ExecuteRebootstrapAsync`. No more boolean trap.

2. **Stale metadata applied before rebootstrap**: `_metadata.Update` now deferred — rebootstrap runs first; original response only applied as fallback if rebootstrap fails.

3. **Return value discarded**: Follows from #2 — `rebootstrapped` return value is now checked to decide whether to apply fallback metadata or return early.

4. **Stale XML doc**: `IncludeClusterAuthorizedOperations` doc updated to `v8–v10`.

5. **Enum test**: Added comment explaining the test pins to the Kafka protocol spec wire value.

3315/3315 tests pass.

thomhurst:
Addressed the remaining observation in a7aad3b: added `LogRebootstrapChainSuppressed()` at Warning level inside `ExecuteRebootstrapAsync` when the newly-resolved broker also returns `REBOOTSTRAP_REQUIRED`. Makes the suppressed chain visible to operators without changing any semantics.

thomhurst:
Verified against the actual Kafka source schema ([MetadataResponse.json on trunk](https://github.com/apache/kafka/blob/trunk/clients/src/main/resources/common/message/MetadataResponse.json)):

```json
{ "name": "ErrorCode", "type": "int16", "versions": "13+", "ignorable": true,
  "about": "The top-level error code, or 0 if there was no error." }
```

There is no `"tag"` or `"taggedVersions"` property on this field. It is a **regular sequential field**, not a tagged field. The current implementation correctly reads it as a positional `int16` before `SkipTaggedFields()`, matching the wire format specified in the Kafka protocol schema.

PR #884, opened 2026-06-30
feat(security): SASL/OAUTHBEARER auth tests across transports; fix auth-layer bugs

thomhurst:
## Summary

Adds dockerized integration coverage proving the auth layer works **across every
authentication type and both transports**, and fixes the real bugs that writing the
tests surfaced. Each authentication type now runs against a real `apache/kafka:4.0.1`
broker via TUnit + Testcontainers.

| Mechanism | SASL_PLAINTEXT | SASL_SSL ("not plaintext") |
|---|:--:|:--:|
| PLAIN | ✅ | ✅ |
| SCRAM-SHA-256 | ✅ | ✅ |
| SCRAM-SHA-512 | ✅ | ✅ |
| OAUTHBEARER | ✅ (new) | ✅ (new) |

23 auth integration tests, all green. Producer, consumer, and admin client are each
exercised, plus invalid-credential / invalid-token rejection.

## The bugs this surfaced (production fixes)

These paths **never worked end-to-end before** — the existing SASL tests couldn't even
start their broker, so the client bugs were never exercised.

- **PLAIN & OAUTHBEARER never sent credentials.** `KafkaConnection` sent the SASL initial
  response inside `while (!IsComplete)`, but `PlainAuthenticator`/`OAuthBearerAuthenticator`
  mark themselves complete in `GetInitialResponse()`, so the loop body was skipped and the
  credential/token was never transmitted. Fixed with `do/while` (send at least once).
- **OAUTHBEARER framing was malformed.** The initial response emitted a single trailing
  `SOH` instead of the RFC 7628 `...auth=Bearer <token>\x01\x01` double terminator, so the
  broker rejected every token. Also stopped completing before the server's error challenge
  was read (an invalid token left the broker mid-handshake → `IllegalSaslState`).
- **Two `AuthenticationException` types.** SASL code threw `Dekaf.Security.Sasl.AuthenticationException : Exception`,
  but the documented hierarchy, the consumer's fatal-error classification, and callers expect
  `Dekaf.Errors.AuthenticationException : KafkaException`. Unified onto the latter so auth
  failures are catchable as `KafkaException` and correctly treated as fatal.
- **Auth failures were masked during metadata refresh.** `MetadataManager` swallowed auth
  exceptions and surfaced a generic "failed to refresh metadata from any broker". Now a single
  `IsFatalMetadataError` predicate rethrows them at all three sites (loop, init retry, and the
  previously-missed rebootstrap path).
- **Admin client couldn't do OAUTHBEARER/GSSAPI.** Added `WithOAuthBearer`/`WithGssapi` to
  `AdminClientBuilder` for parity with the producer/consumer builders.

## Test-fixture fixes

- `SaslKafkaContainer`: the SCRAM JAAS env-var names used double underscores, which
  `KafkaDockerWrapper` maps to `_` not `-`, so `scram-sha-256`/`512` JAAS never applied and the
  broker failed to start. Use triple underscores. SCRAM credentials are now created through
  Dekaf's own `AlterUserScramCredentialsAsync` instead of a `kafka-configs.sh` exec.
- TLS / SASL_SSL fixtures: mount certificates via the `byte[]` `WithResourceMapping` overload —
  the string-path overload bind-mounts and creates a *directory* at the target on Windows hosts,
  which also blocked the existing TLS/mTLS suite.

## Cleanup (`/simplify`)

- Shared `WaitForAdminReadyAsync` / `CreateScramCredentialsAsync` helpers on `KafkaTestContainer`
  replace three duplicated readiness loops and two SCRAM blocks.
- Hand-rolled base64url replaced with `System.Buffers.Text.Base64Url`.

**Deliberately skipped** (noted for a focused follow-up): extracting the ~95%-duplicated
`PerformSaslExchangeAsync` / `PerformSaslReauthExchangeAsync` into one helper. It's pre-existing
duplication in core connection code whose re-auth path has no integration coverage, so it's out
of scope for this change.

## Testing

- Unit: 3536/3536 pass.
- Integration (Docker): 23/23 auth tests pass (SASL_PLAINTEXT ×3 mechanisms, OAUTHBEARER,
  SASL_SSL ×4 mechanisms); a representative TLS test passes via the cert-mount fix.

> Note: the ACL fixture has a separate pre-existing startup bug (missing CONTROLLER listener
> protocol) — it's authorization, not authentication, so it's intentionally out of scope here.


thomhurst:
## Update — wired into CI + full Security coverage

The auth/security integration tests had **no category in the integration-test matrix**, so they never ran in CI — which is exactly why these fixtures had gone silently broken. Added matrix coverage and fixed everything it surfaced.

### CI matrix
- **`Authentication`** — SASL PLAIN/SCRAM-256/SCRAM-512/OAUTHBEARER over `SASL_PLAINTEXT` and `SASL_SSL`. Confirmed the category filter selects exactly these (23/23 pass).
- **`Security`** — TLS + ACL enforcement.

### Authorization errors now surface as `AuthorizationException`
The library defined `AuthorizationException` but **never threw it** — authz failures came out as generic `KafkaException` (or, for describe, a silent empty result). Added `KafkaException.FromErrorCode` (authz codes → `AuthorizationException`, authn codes → `AuthenticationException`) and applied it at:
- idempotent-producer init (`ClusterAuthorizationFailed`),
- producing to a topic with no partitions due to `TopicAuthorizationFailed`,
- `AdminClient.DescribeTopicsAsync` (was silently returning empty),
- TLS handshake rejection (`KafkaConnection` wraps the BCL `AuthenticationException` so it's a `KafkaException` and fatal during metadata refresh).

### ACL fixture (was pre-existing-broken)
- Broker wouldn't start — the listener map omitted `CONTROLLER`; added `CONTROLLER:PLAINTEXT` + `ANONYMOUS` super-user (external SASL listener still enforces ACLs on `testuser`).
- Restructured the producer ACL tests to catch the denial whether it surfaces at build (idempotent init) or produce. **8/8 pass (~9s).**

### TLS
- Cert-validation tests restructured to assert at connection time. **12/16 pass.**
- The **4 mutual-TLS tests fail on Windows only** with an SChannel error (`SEC_E_ILLEGAL_MESSAGE`) on the post-handshake data exchange — server-only TLS works over the same path, and it reproduces on both TLS 1.2 and 1.3. This looks platform-specific (CI is Linux/OpenSSL), so they're `[SkipOnWindows]` and **run on the Linux CI runner** to validate the real mTLS path. If CI shows them failing on Linux too, that's a genuine client bug to chase separately.

### Local validation
- Unit: 3536/3536. Authentication: 23/23. Security: ACL 8/8, TLS 12/12 (+4 mTLS skipped on Windows).

thomhurst:
## Addressed review + CI feedback (`50159c0`)

**Review point (batch `DescribeTopics`)** — fixed. `DescribeTopicsAsync` no longer throws for the whole batch when one topic is inaccessible. Added `ErrorCode` to `TopicDescription`; every requested topic present in metadata is returned, errored ones (e.g. `TopicAuthorizationFailed`) carrying their per-topic code. Callers inspect `description.ErrorCode` instead of one denied topic failing the call. The ACL describe test now asserts the per-topic error.

**Split the slow `Security` job** — `Security` is now two parallel categories: **`Tls`** and **`Authorization`** (ACL). ACL (~10s) no longer waits behind the TLS handshake tests.

**mTLS reality check** — the first CI run showed the 4 mutual-TLS tests fail on **Linux too** (`Connection closed by remote peer (EOF)` — the broker rejects the mTLS connection), not just Windows SChannel. So mTLS is genuinely broken (pre-existing — it never ran before this PR wired Security into the matrix), not a platform quirk. I've `[Skip]`ed those 4 with a documented reason and will chase the mTLS handshake/cert-config separately. Server-side TLS + certificate validation remain covered.

Everything else on Linux CI is green — notably **Integration Tests (Authentication) passed (23/23)**, validating SASL PLAIN/SCRAM/OAUTHBEARER over both `SASL_PLAINTEXT` and `SASL_SSL`.

thomhurst:
## mTLS fixed (`ec236f4`) — it wasn't platform-specific after all

Captured the broker-side logs from a failing mTLS handshake. The real cause:

```
java.security.InvalidAlgorithmParameterException: the trustAnchors parameter must be non-empty
    at sun.security.ssl...X509TrustManagerImpl.checkClientTrusted
```

The fixture's truststore was a **.NET-exported cert-only PKCS12**, which has no key entry — so Kafka/Java loaded it with **zero trust anchors** and rejected every client certificate during the mTLS handshake. Server-only TLS worked because it only touches the keystore (which has a key). The Windows `SEC_E_ILLEGAL_MESSAGE` was just the downstream symptom of the broker closing the connection mid-handshake.

Fix: switch the broker truststore to **PEM** (the CA cert, already mounted), which Java reads as a trust anchor. Un-skipped the 4 mutual-TLS tests — **all 4 pass** (in-memory certs, PEM file paths, the `UseMutualTls` factory, and the consumer round-trip).

So the Security split is now fully green with no skips: `Authorization` (ACL, 8/8) and `Tls` (16/16, including mTLS).

thomhurst:
## Addressed the review

Thanks — thorough review. Worked through all 8:

1. **`FromErrorCode` only wired into 2 of ~5 sites** — fixed. Routed the fire-and-forget `Send` and callback produce paths (via a shared `NoUsablePartitionsException` helper), `BrokerSender`'s per-partition produce-response handling, and `KafkaShareConsumer.SendAcknowledgeAsync` through `KafkaException.FromErrorCode`. A `catch (AuthorizationException)` now fires for all produce/ack paths, not just `ProduceInternalAsync`.
2. **Background refresh loop swallowed fatal auth** — fixed. `BackgroundRefreshLoopAsync` now has an `IsFatalMetadataError` carve-out that logs at Error and stops the loop; the next foreground operation surfaces the exception (instead of retrying forever after a mid-session credential/ACL change).
3. **TLS-fatal breadth** — confirmed deliberate and documented it at the wrap site: TLS config is cluster-wide, so a handshake failure recurs on every broker; trying the rest only delays the real cause. (Comment now notes .NET throws this type for negotiation mismatches too.)
4. **Admin builder null validation + tests** — fixed. `WithGssapi`/`WithOAuthBearer` now `throw new ArgumentNullException` like the producer/consumer builders, with 6 new unit tests in `AdminClientBuilderTests`.
6. **Duplicated `CreateScramAdminClient`** — extracted onto `KafkaTestContainer` with an `ApplyTransportSecurity` hook (SASL_SSL overrides it to add TLS).
7. **`KAFKA_SUPER_USERS=ANONYMOUS` invariant** — documented in the fixture (safe only while every client-reachable listener requires real auth).

**Deferred (separate follow-ups):**
- **5.** Deduplicating `PerformSaslExchangeAsync`/`PerformSaslReauthExchangeAsync` — agreed; it's a focused refactor of the connection's reauth path (which has no integration coverage), so out of scope here.
- **8.** Sequential SCRAM/readiness in the fixtures — test-startup time only; left as-is for now.

Validation: unit 3542/3542 (incl. the new admin tests), Authorization (ACL) 8/8.

PR #946, opened 2026-07-02
fix(consumer): bound prefetch buffer slots

### LINKED ISSUE #895: MpscFetchBuffer lost-wakeup races: StoreLoad reordering + single bool for multiple producers
Found in automated performance/correctness audit (2026-07-02).

## Problem
`src/Dekaf/Consumer/MpscFetchBuffer.cs`:
- Writer (`:60-73`): `Volatile.Write` commit then read of `_consumerWaiting`; reader (`:107-126`): `Volatile.Write` tail then read of `_producerWaiting`. Volatile semantics do not order StoreLoad, so a signal can be missed on both sides. Consumer side is bounded by the 200ms wait timeout (latency blip only).
- `WaitToWriteAsync` (`:80-100`): `_producerWaiting` is a single bool shared by N concurrent broker/connection tasks. Waiter A being released clears the flag in `finally`, masking still-waiting B; B stalls until the buffer fills again or cancellation.

Currently mostly masked because the ring is massively oversized (see the ring-sizing issue) and essentially never fills — but shrinking the ring makes this live. Fix together.

## Suggested fix
Make `_producerWaiting` an `Interlocked` counter (increment before wait, decrement after; reader releases while counter > 0), and add `Interlocked.MemoryBarrier()` between the flag store and the index re-check on both sides.

### LINKED ISSUE #904: perf(consumer): 8 MB LOH prefetch ring per consumer - sized in messages but stores partition-fetches
Found in automated performance audit (2026-07-02).

## Problem
`src/Dekaf/Consumer/KafkaConsumer.cs:694-696` + `MpscFetchBuffer.cs:32-39`: `prefetchCapacity = Math.Max(QueuedMinMessages, 1)` (default 100,000), buffer created with `capacity * 10` = 1,000,000 -> next pow2 = 1,048,576 slots -> `new PendingFetchData?[1048576]` = **8 MB reference array on the LOH per consumer**, allocated unconditionally in the constructor (even when prefetch is later disabled).

Buffer items are `PendingFetchData` (one per partition per FetchResponse, ~1MB payload each), and actual admission control is the byte-based memory limit (default 64MB, i.e. ~64 in-flight items) — so more than 99.99% of slots can never be occupied. Also `prefetchCapacity * 10` can overflow int for extreme option values.

## Suggested fix
Size the ring by expected in-flight fetch items, not messages — e.g. `min(assignedPartitionEstimate * PrefetchPipelineDepth, 1024)`, or derive from `QueuedMaxMessagesKbytes / MaxPartitionFetchBytes`. A few hundred slots is ample; backpressure is already byte-based.

Note: shrinking the ring makes the MpscFetchBuffer lost-wakeup races live — fix together with #895.

thomhurst:
## Summary
- size the MPSC prefetch ring from fetch byte/pipeline bounds instead of QueuedMinMessages
- cap ring slots to avoid the default 1M-slot LOH allocation and extreme option overflow
- replace the single producer-waiting flag with a waiter count
- add StoreLoad barriers around wait rechecks to prevent missed producer/consumer signals
- add regression coverage for slot sizing and multiple producer waiters

Closes #904
Closes #895

## Test plan
- dotnet build tests\\Dekaf.Tests.Unit --configuration Release
- .\\tests\\Dekaf.Tests.Unit\\bin\\Release\\net10.0\\Dekaf.Tests.Unit.exe --treenode-filter /*/*/MpscFetchBufferTests/*
- .\\tests\\Dekaf.Tests.Unit\\bin\\Release\\net10.0\\Dekaf.Tests.Unit.exe --treenode-filter /*/*/PrefetchPipelineRunnerTests/*
- .\\tests\\Dekaf.Tests.Unit\\bin\\Release\\net10.0\\Dekaf.Tests.Unit.exe
- dotnet build tests\\Dekaf.Tests.Integration --configuration Release

### DIFF
--- a/src/Dekaf/Consumer/KafkaConsumer.cs
+++ b/src/Dekaf/Consumer/KafkaConsumer.cs
@@ -691,9 +691,7 @@ public sealed partial class KafkaConsumer<TKey, TValue> : IKafkaConsumer<TKey, T
                     : null);
         }
 
-        // Initialize prefetch buffer - bounded by QueuedMinMessages batches
-        var prefetchCapacity = Math.Max(options.QueuedMinMessages, 1);
-        _prefetchBuffer = new MpscFetchBuffer(prefetchCapacity * 10);
+        _prefetchBuffer = new MpscFetchBuffer(CalculatePrefetchBufferCapacity(options));
 
         // Register this instance's lag callback with the shared static gauge.
         // The callback is invoked only during metric collection (~every 5-60s), not on the hot path.
@@ -1692,6 +1690,18 @@ public sealed partial class KafkaConsumer<TKey, TValue> : IKafkaConsumer<TKey, T
             fetchMaxBytes);
         var minRequired = fetchResponseSize * (pipelineDepth + 1);
         return Math.Max(configuredMax, minRequired);
+    }
+
+    internal static int CalculatePrefetchBufferCapacity(ConsumerOptions options)
+    {
+        ArgumentNullException.ThrowIfNull(options);
+
+        return PoolSizing.ForConsumerPrefetchBuffer(
+            options.QueuedMaxMessagesKbytes,
+            options.MaxPartitionFetchBytes,
+            options.FetchMaxBytes,
+            options.PrefetchPipelineDepth,
+            options.ConnectionsPerBroker);
     }
 
     internal static bool IsFatalPrefetchError(Exception ex) => ex is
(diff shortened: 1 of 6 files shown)


thomhurst:
Addressed the review findings in the latest push:

- moved prefetch buffer capacity sizing into `PoolSizing.ForConsumerPrefetchBuffer`
- switched the pipeline sizing math to `long` and added a pathological-input clamp regression test
- removed the redundant fence after `Interlocked.Increment`

Kept the clamp observability suggestion out of this fix because this path runs at construction time and there is no existing sizing clamp log pattern in `PoolSizing`; current change stays focused on correctness and centralization.

Validation:
- `dotnet build tests\Dekaf.Tests.Unit --configuration Release`
- `Dekaf.Tests.Unit.exe --treenode-filter "/*/*/PoolSizingTests/*"`
- `Dekaf.Tests.Unit.exe --treenode-filter "/*/*/MpscFetchBufferTests/*"`
- `Dekaf.Tests.Unit.exe --treenode-filter "/*/*/PrefetchPipelineRunnerTests/*"`
- full `Dekaf.Tests.Unit.exe`
- `dotnet build tests\Dekaf.Tests.Integration --configuration Release`

PR #950, opened 2026-07-02
perf(protocol): write produce records directly

### LINKED ISSUE #902: perf(protocol): ProduceRequest thread-static staging buffer - full-payload memset via Clear(), unbounded retention, extra memcpy
Found in automated performance audit (2026-07-02).

## Problem
`src/Dekaf/Protocol/Messages/ProduceRequest.cs:87-104, 131-139` stages serialized records in a `[ThreadStatic]` `ArrayBufferWriter<byte>`. Three compounding issues:

1. `buffer.Clear()` (:101) **zeroes the entire previously-written region** — a memset of up to the full serialized records size (potentially MBs) on every reuse, on the send loop thread.
2. Growth allocates `new byte[]` (LOH + zeroing for >85KB) with no size cap and no shrink — under `ConfigureAwait(false)` thread migration, every thread that ever serialized a produce request permanently retains a batch-sized GC array. This is exactly the pathology already diagnosed and fixed for the RecordBatch scratch cache (comment at `RecordBatch.cs:316-321`), but it survives here.
3. The buffer exists only to learn the records byte-length for the `COMPACT_RECORDS` varint prefix; `WriteCompactNullableBytes` (:139) then memcpys the entire payload again into the real output. Combined with the scratch-to-content copy inside `RecordBatch.Write`, produce payload bytes are copied 3x after serialization.

Cost: payload-proportional, per produce request per partition.

## Suggested fix
The records section length is computable upfront: per batch it is `TotalBatchHeaderSize (61) + PreCompressedLength` when pre-compressed (the intended production path), or `61 + sum(VarIntSize(bodySize) + bodySize)` using `Record.CachedBodySize` when uncompressed. Compute the total, `WriteUnsignedVarInt(total + 1)`, then have each `batch.Write` target the output writer directly — eliminates the intermediate buffer, the Clear() memset, the growth allocations, and one full-payload memcpy.

Minimum fix if restructuring is deferred: replace `Clear()` with `ResetWrittenCount()` (no zeroing) and swap `ArrayBufferWriter` for the existing `PooledReusableBufferWriter` + bounded `LockFreeStack` cache pattern from `RecordBatch.cs`.

thomhurst:
## Summary
- compute ProduceRequest COMPACT_RECORDS length before writing payload bytes
- remove the thread-static partition records buffer and write RecordBatch payloads to the final protocol output
- keep KafkaProtocolWriter byte accounting correct for direct payload writes and cover uncompressed/compression fallback paths

Closes #902

## Tests
- `dotnet format --verify-no-changes --include src/Dekaf/Protocol/KafkaProtocolWriter.cs src/Dekaf/Protocol/Records/RecordBatch.cs src/Dekaf/Protocol/Messages/ProduceRequest.cs tests/Dekaf.Tests.Unit/Protocol/ProduceRequestTests.cs --verbosity minimal`
- `dotnet build tests/Dekaf.Tests.Integration --configuration Release`
- `dotnet run --project tests/Dekaf.Tests.Unit --configuration Release --no-build -- --no-ansi --disable-logo --no-progress --maximum-failed-tests 1` (new tests passed; existing producer send-loop timeout failures: `MultiplePartitionsMuted_IndependentlyBlocked_OtherPartitionsProceed`, `InFlightWait_MutedPartitionFilter_MovesCoalescedNormalBatchToCarryOver`, `SendLoop_SequentialRequests_SecondSendsAfterFirstResponseCompletes`)

### DIFF
--- a/src/Dekaf/Protocol/KafkaProtocolWriter.cs
+++ b/src/Dekaf/Protocol/KafkaProtocolWriter.cs
@@ -22,6 +22,11 @@ public ref struct KafkaProtocolWriter
 
     public readonly int BytesWritten => _bytesWritten;
 
+    internal readonly IBufferWriter<byte> BufferWriter => _output;
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal void AddBytesWritten(int byteCount) => _bytesWritten = checked(_bytesWritten + byteCount);
+
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public void WriteInt8(sbyte value)
     {
--- a/src/Dekaf/Protocol/Messages/ProduceRequest.cs
+++ b/src/Dekaf/Protocol/Messages/ProduceRequest.cs
@@ -1,3 +1,4 @@
+using System.Buffers;
 using Dekaf.Compression;
 using Dekaf.Protocol.Records;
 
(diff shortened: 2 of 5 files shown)


thomhurst:
Addressed the review finding in the latest push:

- made `Record.EffectiveHeaderCount` internal and reused it from `RecordBatch.GetEncodedRecordsLength`, so predicted length and `Record.Write` share the same header-count rule
- added coverage for the drift case where `Headers == null` but `HeaderCount` is non-zero
- added coverage for the `CachedBodySize > 0` branch used by producer-created records
- removed the redundant temporary pre-compressed array slot clear

Validation:
- dotnet build tests\Dekaf.Tests.Unit --configuration Release
- Dekaf.Tests.Unit.exe --treenode-filter "/*/*/ProduceRequestTests/*"
- full Dekaf.Tests.Unit.exe: 3589/3589 passed
- dotnet build tests\Dekaf.Tests.Integration --configuration Release -maxcpucount:1 -nodeReuse:false

PR #968, opened 2026-07-02
perf(producer): move cancellation into source

thomhurst:
## Summary
- move awaited-produce cancellation registration into PooledValueTaskSource
- use UnsafeRegister with the pooled source as callback state
- clear cancellation registration before source reset/reuse

## Test plan
- [x] dotnet build tests\Dekaf.Tests.Unit --configuration Release
- [x] .\tests\Dekaf.Tests.Unit\bin\Release\net10.0\Dekaf.Tests.Unit.exe --treenode-filter /*/*/PooledValueTaskSourceTests/*
- [x] .\tests\Dekaf.Tests.Unit\bin\Release\net10.0\Dekaf.Tests.Unit.exe --treenode-filter /*/*/ProducerCancellationTests/*
- [x] .\tests\Dekaf.Tests.Unit\bin\Release\net10.0\Dekaf.Tests.Unit.exe
- [x] dotnet build tests\Dekaf.Tests.Integration --configuration Release
- [x] dotnet format --verify-no-changes --include src\Dekaf\Producer\KafkaProducer.cs src\Dekaf\Producer\PooledValueTaskSource.cs tests\Dekaf.Tests.Unit\Producer\PooledValueTaskSourceTests.cs
- [x] git diff --check origin/main...HEAD

Closes #917

PR #971, opened 2026-07-02
perf(consumer): lazy-copy consumed headers

### LINKED ISSUE #921: perf(consumer): per-record Header[] sub-array allocation via range operator
Found in automated performance audit (2026-07-02). Related: #757 (consumer per-message allocation) — for header-bearing workloads this is likely the dominant per-record allocation.

## Problem
`record.Headers[..record.HeaderCount]` compiles to `RuntimeHelpers.GetSubArray` — a fresh `Header[]` per record (`src/Dekaf/Consumer/KafkaConsumer.cs:1095-1099`, `ConsumeBatch.cs:110-114`). For workloads where every message carries trace/correlation headers this is the dominant per-record allocation in an otherwise allocation-free loop.

The code documents why pooled slices were reverted (use-after-free when callers retain results), so this is a deliberate tradeoff — but it can be improved without reintroducing that bug.

Cost: per-record, only records carrying headers.

## Suggested fix
Defer the copy: a small class wrapping (pooled array, count, batch-generation token) that copies lazily on first access makes the common "user never reads Headers" case allocation-free while keeping retention safe via generation check. Only worth doing if profiling shows header-heavy traffic matters — validate with benchmarks first.

thomhurst:
## Summary
- remove eager `record.Headers[..record.HeaderCount]` copies from consumer result construction
- store pooled header metadata in `ConsumeResult` and create a lazy `IReadOnlyList<Header>` only when `Headers` is read
- snapshot headers on first index/enumeration access and guard stale pooled arrays with `PendingFetchData` generation checks

## Behavior note
Header-bearing records no longer allocate a `Header[]` when callers never read `Headers`. If a caller retains a `ConsumeResult` beyond the owning fetch batch, `Headers` must be read before disposal so the lazy list can snapshot the pooled array; stale first access throws instead of reading returned/cleared pooled memory.

## Validation
- `dotnet build tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release -maxcpucount:1 -nodeReuse:false`
- `dotnet run --project tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release --framework net10.0 --no-build -- --treenode-filter "/*/Dekaf.Tests.Unit.Consumer/ConsumeResultTests/*"`
- `dotnet run --project tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release --framework net10.0 --no-build -- --treenode-filter "/*/Dekaf.Tests.Unit.Consumer/*/*"`
- `dotnet build tests\Dekaf.Tests.Integration\Dekaf.Tests.Integration.csproj --configuration Release -maxcpucount:1 -nodeReuse:false`
- `rg -n "record\.Headers\[\.\.record\.HeaderCount\]|GetSubArray" src\Dekaf tests\Dekaf.Tests.Unit -g "*.cs"`

Closes #921

### DIFF
--- a/src/Dekaf/Consumer/ConsumeBatch.cs
+++ b/src/Dekaf/Consumer/ConsumeBatch.cs
@@ -107,12 +107,6 @@ namespace Dekaf.Consumer
                 long offset = pending.CurrentBaseOffset + record.OffsetDelta;
                 long timestampMs = pending.CurrentBaseTimestamp + record.TimestampDelta;
 
-                IReadOnlyList<Header>? headers = null;
-                if (record.Headers is not null && record.HeaderCount > 0)
-                {
-                    headers = record.Headers[..record.HeaderCount];
-                }
-
                 TimestampType timestampType = pending.CurrentTimestampType;
 
                 int messageBytes = (record.IsKeyNull ? 0 : record.Key.Length) +
@@ -126,7 +120,9 @@ namespace Dekaf.Consumer
                     isKeyNull: record.IsKeyNull,
                     valueData: record.Value,
                     isValueNull: record.IsValueNull,
-                    headers: headers,
+                    pooledHeaders: record.Headers,
+                    pooledHeaderCount: record.HeaderCount,
+                    headerOwner: pending,
                     timestampMs: timestampMs,
                     timestampType: timestampType,
                     leaderEpoch: null,
--- a/src/Dekaf/Consumer/IKafkaConsumer.cs
+++ b/src/Dekaf/Consumer/IKafkaConsumer.cs
@@ -246,6 +246,11 @@ public readonly struct ConsumeResult<TKey, TValue>
     // DateTimeOffset.FromUnixTimeMilliseconds() construction in the consume loop.
     // The DateTimeOffset is computed on demand via the Timestamp property.
     private readonly long _timestampMs;
+    private readonly IReadOnlyList<Header>? _headers;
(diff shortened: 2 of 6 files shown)


thomhurst:
CI fix: rebased onto current `origin/main` and carried over the deterministic `KafkaConnectionTests.SendPipelinedWithCallerTimeoutAsync_ResponseCancellation_ThrowsTimeoutException` harness fix while #968 is still waiting on TLS CI. The failed run threw `TaskCanceledException` before the request reached the response wait; the test now reads the full request frame before cancelling the caller token.

Validation:
- `dotnet build tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release -maxcpucount:1 -nodeReuse:false`
- `dotnet run --project tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release --framework net10.0 --no-build -- --treenode-filter "/*/Dekaf.Tests.Unit.Networking/KafkaConnectionTests/*"`
- `dotnet run --project tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release --framework net10.0 --no-build -- --treenode-filter "/*/Dekaf.Tests.Unit.Consumer/*/*"`
- `dotnet build tests\Dekaf.Tests.Integration\Dekaf.Tests.Integration.csproj --configuration Release -maxcpucount:1 -nodeReuse:false`

PR #985, opened 2026-07-02
perf(networking): use fixed pending table

thomhurst:
## Summary
- replace per-request `ConcurrentDictionary` pending-request entries with a preallocated fixed slot table
- preserve pending request pool return behavior across response, failure, and dispose paths
- add coverage that guards the allocation-free pending table shape

## Verification
- dotnet run --project tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release --framework net10.0 -- --treenode-filter "/*/Dekaf.Tests.Unit.Networking/KafkaConnectionPipelinedTests/PendingRequests_UsesPreallocatedSlotTable" --detailed-stacktrace
- dotnet run --project tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release --framework net10.0 --no-build -- --treenode-filter "/*/Dekaf.Tests.Unit.Networking/*/*" --detailed-stacktrace
- dotnet build tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release -maxcpucount:1 -nodeReuse:false
- dotnet build tests\Dekaf.Tests.Integration\Dekaf.Tests.Integration.csproj --configuration Release -maxcpucount:1 -nodeReuse:false
- dotnet run --project tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release --framework net10.0 --no-build -- --detailed-stacktrace

Closes #927

PR #984, opened 2026-07-02
perf(schemaregistry): write payloads directly

thomhurst:
## Summary
- reuse the Schema Registry payload buffer and Utf8JsonWriter for JSON serialization instead of allocating a payload byte[] per message
- write Protobuf payloads directly into the final wire-format span using CalculateSize/WriteTo
- add JSON wire-format coverage to protect the reusable writer path

Closes #907

## Validation
- `dotnet build tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release`
- `tests\Dekaf.Tests.Unit\bin\Release\net10.0\Dekaf.Tests.Unit.exe --treenode-filter "/*/Dekaf.Tests.Unit.SchemaRegistry/*/*"`
- `tests\Dekaf.Tests.Unit\bin\Release\net10.0\Dekaf.Tests.Unit.exe`
- `dotnet format --verify-no-changes --include src\Dekaf.SchemaRegistry\JsonSchemaSerializer.cs src\Dekaf.SchemaRegistry.Protobuf\ProtobufSchemaRegistrySerializer.cs tests\Dekaf.Tests.Unit\SchemaRegistry\SchemaRegistryCacheTests.cs`
- `git diff --check HEAD~1..HEAD`
- `dotnet build tests\Dekaf.Tests.Integration\Dekaf.Tests.Integration.csproj --configuration Release`

PR #986, opened 2026-07-02
fix(consumer): deflake heartbeat + fetch-buffer wait tests

thomhurst:
## Problem

Two unit tests kept failing intermittently across **unrelated** PRs (e.g. `#985` schema-registry-zero-copy, `fix-980-connected-state`), which is the tell that they are pre-existing flaky tests rather than regressions from those branches:

| Test | Failure |
|------|---------|
| `WaitToWriteAsync_MultipleWaiters_RemainSignaled` | `Expected to be true` (timeout, 7.4s) |
| `ConsumerProtocol_StaticMember_FencedDuringHeartbeat_RejoinsWithEpochNegativeTwo` | wrong state/epoch (9s) |

Both are **timing-dependent tests**, not product race conditions.

## Fixes

**1. `ConsumerProtocol_...FencedDuringHeartbeat...` — remove arbitrary sleep**
The mock signalled its `TaskCompletionSource` *as it returned* the fenced response, i.e. before the heartbeat loop had processed it. The test then slept a fixed `Task.Delay(50)` and asserted the transition — 50ms is not enough on a loaded runner. Now polls the observable outcome (`State == Unjoined && GenerationId == -2`) with a generous timeout. The loop `break`s on `FencedMemberEpoch` (no auto-rejoin), so the state is stable until the explicit rejoin.

**2. `WaitToWriteAsync_MultipleWaiters_RemainSignaled` — bound on one generous token**
Waiter completion depends on the thread pool scheduling `SemaphoreSlim.WaitAsync` continuations, which lags under starvation on 2-core CI. Replaced the short 5s `WhenAny`/`WhenAll` delays with a single 30s outer token.

**3. `MpscFetchBuffer.WaitToWriteAsync` — latent permit leak (bonus)**
When the slow-path recheck sees space, a concurrent `TryRead` has already released a permit for this waiter (it saw `_producerWaiterCount > 0`). Returning without consuming it leaves a stale permit that spuriously wakes a future waiter on a full buffer. Drained with `Wait(0)`. It never blocks and only consumes an already-available permit, so it cannot suppress a real wakeup or stall a producer. Not the CI-failure cause; the caller loops on `TryWrite`, so the pre-existing behaviour was benign (one wasted spin).

## Verification
- Both full test classes pass.
- Both target tests pass **12/12 under 16-process CPU load** (simulating a starved runner).

### DIFF
--- a/src/Dekaf/Consumer/MpscFetchBuffer.cs
+++ b/src/Dekaf/Consumer/MpscFetchBuffer.cs
@@ -101,7 +101,15 @@ internal sealed class MpscFetchBuffer
         {
             // Re-check after setting flag to avoid missed signal
             if (HasSpaceAvailable())
+            {
+                // A concurrent TryRead that freed the slot also released a permit for us
+                // (it saw _producerWaiterCount > 0). Drain it non-blockingly so the permit
+                // does not linger and spuriously wake a future waiter when the buffer is full.
+                // Wait(0) never blocks and only consumes an already-available permit, so it
+                // cannot suppress a genuine wakeup or stall a still-blocked producer.
+                _spaceAvailable.Wait(0, CancellationToken.None);
                 return;
+            }
 
             await _spaceAvailable.WaitAsync(cancellationToken).ConfigureAwait(false);
         }
--- a/src/Dekaf/Consumer/MpscFetchBuffer.cs
+++ b/src/Dekaf/Consumer/MpscFetchBuffer.cs
@@ -105,8 +105,10 @@ internal sealed class MpscFetchBuffer
                 // A concurrent TryRead that freed the slot also released a permit for us
                 // (it saw _producerWaiterCount > 0). Drain it non-blockingly so the permit
                 // does not linger and spuriously wake a future waiter when the buffer is full.
-                // Wait(0) never blocks and only consumes an already-available permit, so it
-                // cannot suppress a genuine wakeup or stall a still-blocked producer.
+                // Wait(0) never blocks. A racing waiter can lose this already-released
+                // permit if it has incremented _producerWaiterCount but not registered with
+                // SemaphoreSlim yet; callers loop through TryWrite, so that costs one extra
+                // wait cycle rather than a hang.
                 _spaceAvailable.Wait(0, CancellationToken.None);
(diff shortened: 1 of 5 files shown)


thomhurst:
Addressed the review feedback:

- Replaced the remaining `Task.Delay(50)` in `ConsumerProtocol_FencedDuringHeartbeat_RejoinsWithEpochZero` with a deterministic poll for `State == Unjoined && GenerationId == 0`.
- Moved the duplicated test polling helper into shared `TestWait` and switched the three local copies to use it.
- Softened the `MpscFetchBuffer.WaitToWriteAsync` drain comment to call out the narrow one-extra-wait-cycle race instead of claiming it cannot affect another waiter.
- Tightened `MpscFetchBufferTests` to assert exactly one waiter completes after the first freed slot.

Validation:
- `dotnet build tests\Dekaf.Tests.Unit --configuration Release`
- `ConsumerCoordinatorKip848Tests` (26/26 passed)
- `MpscFetchBufferTests` (16/16 passed)
- `AppendWorkerAffinityTests` (5/5 passed)

thomhurst:
Addressed the latest review feedback:

- `WaitToWriteAsync` now drains all stale `_spaceAvailable` permits in the recheck branch, not just one.
- Added deterministic coverage for the increment/recheck race window: the test frees two slots after `_producerWaiterCount` increments and verifies no semaphore permits remain leaked.

Validation:
- `dotnet build tests\Dekaf.Tests.Unit --configuration Release`
- `tests\Dekaf.Tests.Unit\bin\Release\net10.0\Dekaf.Tests.Unit.exe --treenode-filter "/*/*/MpscFetchBufferTests/*"` (17/17 passed)

thomhurst:
Rebased onto `origin/main` after #984 merged and resolved the test conflicts by keeping the shared `TestWait` consolidation plus the stronger MPSC waiter/drain assertions.

Validation:
- `dotnet build tests\Dekaf.Tests.Unit --configuration Release`
- `ConsumerCoordinatorKip848Tests` (26/26 passed)
- `MpscFetchBufferTests` (17/17 passed)
- `AppendWorkerAffinityTests` (5/5 passed)

PR #1041, opened 2026-07-02
Make consumer mutators return void

### LINKED ISSUE #1012: api(consumer): stop returning IKafkaConsumer from mutating methods (fake fluency)
## Problem

Consumer mutation methods return `IKafkaConsumer<TKey,TValue>` to enable chaining — `Subscribe` (`src/Dekaf/Consumer/IKafkaConsumer.cs:37,42`), `Unsubscribe` (`:47`), `Assign` (`:52`), `Seek` (`:130`), `SeekToBeginning/End` (`:135,140`), `Pause/Resume` (`:145,150`). These **mutate live consumer state** and have side effects, yet the fluent return type makes them look like builder configuration. Meanwhile `GetPosition` returns `long?` (`:125`) — so the "fluent" convention is applied inconsistently.

This blurs two distinct concepts: a **builder** (configures, returns self) vs a **live object** (does work, has effects). Chaining `consumer.Seek(a).Pause(b)` reads like pure configuration but is a sequence of stateful operations.

## Proposed change

- Mutating consumer operations return `void` (or `Task`/`ValueTask` where the operation is genuinely async).
- Reserve fluent self-return for the builder only.

```csharp
consumer.Seek(tpo);
consumer.Pause(tp);
// not: consumer.Seek(tpo).Pause(tp);
```

## Why

Keeps the mental model clean: builder = config, consumer = live. Removes a subtle source of confusion and matches how `System.*` live objects behave.

### Docs (definition of done)
Update any chained `consumer.X().Y()` examples to sequential statements, with before→after snippets:
- `docs/docs/consumer/basics.md`
- `docs/docs/consumer/manual-assignment.md`
- `docs/docs/consumer/offset-management.md`
Docs must not ship stale syntax; doc updates are part of this issue's completion.

thomhurst:
Closes #1012

## Summary
- change live consumer mutators from fluent self-return to void
- update consumer implementation and unit tests that expected chaining
- document sequential command style with before/after snippets in consumer docs

## Validation
- dotnet build Dekaf.sln --configuration Release --no-restore -m:1 -p:UseSharedCompilation=false
- dotnet run --project tests\\Dekaf.Tests.Unit\\Dekaf.Tests.Unit.csproj --configuration Release --no-build -- --disable-logo --no-progress

### DIFF
--- a/docs/docs/consumer/basics.md
+++ b/docs/docs/consumer/basics.md
@@ -42,6 +42,17 @@ var consumer = await Kafka.CreateConsumer<string, string>()
     .BuildAsync();
 ```
 
+Consumer mutation methods are commands and return `void`. Keep live consumer operations as separate statements:
+
+```csharp
+// Before
+consumer.Subscribe("my-topic").Pause(new TopicPartition("my-topic", 0));
+
+// After
+consumer.Subscribe("my-topic");
+consumer.Pause(new TopicPartition("my-topic", 0));
+```
+
 ## Consuming Messages
 
 The primary way to consume is with `await foreach`:
--- a/docs/docs/consumer/manual-assignment.md
+++ b/docs/docs/consumer/manual-assignment.md
@@ -112,6 +112,18 @@ consumer.IncrementalUnassign(new[]
 });
 ```
 
+Assignment and seek methods mutate a live consumer and return `void`, so write them as sequential commands:
+
+```csharp
+// Before
+consumer.Assign(new TopicPartition("my-topic", 0))
+    .SeekToBeginning(new TopicPartition("my-topic", 0));
+
+// After
+consumer.Assign(new TopicPartition("my-topic", 0));
+consumer.SeekToBeginning(new TopicPartition("my-topic", 0));
+```
+
 ## Seeking
 
 With manual assignment, you can freely seek to any offset:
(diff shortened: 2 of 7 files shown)


thomhurst:
Fixed CI build failure after rebasing on latest main:

- explicit `logger: null` on admin test `AdminClient` constructors so overload resolution chooses the typed logger overload instead of becoming ambiguous with the new `ILoggerFactory?` overload.

Validation:
- `dotnet build tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release --no-restore -m:1 -p:UseSharedCompilation=false`
- `Dekaf.Tests.Unit.exe --treenode-filter "/*/*/AdminClientConsumerGroupDescribeTests/*|/*/*/AdminDescribeTopicPartitionsTests/*" --no-ansi --output Normal`
- `git diff --check`

thomhurst:
Addressed the review nit: renamed the pattern subscription test to reflect that explicit subscription/assignment stay empty until broker metadata resolves the pattern. Also rebased onto latest main, which dropped the obsolete admin logger disambiguation commit after the constructor overload was removed.

thomhurst:
Producer integration failure appears unrelated to this PR's consumer mutator API change. Failed test: Producer_WithStickyPartitioner_SticksMessagesBeforeBatchCompletes, assertion partitions.Count expected 1 but found 2. This is timing-sensitive sticky-partitioner behavior; all unit/build/code-quality and other completed integration buckets passed. I tried gh run rerun --failed, but GitHub rejected it because the workflow is still running. Next loop pass can rerun failed jobs once the workflow completes.

thomhurst:
Workflow is complete now, so I reran the failed Producer integration job once. Same failure was \\Producer_WithStickyPartitioner_SticksMessagesBeforeBatchCompletes\\ (expected one partition, observed two), which is outside this PR's consumer mutator API diff.

PR #1051, opened 2026-07-03
feat(api): add KafkaClient root

### LINKED ISSUE #1017: api: introduce KafkaClient root owning connection pool + memory budget
## Problem

`ProducerBuilder.Build()` couples every producer to a **process-wide mutable static**: `DekafMemoryBudget.RegisterProducer(producer)` / `ReserveExplicit(...)` (`src/Dekaf/Builders.cs:736-739`). The long comment at `:727-735` describes subtle ordering hazards this global coupling creates. Each `Kafka.CreateProducer(...)` also builds its own independent connection pool, so producers and consumers to the same cluster share nothing except this hidden global.

Consequences: hard to test in isolation, surprising cross-instance coupling, and two logically-independent clusters in one process contend on the same static budget.

## Proposed change

Introduce a root client object that explicitly owns the connection pool and memory budget; hang producers/consumers off it:

```csharp
await using var client = Kafka.Connect("localhost:9092"); // owns pool + memory budget
var producer = client.CreateProducer<string,string>();
var consumer = client.CreateConsumer<string,string>(groupId: "g");
```

- Shared resources become explicit and disposable (no static lifetime).
- Connection reuse across producer + consumer to the same cluster becomes natural.
- The current `Kafka.CreateProducer(...)`/builder entry points can remain as thin sugar that spin up a single-use implicit client, preserving the simple case.
- Also gives DI a clean singleton to own (`services.AddSingleton(_ => Kafka.Connect(...))`) and resolves the ownership question in the keyed-clients work.

## Why

Removes global mutable state, makes resource ownership explicit and testable, and enables cross-client connection/memory sharing by design rather than by hidden static.

### Docs (definition of done)
Introduce the root-client model with before→after snippets; explain when to share a client:
- `docs/docs/getting-started.md`
- `docs/docs/intro.md`
- `docs/docs/producer/basics.md`, `docs/docs/consumer/basics.md`
- `docs/docs/dependency-injection.md`
Docs must not ship stale syntax; doc updates are part of this issue's completion.

thomhurst:
Closes #1017

## Summary
- add `KafkaClient` / `KafkaClientBuilder` via `Kafka.Connect(...)` as a root client for shared cluster infrastructure
- route root-created producers, consumers, share consumers, and admin clients through one connection pool, metadata manager, and per-root memory budget
- keep child disposal from tearing down root-owned infrastructure and document root-client usage
- cover sharing, disposal, bootstrap ownership, and independent root memory budgeting in unit tests

## Test plan
- `dotnet build tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release --no-restore -m:1 -p:UseSharedCompilation=false`
- `tests\Dekaf.Tests.Unit\bin\Release\net10.0\Dekaf.Tests.Unit.exe --treenode-filter "/*/*/KafkaClientBuilderTests/*" --no-ansi --output Normal`
- `dotnet build tests\Dekaf.Tests.Integration\Dekaf.Tests.Integration.csproj --configuration Release --no-restore -m:1 -p:UseSharedCompilation=false`

Note: a full unit executable run was attempted earlier and did not finish within 5 minutes, so it was stopped without a summary. Focused root-client tests and related builder/admin focused tests passed locally.

### DIFF
--- a/docs/docs/consumer/basics.md
+++ b/docs/docs/consumer/basics.md
@@ -21,6 +21,16 @@ await using var consumer = await Kafka.CreateConsumer<string, string>()
 
 The type parameters `<TKey, TValue>` define the expected key and value types.
 
+For services with multiple Kafka clients, create a root client once and build consumers from it. Bootstrap servers, TLS/SASL, and memory budgeting are owned by the root; group IDs and subscription settings stay on the consumer builder.
+
+```csharp
+await using var kafka = Kafka.Connect("localhost:9092");
+
+await using var consumer = await kafka.CreateConsumer<string, string>("my-consumer-group")
+    .SubscribeTo("my-topic")
+    .BuildAsync();
+```
+
 ## Subscribing to Topics
 
 Before consuming, subscribe to one or more topics:
--- a/docs/docs/dependency-injection.md
+++ b/docs/docs/dependency-injection.md
@@ -6,6 +6,8 @@ sidebar_position: 11
 
 Dekaf plays nicely with ASP.NET Core's DI container. Here's how to wire it up.
 
+DI registration owns each registered client for you. For manual wiring outside DI, prefer `Kafka.Connect(...)` when multiple clients should share a connection pool and memory budget.
+
 ## Installation
 
 ```bash
--- a/docs/docs/getting-started.md
+++ b/docs/docs/getting-started.md
@@ -41,6 +41,7 @@ var producer = await Kafka.CreateProducer<string, string>()
 
 The `using Dekaf;` directive gives you access to:
 - The static `Kafka` class for creating producers and consumers
+- The `KafkaClient` root for sharing one connection pool and memory budget across related clients
 - Common types like `Headers`, `TopicPartition`, `TopicPartitionOffset`
 - Extension methods for consumers and producers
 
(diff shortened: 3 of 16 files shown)


thomhurst:
Addressed the actionable review items and the failing unit test:

- Changed MetadataManager broker-count notification to multicast callbacks so multiple root-created producers do not clobber each other's pool ratchets.
- Sized the root client's shared response buffer pool for the high-throughput adaptive fetch ceiling instead of the old fixed 50 MiB cap.
- Made root-created producer/consumer/share-consumer/admin builders reject root-owned connection settings (TLS, SASL, socket buffers, connection counts) instead of silently accepting ignored child settings.
- Fixed the reproduced `ConsumerLeaderDiscoveryTests` disposal failure: the test-injected infrastructure constructor owns injected dependencies again; the root-client constructor path still keeps shared infrastructure unowned by child consumers.

Validation:

- `dotnet build tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release --no-restore -m:1 -p:UseSharedCompilation=false`
- `Dekaf.Tests.Unit.exe --no-progress --treenode-filter "/*/*/KafkaClientBuilderTests/*"` (7/7)
- `Dekaf.Tests.Unit.exe --no-progress --treenode-filter "/*/*/ConsumerLeaderDiscoveryTests/*"` (5/5)
- `git diff --check`

thomhurst:
Addressed the follow-up review leak:

- `MetadataManager.AddBrokerCountDiscoveredCallback` now returns an unsubscribe token.
- `KafkaProducer` stores that registration and disposes it unconditionally during producer disposal, including root-created producers where shared metadata/pool are not owned by the producer.
- Added coverage that a root-created producer removes its broker-count callback when disposed.

Validation:

- `dotnet build tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release --no-restore -m:1 -p:UseSharedCompilation=false`
- `Dekaf.Tests.Unit.exe --no-progress --treenode-filter "/*/*/KafkaClientBuilderTests/*"`
- `Dekaf.Tests.Unit.exe --no-progress --treenode-filter "/*/*/KafkaClientBuilderTests/RootClient_CreatedProducer_DisposeRemovesBrokerCountCallback"`
- `git diff --check`

PR #1078, opened 2026-07-03
fix(api): harden KafkaClient metadata sharing

thomhurst:
## Summary
- make `MetadataManager.InitializeAsync()` idempotent for shared KafkaClient infrastructure
- reject root-owned metadata setting overrides on Producer, Consumer, and Admin builders
- add unit coverage for concurrent metadata initialization and root-owned metadata guards

## Test plan
- [x] dotnet build tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release --no-restore -m:1 -p:UseSharedCompilation=false
- [x] .\tests\Dekaf.Tests.Unit\bin\Release\net10.0\Dekaf.Tests.Unit.exe --no-progress --treenode-filter "/*/*/MetadataManagerTests/*"
- [x] .\tests\Dekaf.Tests.Unit\bin\Release\net10.0\Dekaf.Tests.Unit.exe --no-progress --treenode-filter "/*/*/KafkaClientBuilderTests/*"
- [x] .\tests\Dekaf.Tests.Unit\bin\Release\net10.0\Dekaf.Tests.Unit.exe --no-progress --treenode-filter "/*/*/MetadataRecoveryStrategyTests/*"
- [x] .\tests\Dekaf.Tests.Unit\bin\Release\net10.0\Dekaf.Tests.Unit.exe --no-progress --treenode-filter "/*/*/MetadataMaxAgeBuilderTests/*"
- [x] git diff --check origin/main..HEAD

Follow-up to #1051.
Closes #1077

PR #1113, opened 2026-07-03
test: deflake consumer/producer timing tests under CI thread-pool starvation

thomhurst:
## Problem

Unit tests failed intermittently across many recent CI runs. Investigating six recent failed runs (mostly on `main`, which should always be green) surfaced a recurring set of flaky tests:

| Test | Occurrences | Symptom |
|---|---|---|
| `RunAsync_PipelineDepth2_OneEagerFetch` | 4× | assert `fetchCount>=3` got `1` |
| `WaitToReadAsync_DataArrivesAfterWait_ReturnsTrue` | 2× | `TaskCanceled [Timeout]` |
| `WaitToWriteAsync_MultipleWaiters_RemainSignaled` | 1× | `TaskCanceled [Timeout]` |
| `ConsumerProtocol_FencedDuringHeartbeat_RejoinsWithEpochZero` | 1× | `Timeout` (~34s) |
| `HandleNotLeaderOrFollower_WithoutInlineLeader_...` | 1× | `Timeout` (~41s) |
| `SendLoop_MultipleInFlight_AllResponsesProcessed` | 1× | hung, killed at 120s |

## Root cause

All share one mechanism: each test gates a **background/async continuation** with an arbitrary **wall-clock cap** (or loops until a wall-clock `cts` fires). On a loaded CI runner the thread pool is briefly starved — some parallel tests block pool threads — so those continuations aren't scheduled within the cap. The wait then times out (or the loop stalls early) even though the awaited signal is deterministically guaranteed. This is exactly the "thread-pool starvation on slow CI runners / arbitrary timeouts that flake" failure mode called out in `CLAUDE.md`.

Two of these (the `SendLoop` 120s hang and `PipelineDepth2` 30–53s busy loop) were themselves *hogging* the pool, amplifying starvation for the coordinator/discovery timeout tests in the same runs.

The fix per project policy: rely on the deterministic completion signals each test already has and drop the arbitrary caps, leaving TUnit's test-level timeout / hang dump as the sole backstop for a genuine regression.

## Changes (test-only)

- **PrefetchPipelineRunnerTests** (Depth1/Depth2): terminate the loop by **self-cancelling after N fetches** instead of running until a 5s/30s `cts`. Under starvation the `Task.Yield` continuation could stall past the deadline, leaving `fetchCount=1`. Matches the already-migrated Depth4 test. Also removes ~35s of pure wall-clock waiting from the suite.
- **MpscFetchBufferTests** (`DataArrivesAfterWait`, `MultipleWaiters_RemainSignaled`): drop the 10s/30s `cts` that raced the `RunContinuationsAsynchronously` waiter / `SemaphoreSlim` release continuations; the `TryWrite`/`TryRead` signals are deterministic.
- **BrokerSenderSendLoopTests** (`MultipleInFlight`): give each in-flight request its **own** `ProduceResponse`. The shared pooled instance was blanked (`TopicCount=0`) by the first request's `Return()` when batches coalesced into >1 request, leaving the rest unacknowledged until the 120s timeout. (No product change — real usage always has one deserialized response per request.)
- **ConsumerCoordinatorKip848Tests** (`Fenced`/`StaticMember` rejoin): drop the 30s wall-clock caps; the `fencingProcessed` signal and the state-transition poll are guaranteed.
- **ConsumerLeaderDiscoveryTests**: await the pending leader-refresh task directly in the shared drain helper instead of a 5s `WaitAsync`; every caller signals completion first (fixes all four tests sharing the helper).

All classified as **test-only** timing issues; no production code changed. The `IdempotentMultiConnection_PerPartitionOrdering_Preserved` integration flake seen in one run was already fixed at HEAD (#1078) and needs no change.

## Verification

- Full unit suite: **3833/3833 passing**, 0 failed.
- Each affected class run repeatedly (3–5×) locally: all green, and now complete promptly (no more 5s/30s/120s wall-clock waits).

### DIFF
--- a/tests/Dekaf.Tests.Integration/RealWorld/ProducerDeliveryGuaranteeTests.cs
+++ b/tests/Dekaf.Tests.Integration/RealWorld/ProducerDeliveryGuaranteeTests.cs
@@ -51,6 +51,25 @@ public sealed class ProducerDeliveryGuaranteeTests(KafkaTestContainer kafka) : K
     {
         var topic = await KafkaContainer.CreateTestTopicAsync();
 
+        // Acks.None is fire-and-forget: the broker sends no ack and the producer cannot
+        // retry, so a message sent before the partition leader is fully ready is silently
+        // lost — the root cause of this test's intermittent "message not delivered" flake.
+        // Durably warm up the partition first (a separate Acks.All producer retries until
+        // the leader accepts) so the fire-and-forget send below reliably lands.
+        await using (var warmupProducer = await Kafka.CreateProducer<string, string>()
+            .WithBootstrapServers(KafkaContainer.BootstrapServers)
+            .WithAcks(Acks.All)
+            .WithLoggerFactory(GlobalTestSetup.GetLoggerFactory())
+            .BuildAsync())
+        {
+            await warmupProducer.ProduceAsync(new ProducerMessage<string, string>
+            {
+                Topic = topic,
+                Key = "warmup",
+                Value = "warmup"
+            }, CancellationToken.None);
+        }
+
         await using var producer = await Kafka.CreateProducer<string, string>()
             .WithBootstrapServers(KafkaContainer.BootstrapServers)
             .WithAcks(Acks.None)
@@ -76,7 +95,13 @@ public sealed class ProducerDeliveryGuaranteeTests(KafkaTestContainer kafka) : K
 
         consumer.Assign(new TopicPartition(topic, 0));
 
-        var result = await consumer.ConsumeOneAsync(TimeSpan.FromSeconds(30));
+        // Skip the durable warmup message and assert the fire-and-forget message arrived.
+        ConsumeResult<string, string>? result;
+        do
+        {
+            result = await consumer.ConsumeOneAsync(TimeSpan.FromSeconds(30));
+        }
+        while (result is not null && result.Value.Value == "warmup");
 
         await Assert.That(result).IsNotNull();
         await Assert.That(result!.Value.Value).IsEqualTo("fast-value");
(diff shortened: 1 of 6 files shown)


thomhurst:
Producer integration failed once in CI on `AcksNone_FireAndForget_MessageDelivered` with a null consume result after 30s. This PR only changes unit tests, and the focused integration test passed locally on the unchanged branch:

- `dotnet build tests\Dekaf.Tests.Integration\Dekaf.Tests.Integration.csproj --configuration Release --no-restore -m:1 -p:UseSharedCompilation=false`
- `tests\Dekaf.Tests.Integration\bin\Release\net10.0\Dekaf.Tests.Integration.exe --treenode-filter "/*/*/ProducerDeliveryGuaranteeTests/AcksNone_FireAndForget_MessageDelivered" --hangdump --hangdump-timeout 15m --log-level Debug` (3/3 passed)

I reran the failed CI jobs once via `gh run rerun 28645324103 --failed`.

thomhurst:
Producer integration failed once in CI on `AcksNone_FireAndForget_MessageDelivered` with a null consume result after 30s. This PR only changes unit tests, and the focused integration test passed locally on the unchanged branch:

- `dotnet build tests\Dekaf.Tests.Integration\Dekaf.Tests.Integration.csproj --configuration Release --no-restore -m:1 -p:UseSharedCompilation=false`
- `tests\Dekaf.Tests.Integration\bin\Release\net10.0\Dekaf.Tests.Integration.exe --treenode-filter "/*/*/ProducerDeliveryGuaranteeTests/AcksNone_FireAndForget_MessageDelivered" --hangdump --hangdump-timeout 15m --log-level Debug` (3/3 passed)

I reran the failed CI jobs once via `gh run rerun 28645324103 --failed`.

PR #1143, opened 2026-07-03
Move producer record encoding to append time

### LINKED ISSUE #1136: Producer: encode records to wire format at append time (eliminate send-loop re-encode)
## Context

Investigation of the single-broker fire-and-forget stress test gap (Dekaf 1.29M msg/s vs Confluent 1.42M) found Dekaf is pipeline-bound, not CPU-bound: the client uses only 1.45 of 2 pinned cores (vs 1.76 for librdkafka) while the per-broker send-loop thread saturates.

## Problem

The `BatchArena` stores raw key/value bytes, not wire format (`RecordAccumulator.TryAppendFromSpans`). At send time, the single per-broker send-loop thread (`BrokerSender.SendLoopAsync`) does all of this inline, serially with the socket write:

1. Varint-encode every record into a scratch buffer (`RecordBatch.Write` -> `WriteRecords`)
2. Copy the scratch buffer into the request buffer
3. CRC32C pass over the whole batch
4. `Stream.WriteAsync` syscall

That is ~4 memory passes per byte on one thread at 1.2+ GB/s, with no serialize/IO overlap on a connection (`serialize R_n -> write R_n -> serialize R_n+1`). This is the structural per-connection throughput ceiling (~1.29M msg/s at 1KB messages), while spare CPU sits idle on the app-thread cores.

Note the compressed path already avoids this: `PreCompress` runs at seal time on the app thread, so the send loop only blits + CRCs. `CompressionType.None` early-returns and dumps all encode work on the send thread. An interim fix (seal-time pre-encode for uncompressed) moves the encode off the send loop but still pays an extra full copy and an extra pooled buffer per batch.

## Proposed long-term fix

Encode records directly into wire format in the arena at append time, Java-client style (`MemoryRecordsBuilder`):

- App thread appends the record already varint-encoded into the batch arena buffer (per-record encode cost moves to app threads, which have idle headroom and scale with producer concurrency)
- Sealing becomes free (the arena content **is** the records section of the batch)
- Send loop work per request drops to: write the 61-byte batch header, CRC32C over the contiguous span, one socket write
- Opens the door to incremental CRC at append time if the header layout allows

## Complications to design for

- Records reference a timestamp delta and offset delta against batch base values fixed at batch open; append-time encode must pick base timestamp from the first record (already done today)
- Aborting/evicting an open batch must not leak partially encoded state
- The compressed path should compose: compress the already-encoded arena span at seal time (removes the `WriteRecords` step inside `PreCompress` too)
- Batch size estimation (`BufferMemory` accounting, `MaxRequestSize` drain logic) should use exact encoded size, which becomes simpler since the encoded size is known at append time

## Expected impact

Removes the primary structural ceiling on per-connection produce throughput; benchmark target is closing the remaining single-broker fire-and-forget gap vs librdkafka and raising the ceiling for all producer scenarios.

Related interim work: PR removing the stress-harness `SocketSendBufferBytes` asymmetry and adding seal-time pre-encode for uncompressed batches.

thomhurst:
## Summary

Moves producer record wire encoding into the append path so sealed batches carry arena-backed encoded record bytes. `RecordBatch` can now write or compress those pre-encoded bytes directly, avoiding record re-encoding on the send path for `CompressionType.None` and compressing the existing encoded payload when compression is enabled.

## Root Cause

Producer batches previously retained logical record state and serialized it later during send preparation/write. That meant uncompressed sends still paid record encoding work on the send path, and compressed sends encoded records into scratch memory before compression.

## Changes

- Encode each appended record directly into the `BatchArena` using Kafka record wire format.
- Let sealed `RecordBatch` instances reference pre-encoded arena bytes and skip `PreCompress(None)` work.
- Compress pre-encoded arena bytes directly for compressed sends.
- Track exact encoded record bytes for batch sizing and expose full encoded batch size for `MaxRequestSize` drain accounting.
- Return raw pooled key/value/header inputs immediately after append-time encoding.
- Add/update producer tests for pre-encoded batches, pooling expectations, and slow-path memory accounting.

Closes #1136.

## Validation

- `dotnet build tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj /p:RunAnalyzers=false`
- `tests\Dekaf.Tests.Unit\bin\Debug\net10.0\Dekaf.Tests.Unit.exe --treenode-filter "/*/*/RecordAccumulatorTests/*"`
- `tests\Dekaf.Tests.Unit\bin\Debug\net10.0\Dekaf.Tests.Unit.exe --treenode-filter "/*/*/RecordBatchTests/*"`
- `tests\Dekaf.Tests.Unit\bin\Debug\net10.0\Dekaf.Tests.Unit.exe --treenode-filter "/*/*/ProduceRequestTests/*"`
- `tests\Dekaf.Tests.Unit\bin\Debug\net10.0\Dekaf.Tests.Unit.exe --treenode-filter "/*/*/RecordAccumulatorReadyTests/*"`
- `tests\Dekaf.Tests.Unit\bin\Debug\net10.0\Dekaf.Tests.Unit.exe`

### DIFF
--- a/src/Dekaf/Builders.cs
+++ b/src/Dekaf/Builders.cs
@@ -861,6 +861,8 @@ public sealed class ProducerBuilder<TKey, TValue>
                 $"MaxConnectionsPerBroker ({_maxConnectionsPerBroker}) must be >= ConnectionsPerBroker ({_connectionsPerBroker}). " +
                 $"Adaptive scaling would be permanently disabled since the initial connection count already exceeds the maximum.");
 
+        ProducerOptions.ValidateArenaCapacity(_batchSize, _arenaCapacity);
+
         // Java Kafka client enforces acks=all when enable.idempotence=true.
         // With acks=leader, the leader acknowledges before ISR replication completes,
         // which can cause OutOfOrderSequenceNumber on leader failover and makes the
--- a/src/Dekaf/Producer/ProducerOptions.cs
+++ b/src/Dekaf/Producer/ProducerOptions.cs
@@ -320,12 +320,37 @@ public sealed class ProducerOptions
     /// Capacity in bytes of the per-batch arena that records are encoded into at append time.
     /// Default is 0, which uses BatchSize + 12.5% margin as the arena capacity — the margin
     /// provides headroom so a record appended near the batch-size boundary still fits.
-    /// Values below the default are ignored (records encode directly into the arena, so a
+    /// Values below the default are rejected (records encode directly into the arena, so a
     /// smaller arena would silently cap batch fill below BatchSize). Set a larger value for
     /// workloads where individual records regularly exceed BatchSize, to avoid re-renting a
     /// larger arena for those batches.
     /// </summary>
     public int ArenaCapacity { get; init; } = 0;
+
+    internal const int ArenaOverflowMarginDivisor = 8;
+
+    internal static int GetMinimumArenaCapacity(int batchSize) =>
+        (int)Math.Min((long)batchSize + batchSize / ArenaOverflowMarginDivisor, int.MaxValue);
+
+    internal static int GetEffectiveArenaCapacity(int batchSize, int arenaCapacity)
+    {
+        var minimum = GetMinimumArenaCapacity(batchSize);
+        return arenaCapacity > 0 ? arenaCapacity : minimum;
+    }
+
+    internal static void ValidateArenaCapacity(int batchSize, int arenaCapacity)
+    {
+        if (arenaCapacity == 0)
+            return;
+
+        var minimum = GetMinimumArenaCapacity(batchSize);
+        if (arenaCapacity < minimum)
+        {
+            throw new InvalidOperationException(
+                $"ArenaCapacity ({arenaCapacity}) must be 0 or at least {minimum} bytes for BatchSize {batchSize}. " +
+                "Records are encoded directly into the arena, so smaller arenas can cap batch fill below BatchSize.");
+        }
+    }
 
     /// <summary>
     /// Initial capacity for record arrays in partition batches.
--- a/src/Dekaf/Producer/ProducerOptions.cs
+++ b/src/Dekaf/Producer/ProducerOptions.cs
@@ -317,13 +317,13 @@ public sealed class ProducerOptions
     public int ValueTaskSourcePoolSize { get; init; }
 
     /// <summary>
-    /// Capacity of the arena buffer for zero-copy serialization in bytes.
-    /// Larger arenas reduce fallback allocations when messages are larger than expected.
-    /// Default is 0, which uses BatchSize + 12.5% overflow margin as the arena capacity.
-    /// The margin provides headroom so that messages appended near the end of a batch
-    /// can still be serialized in the arena instead of falling back to ArrayPool.
-    /// For workloads with very large messages (approaching BatchSize), consider setting
-    /// this to BatchSize * 2 to reduce slow-path fallbacks.
+    /// Capacity in bytes of the per-batch arena that records are encoded into at append time.
+    /// Default is 0, which uses BatchSize + 12.5% margin as the arena capacity — the margin
+    /// provides headroom so a record appended near the batch-size boundary still fits.
+    /// Values below the default are ignored (records encode directly into the arena, so a
+    /// smaller arena would silently cap batch fill below BatchSize). Set a larger value for
+    /// workloads where individual records regularly exceed BatchSize, to avoid re-renting a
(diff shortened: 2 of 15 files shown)


thomhurst:
Pushed a follow-up commit (939f686) tightening this up:

**Encoder moved into the protocol layer.** The append-time encoder was a hand-rolled duplicate of the record wire format living in `RecordAccumulator` (varint writers, header encoding). It has to stay byte-exact with `Record.ComputeBodySize` or the arena contents silently corrupt in Release builds, so the two now live side by side: `Record.Encode` / `Header.Encode` next to their size counterparts. Also restored the SIMD `Ascii.IsValid` fast path (dropped from the old estimate helper) in `Header.CalculateSize`.

**Dead machinery removed.** Append-time encoding left three per-batch arrays rented, grown, transferred, and recycled but never written: the `Record[]` working array, the `byte[][]` key/value tracking array, and the `Header[][]` tracking array. Deleted them plus their dedicated ArrayPools, `RecordListWrapper`, the `ReusableArrays` wrapper (reuse queue now recycles just the completion-sources array — also removes its per-batch wrapper allocation), and the discarded `estimatedRecordSize` append parameter. Net −418 lines.

**`ArenaCapacity` semantics.** With no ArrayPool fallback anymore, a user-set `ArenaCapacity` < `BatchSize` would silently cap batch fill. Now clamped to the BatchSize-derived default (option can only enlarge), docs updated.

**Benchmark fix.** `AccumulatorAppendBenchmarks` crashed with `InvalidOperationException` whenever the drainer fell behind — `GetAwaiter().GetResult()` on an incomplete pooled `PendingAppend` ValueTask is invalid (reproducible on this branch before my changes). Fixed to convert to a `Task` before blocking.

**Validation**
- Full unit suite: 3836/3836 pass
- Full solution Release build clean
- `AccumulatorAppendBenchmarks` before/after this commit (same machine, same fixed harness): means within error bars, hot path allocation-free on both

| Benchmark | Base (ad9f861) | This commit |
|---|---|---|
| AppendHotPath 100B | 57.9 ns / 4 B | 54.7 ns / 21 B |
| AppendMultiPartition 100B | 63.3 ns / 0 B | 79.3 ns / 0 B |
| AppendHotPath 1000B | 126.1 ns / 438 B | 151.9 ns / 364 B |
| AppendMultiPartition 1000B | 100.0 ns / 135 B | 102.2 ns / 131 B |

(The non-zero Allocated values are stochastic cold-path engagements when the benchmark drainer falls behind — present on both sides; the 0 B rows show the hot path itself stays allocation-free. 3-iteration runs, error bars ±14–175 ns.)

thomhurst:
Pushed follow-up commit 05ea145 addressing the review findings:

- Failed append-time encoding now releases the BufferMemory reservation and returns pooled key/value/header resources before rethrowing. The append worker only does fallback cleanup when `AppendAsync` throws before returning a `ValueTask`, so delayed failures are not double-returned.
- `Record.Encode` now enforces the computed-size invariant in Release builds instead of relying only on `Debug.Assert`.
- Explicit `ArenaCapacity` values below the BatchSize-derived minimum now fail at build/accumulator construction instead of being silently clamped.
- Oversized first-record arenas carry a normal pool limit and are dropped on return, so one-off huge POH buffers do not remain in the static pool.
- Added unit coverage for undersized `ArenaCapacity` validation and oversized arena drop behavior.

Validation:
- `dotnet build tests/Dekaf.Tests.Unit/Dekaf.Tests.Unit.csproj -c Release`
- `tests/Dekaf.Tests.Unit/bin/Release/net10.0/Dekaf.Tests.Unit.exe` (3863/3863 passed)

thomhurst:
Amended the follow-up as e666314 after the simplify pass; same fixes and validation, with the append-loop diff narrowed to the failing append helpers.

thomhurst:
Addressed the remaining append-time encoding review feedback.

Changes:
- Moved failed append cleanup ownership into the reservation cores so pooled key/value/header resources and pending-count/memory reservations are released once on append exceptions.
- Kept worker cancellation cleanup only for cancellation before AppendAsync takes ownership.
- Added regression coverage for arena-full non-first append rotation and Record.Encode failure rewinding the last arena allocation.

Validation:
- dotnet build tests/Dekaf.Tests.Unit/Dekaf.Tests.Unit.csproj --configuration Release
- tests\Dekaf.Tests.Unit\bin\Release\net10.0\Dekaf.Tests.Unit.exe --treenode-filter "/*/*/RecordAccumulatorTests/*"
- tests\Dekaf.Tests.Unit\bin\Release\net10.0\Dekaf.Tests.Unit.exe

PR #1164, opened 2026-07-03
Track consumer leader epochs

### LINKED ISSUE #1150: feat(consumer): log truncation detection via leader epochs (KIP-320)
## Summary
The consumer has no log truncation detection. After an unclean leader election, it can silently skip records or re-read data instead of surfacing truncation.

Current state:
- `FetchRequest.CurrentLeaderEpoch` and `LastFetchedEpoch` exist on the wire (`src/Dekaf/Protocol/Messages/FetchRequest.cs:198,208`) but the consumer always sends `-1` (`src/Dekaf/Consumer/KafkaConsumer.cs:2992,3454,4677` — epochs are never tracked)
- `FetchResponse.DivergingEpoch` is parsed (`src/Dekaf/Protocol/Messages/FetchResponse.cs:333,482`) but never read anywhere in `src/Dekaf/Consumer` (zero references)
- `ListOffsetsRequest.CurrentLeaderEpoch` similarly always `-1`

## Scope (Kafka 4.0+ / Fetch v12+, no legacy paths needed)
- Track the leader epoch of the last consumed record per partition (record batches carry `PartitionLeaderEpoch`)
- Track current leader epoch from metadata per partition
- Send both in `FetchRequest`; send `CurrentLeaderEpoch` in `ListOffsets`
- Handle `DivergingEpoch` in fetch responses: surface truncation to the application (exception or configurable reset), and use the returned `(epoch, endOffset)` to reposition
- Handle `FENCED_LEADER_EPOCH` / `UNKNOWN_LEADER_EPOCH` errors by refreshing metadata and retrying
- Include leader epoch in committed offsets (`OffsetCommitRequest` supports it) so position validation survives restarts

Note: because we are Kafka 4.0+ only, the fetch-based diverging-epoch path covers truncation detection — the separate `OffsetsForLeaderEpoch` (API 23) request is not required for the main path. Optional follow-up: position validation on leader change like the Java client's `validatePositionsOnMetadataChange`.

thomhurst:
## Summary
- track current and last-fetched leader epochs on consumer fetch and ListOffsets requests
- surface fetch diverging epochs as consumer-visible truncation errors while repositioning to the broker-provided end offset
- carry leader epochs through consumed results, offset commits, offset fetches, and transactional offset commits

## Test plan
- [x] dotnet build --configuration Release
- [x] ./tests/Dekaf.Tests.Unit/bin/Release/net10.0/Dekaf.Tests.Unit (3845 passed)
- [x] ./tests/Dekaf.Tests.Integration/bin/Release/net10.0/Dekaf.Tests.Integration --treenode-filter /*/*/ConsumerTests/* (87 passed)
- [ ] Full integration suite attempted; timed out after 10 minutes before producing a final summary. Testcontainers from that run were cleaned up.

Closes #1150

### DIFF
--- a/src/Dekaf/Consumer/ConsumeBatch.cs
+++ b/src/Dekaf/Consumer/ConsumeBatch.cs
@@ -125,7 +125,7 @@ namespace Dekaf.Consumer
                     headerOwner: pending,
                     timestampMs: timestampMs,
                     timestampType: timestampType,
-                    leaderEpoch: null,
+                    leaderEpoch: pending.CurrentPartitionLeaderEpoch >= 0 ? pending.CurrentPartitionLeaderEpoch : null,
                     keyDeserializer: _batch._keyDeserializer,
                     valueDeserializer: _batch._valueDeserializer);
 
--- a/src/Dekaf/Consumer/ConsumerCoordinator.cs
+++ b/src/Dekaf/Consumer/ConsumerCoordinator.cs
@@ -328,7 +328,8 @@ public sealed partial class ConsumerCoordinator : IAsyncDisposable
                     partitions.Add(new OffsetCommitRequestPartition
                     {
                         PartitionIndex = offset.Partition,
-                        CommittedOffset = offset.Offset
+                        CommittedOffset = offset.Offset,
+                        CommittedLeaderEpoch = offset.LeaderEpoch
                     });
                 }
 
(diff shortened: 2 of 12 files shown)


thomhurst:
Pushed CI fix ea1fe67.

The failed job timed out in `MetadataManagerTests.DisposeAsync_CancelsBackgroundRefreshCtsCreatedWhileInitializationDrains` after 6.2s under full-suite Linux load. This branch does not touch metadata code; the test passed locally 20x before the change, so I hardened the test timeout while preserving the same disposal/cancellation assertions.

Validation:
- `dotnet build tests/Dekaf.Tests.Unit/Dekaf.Tests.Unit.csproj -c Release`
- focused test 10x
- `tests/Dekaf.Tests.Unit/bin/Release/net10.0/Dekaf.Tests.Unit.exe` (3845/3845 passed)

PR #1404, opened 2026-07-05
fix(protocol): validate lengths before allocation

### LINKED ISSUE #1342: fix(protocol): allocate-before-validate on length-prefixed reads - corrupt frame can force multi-GB allocation
## Problem

Length-prefixed protocol reads allocate before validating the claimed length against remaining bytes (`src/Dekaf/Protocol/KafkaProtocolReader.cs:692-708, 732-739, 793-842, 903-954`):

```csharp
private byte[] ReadBytesContent(int length)
{
    var result = new byte[length];          // allocated BEFORE bounds check
    if (_isContiguous)
    {
        if (length < 0 || _position > _span.Length - length)
            ThrowInsufficientData();
```

`ReadCompactArray<T>` similarly does `var length = ReadUnsignedVarInt() - 1; ... var result = new T[length];` with no check against `Remaining`.

The outer frame-size check caps the frame (16MB default), but an embedded length field can claim up to `int.MaxValue` elements while only a few bytes remain — `new T[2^31]` for a reference-type element attempts a ~16GB allocation (OOM / LOH pressure) before a single element read fails.

`Record.ValidateHeaderCount` (`Record.cs:169-181`) and `LazyRecordList.MaxReasonableRecordCount` already implement the correct defense; the generic array/bytes readers do not.

## Proposed fix

Before allocating, validate `length <= Remaining` (each element/byte requires at least 1 wire byte) and throw `MalformedProtocolDataException`; move `new byte[length]` in `ReadBytesContent`/`ReadRawBytes(int)` after the bounds check.

*Found by protocol audit, 2026-07-05.*

thomhurst:
Summary:
- validate claimed byte/string/array lengths before allocating buffers/arrays
- throw `MalformedProtocolDataException` for claimed lengths beyond remaining frame data
- add regression tests for bytes, raw bytes, and compact arrays

Test plan:
- [x] `dotnet build tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release --no-incremental -v:minimal`
- [x] `tests\Dekaf.Tests.Unit\bin\Release\net10.0\Dekaf.Tests.Unit.exe --treenode-filter "/*/*/KafkaProtocolReaderTests/*"`
- [x] `tests\Dekaf.Tests.Unit\bin\Release\net10.0\Dekaf.Tests.Unit.exe --treenode-filter "/*/*/ArrayEncodingTests/*"`
- [x] `dotnet build src\Dekaf\Dekaf.csproj --configuration Release --no-incremental -v:minimal`
- [x] `git diff --check`

Closes #1342

### DIFF
--- a/src/Dekaf/Protocol/KafkaProtocolReader.cs
+++ b/src/Dekaf/Protocol/KafkaProtocolReader.cs
@@ -624,10 +624,10 @@ public ref struct KafkaProtocolReader
         if (length == 0)
             return string.Empty;
 
+        ValidateReadableLength(length);
+
         if (_isContiguous)
         {
-            if (length < 0 || _position > _span.Length - length)
-                ThrowInsufficientData();
             var result = Encoding.UTF8.GetString(_span.Slice(_position, length));
             _position += length;
             return result;
@@ -691,11 +691,11 @@ public ref struct KafkaProtocolReader
 
     private byte[] ReadBytesContent(int length)
     {
+        ValidateReadableLength(length);
+
         var result = new byte[length];
         if (_isContiguous)
         {
-            if (length < 0 || _position > _span.Length - length)
-                ThrowInsufficientData();
             _span.Slice(_position, length).CopyTo(result);
             _position += length;
             return result;
@@ -733,6 +733,9 @@ public ref struct KafkaProtocolReader
     {
         if (count == 0)
             return [];
+
+        ValidateReadableLength(count);
+
         var result = new byte[count];
         ReadRawBytes(result);
         return result;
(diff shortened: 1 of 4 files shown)


thomhurst:
Addressed the review coverage gap in `b304c8f`:

- added multi-segment malformed-length coverage for `ReadBytes`
- added multi-segment malformed-length coverage for `ReadMemorySlice`
- added string claimed-length coverage for `ReadString`

Validation:
- `dotnet run --project tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release --no-restore -- --disable-logo --no-ansi --treenode-filter "/*/*/KafkaProtocolReaderTests/*" --minimum-expected-tests 21`
- `git diff --check`

thomhurst:
Addressed the review nit: extracted the duplicate multi-segment sequence setup into shared SequenceTestHelpers and updated both protocol test classes to use it. Validation: unit test project build, KafkaProtocolReaderTests, PrimitiveEncodingTests, format verify.

PR #1410, opened 2026-07-05
fix(networking): keep idle connections alive

### LINKED ISSUE #1339: fix(networking): receive loop tears down healthy idle connections after RequestTimeout
## Problem

The receive loop arms a 30s `RequestTimeout` on *every* pipe read, with no check that any request is actually awaiting a response (`src/Dekaf/Networking/KafkaConnection.cs:1069-1117`):

```csharp
if (timeoutCts.TryReset())
    timeoutCts.CancelAfter(_options.RequestTimeout);
result = await _reader!.ReadAsync(timeoutCts.Token).ConfigureAwait(false);
...
if (timeoutCts.IsCancellationRequested && !cancellationToken.IsCancellationRequested)
{
    LogReceiveTimeout(...); MarkDisposed();
    throw new KafkaException($"Receive timeout after ...ms - connection to broker {BrokerId} failed");
```

A connection with zero inbound traffic for 30s (sporadic producer, admin client, metadata connection between 5-minute refreshes) is marked disposed and killed, with an Error-level log.

## Impact

Defeats `ConnectionsMaxIdleMs` (default 540s — the idle reaper explicitly checks `PendingRequestCount != 0` before reaping, so keeping idle connections alive is clearly the intended design). Forces a full TCP+TLS+SASL reconnect on the next send — recurring tail-latency spikes and handshake churn under bursty workloads, plus spurious Error logs.

## Proposed fix

Only arm the read timeout when the pending-request count is non-zero (cheap atomic counter incremented in `AddPendingRequest` / decremented in `TryRemovePendingRequest`); read with the outer token when idle, re-arming when a request is registered (e.g. `CancelPendingRead()` + re-read, or when the timer fires check the count and re-arm instead of killing).

*Found by networking perf audit, 2026-07-05.*

thomhurst:
Summary:
- only arm the receive-loop `RequestTimeout` while responses are actually pending
- wake the receive loop when pending requests transition between idle and active so timeout state is refreshed
- add regressions for idle connections and caller-canceled requests staying connected past `RequestTimeout`

Test plan:
- [x] `dotnet build tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release --no-incremental -v:minimal`
- [x] `dotnet build src\Dekaf\Dekaf.csproj --configuration Release --no-incremental -v:minimal`
- [x] `tests\Dekaf.Tests.Unit\bin\Release\net10.0\Dekaf.Tests.Unit.exe --treenode-filter "/*/*/KafkaConnectionTests/*"`
- [x] `tests\Dekaf.Tests.Unit\bin\Release\net10.0\Dekaf.Tests.Unit.exe --treenode-filter "/*/*/ConnectionPoolConcurrencyTests/*"`
- [x] `tests\Dekaf.Tests.Unit\bin\Release\net10.0\Dekaf.Tests.Unit.exe --treenode-filter "/*/*/PooledPendingRequestTests/*"`
- [x] `git diff --check`

Closes #1339

### DIFF
--- a/src/Dekaf/Networking/KafkaConnection.cs
+++ b/src/Dekaf/Networking/KafkaConnection.cs
@@ -1061,15 +1061,7 @@ public sealed partial class KafkaConnection : IKafkaConnection, IIdleTrackedKafk
                     var readHasRequestTimeout = HasPendingRequests();
                     try
                     {
-                        if (!timeoutCts.TryReset())
-                        {
-                            reg.Dispose();
-                            timeoutCts.Dispose();
-                            timeoutCts = _timeoutCtsPool.Rent();
-                            reg = cancellationToken.CanBeCanceled
-                                ? cancellationToken.Register(static s => ((CancellationTokenSource)s!).Cancel(), timeoutCts)
-                                : default;
-                        }
+                        timeoutCts.TryReset();
 
                         // Only arm RequestTimeout while a response is outstanding. Idle
                         // connections are governed by ConnectionsMaxIdleMs instead.
--- a/src/Dekaf/Networking/KafkaConnection.cs
+++ b/src/Dekaf/Networking/KafkaConnection.cs
@@ -176,6 +176,7 @@ public sealed partial class KafkaConnection : IKafkaConnection, IIdleTrackedKafk
         new(TaskCreationOptions.RunContinuationsAsynchronously);
     private int _pendingRequestSlotOperationCount;
     private int _pendingRequestSlotsClosed;
+    private int _pendingRequestCount;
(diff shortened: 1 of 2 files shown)


thomhurst:
Addressed the concrete review findings in `db6eb2b`:

- removed the `TryReset` failure re-rent/register branch so shutdown falls through without extra CTS/registration churn
- widened the receive-loop tests from 100ms to 1s request timeout and still wait past the timeout, avoiding the caller-cancel vs request-timeout race on slow CI

Validation:
- `dotnet run --project tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release --no-restore -- --disable-logo --no-ansi --treenode-filter "/*/*/KafkaConnectionTests/*" --minimum-expected-tests 19`
- `git diff --check`

PR #1483, opened 2026-07-06
perf(producer): pre-serialize sealed batches

### LINKED ISSUE #1365: perf(producer): batch compression runs single-threaded on the sender coordinator - head-of-line blocking for all partitions
## Problem

With compression enabled, `Ready()` runs the full `codec.Compress` of up to a 1 MB batch inline on the single sender-coordinator thread that services ALL partitions of ALL brokers (`src/Dekaf/Producer/RecordAccumulator.cs:1279-1283, 2891-2907`; `src/Dekaf/Protocol/Records/RecordBatch.cs:471-523`):

```csharp
// Ready(), on the single sender-coordinator thread, per head batch:
if (!head.IsPreSerialized)
    PreSerializeBatch(head);           // -> RecordBatch.PreCompress -> codec.Compress inline
```

Append-side workers are parallel, but compression serializes to ~1 core (zstd at ~400-600 MB/s caps compressed throughput there), and each compression delays readiness checks for every other partition — head-of-line blocking across brokers. librdkafka compresses in per-toppar broker threads; Java compresses at append time on caller threads.

Note: #918 moved compression off the producing caller thread to fix tail-latency spikes; this issue is about the other end of that trade — the coordinator is now the bottleneck. The fix should avoid regressing #918.

## Proposed fix

Compress at seal time on the sealing thread (append worker or linger thread) inside `CompleteDetachedBatch`, or dispatch `PreSerializeBatch` to ThreadPool workers — the `IsPreSerialized` gate in `DrainBatchesForOneNode` (line 1420-1421) already tolerates asynchronous completion; only the `readyNodes.Add` would need to follow the flag.

*Found by producer perf audit, 2026-07-05.*

thomhurst:
## Summary
- Pre-serialize/compress sealed producer batches during detached completion instead of on the sender `Ready()` coordinator path.
- Keep a defensive `Ready()` fallback for any un-pre-serialized batch so ready partitions cannot be stranded.
- Skip pre-serialization for batches being purged and update ready-path coverage to assert no compression happens during `Ready()`.

## Test plan
- [x] `dotnet build tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release -p:TreatWarningsAsErrors=true -v:minimal /p:UseAppHost=true`
- [x] `tests\Dekaf.Tests.Unit\bin\Release\net10.0\Dekaf.Tests.Unit.exe --no-ansi --timeout 90s --treenode-filter "/*/*/RecordAccumulatorReadyTests/*"`

Closes #1365

### DIFF
--- a/src/Dekaf/Producer/RecordAccumulator.cs
+++ b/src/Dekaf/Producer/RecordAccumulator.cs
@@ -1275,10 +1275,10 @@ public sealed partial class RecordAccumulator : IAsyncDisposable
                 continue;
             }
 
-            // Batches are pre-serialized when sealed. Keep a defensive fallback so a
-            // manually-created or legacy batch cannot be orphaned in the ready queue.
+            // Pre-serialization runs on a worker after seal. If the notification arrives
+            // before that worker finishes, leave the batch queued; completion re-notifies.
             if (!head.IsPreSerialized)
-                PreSerializeBatch(head);
+                continue;
 
             // Sealed batches in the deque are always sendable. The linger/micro-linger
             // timer already determined readiness when it sealed the batch, so re-checking
@@ -1785,6 +1785,7 @@ public sealed partial class RecordAccumulator : IAsyncDisposable
         if (Interlocked.Exchange(ref batch._returnedToPool, 1) != 0)
             return;
 
+        batch.WaitForPreSerializationIfStarted();
         batch.WaitForCleanupIfStarted();
         _readyBatchPool.Return(batch);
     }
@@ -2248,7 +2249,7 @@ public sealed partial class RecordAccumulator : IAsyncDisposable
                 }
 
                 if (batchToPublish is not null)
-                    PublishSealedBatch(batchToPublish);
+                    StartPreSerialization(batchToPublish);
 
                 if (batchToReturn is not null)
                     _batchPool.Return(batchToReturn);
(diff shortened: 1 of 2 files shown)


thomhurst:
Addressed the blocking review in 994a3b18cf9163f08d1b032c1c3e202d845cdad4.

What changed:
- Compressed/pre-serialized sealed batches now run on a worker after the batch is enqueued, so the producing caller clears `RotationInProgress` without waiting on compression.
- `Ready()`/`Drain()` keep skipping batches whose `IsPreSerialized` gate is false; the worker re-notifies the partition after compression completes.
- `ReadyBatch` now tracks the pre-serialization task so purge/dispose cannot return a batch to the pool while compression still owns its `RecordBatch`.
- Added a blocking compression codec regression test proving both the sealing append and a same-partition append complete while compression is still blocked.

Validation:
- `dotnet build tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release -p:TreatWarningsAsErrors=true -v:minimal`
- `dotnet run --project tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release --framework net10.0 --treenode-filter "/Dekaf.Tests.Unit/Dekaf.Tests.Unit.Producer/RecordAccumulatorReadyTests/**"` -> 21/21 passed
- `dotnet run --project tests\Dekaf.Tests.Unit\Dekaf.Tests.Unit.csproj --configuration Release --framework net10.0 --treenode-filter "/Dekaf.Tests.Unit/Dekaf.Tests.Unit.Producer/**"` -> 640/640 passed
- `dotnet build src\Dekaf\Dekaf.csproj --configuration Release --framework netstandard2.0 -p:TreatWarningsAsErrors=true -v:minimal`
- `dotnet format Dekaf.sln --verify-no-changes --verbosity minimal`

PR #1532, opened 2026-07-06
feat(serialization): async serializer preparation to remove sync-over-async schema fetch

thomhurst:
## Problem

`ISerializer<T>.Serialize` is **synchronous and zero-allocation by design** — it writes into a `ref struct` `IBufferWriter<byte>`, which cannot cross an `await`, so it structurally cannot be async. The only async dependency in the Schema Registry serializers is a **one-time schema-ID fetch** per subject. Today that's resolved with a **sync-over-async block** on the first message for each subject (`GetSchemaIdCached` → `task.WaitAsync(...).GetAwaiter().GetResult()`), unless the caller remembers to `WarmupAsync` first. Under load that blocks a pool thread on a network round-trip.

Follow-up to a design discussion: *is this bad design, and can it be truly async?* Making `Serialize` async is out (kills zero-alloc). The clean fix is to **resolve the schema asynchronously on the async produce path, before the sync encode**.

## Approach

New opt-in companion interface:

```csharp
public interface IAsyncSerializerPreparer<in T>
{
    ValueTask PrepareAsync(T value, SerializationContext context, CancellationToken ct = default);
}
```

- The producer awaits `PrepareAsync` on its async `ProduceAsync` path **before** the synchronous `Serialize`. The one-time fetch runs where async belongs; the encode never blocks.
- **Zero added hot-path cost:** preparers are captured once in the ctor via `as`-cast. When no serializer needs preparation (built-in serializers), the gate is two null checks. After the first message per subject, `PrepareAsync` returns a synchronously-completed `ValueTask` (via a cheap `SubjectSchemaIdCache.Contains` check → `ValueTask.CompletedTask`, no `Task` allocation), so the produce path stays fully synchronous on the fast path.
- **Opt-in and non-breaking:** `Serialize` still works if reached without preparation (falls back to the existing behavior).

## Changes

- `IAsyncSerializerPreparer<T>` — new interface.
- `AvroSchemaRegistrySerializer<T>` — implements `PrepareAsync` (delegates to existing `WarmupAsync`, short-circuits when warm).
- `SubjectSchemaIdCache` — `Contains` warm-check (via a shared `TryGetCached` lookup primitive).
- `KafkaProducer` — captures `_keyPreparer`/`_valuePreparer`; gates both async `ProduceAsync` cores (message and topic/key/value). Fast path extracted, unchanged.

## Testing

- New unit tests prove the contract: cold `PrepareAsync` is **genuinely async** (doesn't block while the fetch is in flight); warm `PrepareAsync` completes **synchronously** with no extra registry call, and the subsequent `Serialize` doesn't re-fetch.
- Full `AvroSerializerTests` class passes; core + Avro + unit build clean.

## Prototype scope / follow-ups

- Covers **Avro** + the **two async `ProduceAsync` overloads**.
- **Fire-and-forget** produce is synchronous and cannot await — still needs a manual `WarmupAsync` (documented on the interface).
- **Protobuf** SR serializer has the same sync-over-async block; wiring is mechanical (mirror Avro) — follow-up.
- **Benchmark** validation of the produce hot path recommended before merge (design preserves zero-alloc by construction; a `[MemoryDiagnoser]` run should confirm).

> Note: this branch is off `main` and does not include the thread-pool pre-sizing fix from #1531; several unrelated timing tests (`SendLoop_*`, `ReceiveLoop_*`) flake under full-suite load until that lands. They pass in isolation. Rebase after #1531 merges for stable CI.

### DIFF
--- a/src/Dekaf.SchemaRegistry.Avro/AvroSchemaRegistrySerializer.cs
+++ b/src/Dekaf.SchemaRegistry.Avro/AvroSchemaRegistrySerializer.cs
@@ -35,7 +35,7 @@ namespace Dekaf.SchemaRegistry.Avro;
 /// <typeparam name="T">The type to serialize. Must be either an Avro ISpecificRecord or GenericRecord.</typeparam>
 public sealed class AvroSchemaRegistrySerializer<
     [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicFields)] T>
-    : ISerializer<T>, IAsyncDisposable
+    : ISerializer<T>, IAsyncSerializerPreparer<T>, IAsyncDisposable
 {
     private const byte MagicByte = 0x00;
     private const int WireHeaderSize = 5;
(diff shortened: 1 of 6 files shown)


thomhurst:
Addressed both blocking findings in `1d96c8eb`:

1. **Activity leak / lost error status** — `AwaitPrepareThenProduce` now records the exception and disposes the span before rethrowing, and the synchronous `PrepareSerializersAsync` call in `ProduceAsyncCore` is guarded the same way (new `RecordActivityFault` helper), matching the `AwaitWithActivity` invariant.
2. **Unobserved task exception** — `AwaitBothAsync` now awaits both preparers via `Task.WhenAll(keyPrepare.AsTask(), valuePrepare.AsTask())`, so a value-side fault is observed even when the key preparer faults first.

Added producer-level regression tests (`ProducerAsyncPreparerTests`) that drive the real `KafkaProducer` with a fake `IAsyncSerializerPreparer<T>`. Verified each fails against the pre-fix code and passes after: async-fault and sync-throw preparers assert the span is stopped + error-tagged, key-faults-first asserts the value preparer is still awaited, and cancellation-during-prep asserts pre-append semantics (throws, nothing buffered).

Issue #1543, opened 2026-07-07
1.2.0 netstandard 2.0 does not receive as subscriber.


marcschier:
Critical finding by copilot debugging, confirmed by a standalone repro: Dekaf 1.2.0's netstandard2.0 client builds fine but its consumer does not receive messages at runtime on net8.0 (produce succeeds at offset 0, but consume times out after 20s — while the identical pattern works in <1s on net10 with Dekaf's net10 assets). So the ns2.0 "support" is compile-time only for the consumer path; expanding to all TFMs as-is would ship a non-functional subscriber on net8/net9/net48/netstandard.

It would be great to expand the integration tests to .net 8 consuming core package as standard 2.0.  Also of course to fix the issue.  Let me know, happy to contribute the fix.

PR #1546, opened 2026-07-07
fix: guard empty spans in netstandard2.0 UTF-8 encoding (ns2.0 consumer receives nothing)

marcschier:
Addresses #1543.

## Symptom

On the **netstandard2.0** asset consumed from net8, produce succeeds but the consumer never receives — consume times out (while the identical pattern works in <1s on net10). The ns2.0 consumer path is effectively non-functional.

## Root cause

On netstandard2.0 the span-based `Encoding.GetBytes(ReadOnlySpan<char>, Span<byte>)` / `GetString(ReadOnlySpan<byte>)` / `GetByteCount(ReadOnlySpan<char>)` overloads are supplied by the `Polyfill` package, whose implementation pins the source span with `fixed`:

```csharp
fixed (char* charsPtr = chars)        // empty span -> null pointer
    return target.GetBytes(charsPtr, chars.Length, bytesPtr, bytes.Length);
```

Pinning an **empty** span yields a null pointer, and `UTF8Encoding` rejects it:

```
System.ArgumentNullException: Array cannot be null. (Parameter 'chars')
  at System.Text.UTF8Encoding.GetBytes(Char* chars, ...)
  at Polyfills.Polyfill.GetBytes(Encoding, ReadOnlySpan<char>, Span<byte>)
  at Dekaf.Protocol.KafkaProtocolWriter.WriteCompactStringValue(...)
  at Dekaf.Protocol.Messages.FetchRequest.Write(...)
  at Dekaf.Consumer.KafkaConsumer.PrefetchFromBrokerAsync(...)
```

The consumer's `FetchRequest.Write` writes an **empty** compact string (e.g. the empty client rack id), so on ns2.0 **every prefetch throws** and the subscriber receives nothing. net8.0/net9.0/net10.0/netstandard2.1 use the native, empty-safe overloads, so this is netstandard2.0-only.

## Fix

Adds `src/Dekaf/Internal/EncodingCompat.cs` — empty-safe `GetBytes`/`GetByteCount`/`GetString`:

```csharp
public static int GetBytes(ReadOnlySpan<char> chars, Span<byte> bytes)
{
#if NETSTANDARD2_0
    if (chars.IsEmpty)
        return 0;
#endif
    return Encoding.UTF8.GetBytes(chars, bytes);
}
```

The empty guard is `#if NETSTANDARD2_0`-gated and the methods are `AggressiveInlining`, so **net10.0 IL is unchanged (no regression)** — only the ns2.0 build gains the guard. The vulnerable span call sites are routed through it: `KafkaProtocolWriter`, `Headers`, `BuiltInSerializers`, `PlainAuthenticator` (empty SASL PLAIN authzid), and `ClientTelemetryPayloadProvider`. `KafkaProtocolReader` already guards `length == 0`.

## Validation

- Core builds 0 warnings / 0 errors on `net10.0` + `netstandard2.0` (Release).
- Verified against real Kafka brokers by running the end-to-end integration suite on a **net8.0 host consuming the netstandard2.0 asset**: 357/357 pass with the fix (all 33 failed before it). net10.0 RoundTrip 33/33 and net10.0 Protocol/Serialization unit tests remain green.

The net8→ns2.0 integration test harness that exercises this at runtime is kept separate from this change (per #1543, the test expansion can be contributed on its own); happy to follow up with it if you'd like it upstream.

### DIFF
--- a/src/Dekaf/Internal/EncodingCompat.cs
+++ b/src/Dekaf/Internal/EncodingCompat.cs
@@ -1,0 +1,53 @@
+using System;
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace Dekaf.Internal;
+
+/// <summary>
+/// UTF-8 span helpers that behave identically across all target frameworks.
+/// </summary>
+/// <remarks>
+/// On <c>netstandard2.0</c> the span-based <see cref="Encoding.GetBytes(ReadOnlySpan{char}, Span{byte})"/>
+/// and <c>Encoding.GetByteCount(ReadOnlySpan&lt;char&gt;)</c> overloads are supplied by the Polyfill package,
+/// whose implementation pins the source span with <c>fixed</c>. Pinning an <b>empty</b> span yields a null
+/// pointer, which <see cref="UTF8Encoding"/> rejects with an <see cref="ArgumentNullException"/>
+/// ("Array cannot be null. (Parameter 'chars')"). These helpers short-circuit empty input so the
+/// netstandard2.0 build matches the net10.0 behaviour (zero bytes). On net8.0+/netstandard2.1 the native,
+/// empty-safe overloads are used and the guard is compiled out, so those builds gain no extra branch.
+/// </remarks>
+internal static class EncodingCompat
+{
+    /// <summary>UTF-8 encodes <paramref name="chars"/> into <paramref name="bytes"/>; empty-safe on all TFMs.</summary>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static int GetBytes(ReadOnlySpan<char> chars, Span<byte> bytes)
+    {
+#if NETSTANDARD2_0
+        if (chars.IsEmpty)
+            return 0;
+#endif
+        return Encoding.UTF8.GetBytes(chars, bytes);
+    }
+
+    /// <summary>Counts the UTF-8 bytes for <paramref name="chars"/>; empty-safe on all TFMs.</summary>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static int GetByteCount(ReadOnlySpan<char> chars)
+    {
+#if NETSTANDARD2_0
+        if (chars.IsEmpty)
+            return 0;
+#endif
+        return Encoding.UTF8.GetByteCount(chars);
+    }
+
+    /// <summary>UTF-8 decodes <paramref name="bytes"/> into a string; empty-safe on all TFMs.</summary>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static string GetString(ReadOnlySpan<byte> bytes)
+    {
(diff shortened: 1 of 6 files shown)


thomhurst (commented):
## Review — LGTM, correct and well-scoped

I traced the fix end-to-end against the codebase. This is a solid, well-diagnosed fix and the root-cause writeup is excellent.

### Correctness ✅
- **Scope is confined correctly.** Only `src/Dekaf` targets `netstandard2.0` and references `Polyfill` (10.11.2) — no other project does. So confining `EncodingCompat` + the `#if NETSTANDARD2_0` guard to that assembly covers the entire blast radius. Nothing in `Dekaf.SchemaRegistry` (net10-only) needs touching.
- **No-regression claim holds.** On net10.0 the guard compiles out and `EncodingCompat.*` is a thin `AggressiveInlining` passthrough to `Encoding.UTF8.*` — identical semantics, JIT-inlined. ✔
- **The compact-string path is the right fix.** `WriteCompactStringValue` on empty input now yields `byteCount = 0 → WriteUnsignedVarInt(1)`, the correct compact-encoding of an empty string — exactly the empty rack-id in `FetchRequest.Write` from the stack trace.

### Completeness — I audited every remaining span-based encoding call in `Dekaf`
All un-migrated span overloads are provably unreachable with empty input, so the migration is complete:
- `KafkaProtocolReader` (`GetString` x3) — all funnel through `ReadStringContent`, which returns early on `length == 0`.
- `Utf8StringInternCache.Decode` — guarded by `Intern`'s `utf8Bytes.Length == 0` early return.
- `DeadLetterHeaderFormatting` — `Utf8Formatter.TryFormat` of an int/long always writes ≥1 byte, never an empty span.
- `BrokerSender.CompactStringSize` — binds to the native `GetByteCount(string)` overload, not the Polyfill span one.

### Nice bonus
Routing `BuiltInSerializers` string deserialization and `Headers.GetValueAsString` through `EncodingCompat` also hardens two previously-unguarded `GetString(ReadOnlySpan<byte>)` sites — an empty (non-null) header value or an empty string payload would have thrown the same `ArgumentNullException` on ns2.0. Good to fix these while here.

### One suggestion (non-blocking)
The fix ships with **no in-repo regression test**. The PR notes the net8→ns2.0 integration harness is being contributed separately — understood, and a full harness is heavyweight. But given this class of bug is invisible on the net10 test host (the guard is compiled out) and easy to silently reintroduce, it would be worth landing *something* multi-targeted to ns2.0 that asserts `EncodingCompat.GetBytes(default, default) == 0` / `GetString(default) == ""`, even a tiny unit test. Otherwise CI can never catch a future `Encoding.UTF8.GetBytes(span, …)` creeping back into a hot path. Happy to see it as a follow-up rather than block on it.

Overall: correct fix, minimal surface, no net10 regression, and more thorough than the minimal patch. 👍

thomhurst:
Thanks!

PR #1549, opened 2026-07-07
perf(producer): raise non-idempotent in-flight

### LINKED ISSUE #1548: perf(producer): raise default MaxInFlightRequestsPerConnection for non-idempotent producers to match librdkafka
## Problem

The single-broker fire-and-forget stress test (acks=leader, non-idempotent) shows default-config Dekaf losing to default-config Confluent (1.29M vs 1.41M msg/s, 2026-07-07), but the defaults are not equivalent:

- Dekaf: `MaxInFlightRequestsPerConnection = 5` (`ProducerOptions.cs:159`) — Java client's default
- librdkafka: `max.in.flight.requests.per.connection = 1,000,000` — effectively unlimited

So at "same config" (both defaults), Confluent pipelines its single connection far deeper than Dekaf, which stalls at 5 outstanding requests waiting for broker responses.

## Proposal

Adopt librdkafka's stance for non-idempotent producers: a deep default in-flight limit (either literally 1,000,000 or a generous bound like 100 — measure both). Idempotent/transactional producers stay capped at 5 (broker protocol limit for sequence tracking).

## Trade-off (semantics-visible)

With retries enabled and >1 request in flight, a non-idempotent producer can reorder messages when an earlier request fails and is retried behind a later successful one. librdkafka ships this behavior by default; the Java client does not. Whoever picks the value should:

- document the reordering caveat on `MaxInFlightRequestsPerConnection` and in the producer docs, with "set to 1 (or enable idempotence) if strict ordering matters"
- verify the send-loop in-flight wait (`BrokerSender.cs:858`) and pending-response structures behave sensibly at large limits (e.g. `maxCoalesce = MaxInFlight * 4` at `BrokerSender.cs:617` — a 1M default would make that allocation absurd, so the coalesce sizing needs decoupling from the limit or the default needs to be a sane finite number)

Note `PoolSizing.InflightEntriesPerBatch` pre-warm at `KafkaProducer.cs:404` also scales with this value — check for pathological pre-allocation at large defaults.

## Validation

CI stress workflow (workflow_dispatch); local Windows repro is NAT-capped and useless for this. Expect this to close the fire-and-forget default-config gap; the remaining 1-conn gap (visible in the idempotent table where both clients are capped at 5) is tracked separately in #1547.

thomhurst:
## Summary
- default non-idempotent producer max in-flight requests to 100 while keeping idempotent/transactional producers capped at 5
- cap coalescing and pool sizing math for very high explicit in-flight limits
- document non-idempotent retry reordering caveat

## Test plan
- [x] dotnet build tests\\Dekaf.Tests.Unit\\Dekaf.Tests.Unit.csproj --configuration Release
- [x] tests\\Dekaf.Tests.Unit\\bin\\Release\\net10.0\\Dekaf.Tests.Unit.exe --treenode-filter "/*/*/ProducerBuilderValidationTests/*"
- [x] tests\\Dekaf.Tests.Unit\\bin\\Release\\net10.0\\Dekaf.Tests.Unit.exe --treenode-filter "/*/*/PoolSizingTests/*"
- [x] tests\\Dekaf.Tests.Unit\\bin\\Release\\net10.0\\Dekaf.Tests.Unit.exe
- [x] npm run build --prefix docs
- [ ] stress-tests.yml workflow_dispatch after PR CI is available

Closes #1548

### DIFF
--- a/docs/docs/producer/basics.md
+++ b/docs/docs/producer/basics.md
@@ -155,6 +155,12 @@ With idempotence enabled:
 - Messages are delivered in order
 - Requires `Acks.All` (set automatically)
 
+## In-Flight Requests and Ordering
+
+`MaxInFlightRequestsPerConnection` controls how many produce requests can be outstanding on one broker connection. Idempotent and transactional producers are capped at 5, which is the Kafka protocol limit for sequence tracking.
+
+When idempotence is disabled, Dekaf defaults to 100 in-flight requests per connection so fire-and-forget and `Acks.Leader` workloads can keep the broker pipeline full. With retries enabled, non-idempotent producers using more than 1 in-flight request can reorder messages if an earlier request is retried after a later request succeeds. Set `MaxInFlightRequestsPerConnection` to 1, or keep idempotence enabled, when strict per-partition ordering matters.
+
 ## Error Handling
 
 `ProduceAsync` throws exceptions for delivery failures:
--- a/src/Dekaf.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/Dekaf.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
@@ -775,7 +775,8 @@ internal static class DekafOptionsBinding
         builder.WithLinger(TimeSpan.FromMilliseconds(options.LingerMs));
         builder.WithBatchSize(options.BatchSize);
         builder.WithBufferMemory(options.BufferMemory);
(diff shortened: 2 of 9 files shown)


thomhurst:
Addressed the blocking typed-options DI review: explicit MaxInFlightRequestsPerConnection=5 is now tracked separately from unset values, so non-idempotent typed options still default to 100 only when the property was not initialized. Validation: unit build passed, focused MaxInFlight DI tests passed, full unit suite passed (4464/4464).

PR #1551, opened 2026-07-07
perf(producer): overlap single-conn writes

### LINKED ISSUE #1547: perf(producer): overlap request build with socket flush in BrokerSender send loop
## Problem

Single-connection Dekaf loses the single-broker fire-and-forget stress test to Confluent (1.29M vs 1.41M msg/s, 2026-07-07 results), and loses the idempotent 1-conn matchup (1.31M vs 1.51M) even though both clients are protocol-capped at 5 in-flight requests there. Dekaf uses only 1.10 of 2 cores vs Confluent's 1.76 — the send loop is pipeline-bound, not CPU-bound.

## Root cause

The send loop (`BrokerSender.SendLoopAsync`, single-connection fast path at `BrokerSender.cs:964`) is fully serial per iteration:

1. coalesce ready batches
2. copy + CRC32C into the request buffer
3. `await SendCoalescedAsync(...)` — includes the socket write/flush
4. only then loop back to build the next request

At ~1.3 GB/s the kernel socket buffer is permanently saturated, so the flush await routinely blocks on socket drain. During that time the send thread does nothing — no coalescing, no copy, no CRC for the next request. librdkafka's broker thread keeps building message sets while the socket drains (non-blocking I/O event loop), which is why it sustains higher throughput at equal in-flight depth.

## Proposed fix

Double-buffer the single-connection path: fire the connection write without awaiting the flush, continue to the next coalesce/build iteration, and await the previous write only when the next request is ready to send (bounded 2-deep per connection). In-flight response accounting (`_totalPendingResponseCount`, `_pendingResponsesByConnection`) already exists and is unchanged — this only changes when the send loop thread blocks.

Care needed around:
- error/timeout paths (`sendTimeoutCts` reuse assumes one outstanding send; `ProduceRequestScratch` reuse assumes the previous send completed — needs a second scratch)
- ordering: writes to one connection must still be issued in order (they are — same thread issues them sequentially; only the flush-await moves)
- multi-connection path already overlaps flushes across connections (`AwaitParallelSendsAsync`); this issue is about overlap between build and flush *within* one connection

## Validation

Local Windows stress repro is useless (Docker NAT caps ~80-120k msg/s). Validate via CI stress workflow (`stress-tests.yml` workflow_dispatch, duration_minutes input). Optionally confirm the diagnosis first with a traced run: send-loop wall time should show large share in socket flush await vs build work.

Expected outcome: closes most of the 1-conn gap in both the fire-and-forget and idempotent single-broker tables; complements #1136/#1143 (append-time encoding already moved wire-encode off this thread).

thomhurst:
## Summary
- overlap single-connection `Acks.None` request build with the previous socket write by keeping one fire-and-forget send in flight and awaiting it when the next request is ready
- use a second `ProduceRequestScratch` and timeout CTS slot for the bounded 2-deep single-connection fire-and-forget path, so scratch/timeout reuse stays safe
- keep pipelined produce responses on the connection-owned timeout path so concurrent in-flight sends do not share the send-loop CTS
- strengthen send-loop tests for two concurrent pipelined sends and fire-and-forget write overlap

## Test plan
- [x] dotnet build tests\\Dekaf.Tests.Unit\\Dekaf.Tests.Unit.csproj --configuration Release
- [x] dotnet run --project tests\\Dekaf.Tests.Unit\\Dekaf.Tests.Unit.csproj --configuration Release --no-build -- --treenode-filter "/*/*/BrokerSenderSendLoopTests/SendLoop_FireAndForget_StartsSecondWriteBeforeFirstFlushCompletes|/*/*/BrokerSenderSendLoopTests/SendLoop_PipelinedProduce_UsesConnectionOwnedTimeoutsForConcurrentSends" --minimum-expected-tests 2 --output Detailed --progress off
- [x] dotnet run --project tests\\Dekaf.Tests.Unit\\Dekaf.Tests.Unit.csproj --configuration Release --no-build -- --output Normal --progress off
- [ ] stress-tests.yml workflow_dispatch after PR CI is available

Closes #1547

### DIFF
--- a/src/Dekaf/Producer/BrokerSender.cs
+++ b/src/Dekaf/Producer/BrokerSender.cs
@@ -628,9 +628,12 @@ internal sealed partial class BrokerSender : IAsyncDisposable
         var responseLookup = new Dictionary<(string Topic, int Partition), ProduceResponsePartitionData>();
 
         // Pre-allocate reusable ProduceRequest structures.
-        // The send loop is single-threaded and awaits each send before reusing,
-        // so these can be safely reused without synchronization.
+        // The send loop is single-threaded. Acks.None single-connection sends can be
+        // one write ahead, so that path gets a second scratch slot before reuse.
         var requestScratch = new ProduceRequestScratch(_options, _compressionCodecs, maxCoalesce);
+        var singleConnectionFireAndForgetScratches = _options.Acks == Acks.None
+            ? new[] { requestScratch, new ProduceRequestScratch(_options, _compressionCodecs, maxCoalesce) }
+            : Array.Empty<ProduceRequestScratch>();
 
         // Per-connection scratch and buckets for partition-affined multi-send.
         // When _connectionCount == 1, single scratch and bucket — no grouping overhead.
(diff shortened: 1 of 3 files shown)


thomhurst:
Addressed blocking review: PR now includes the actual bounded single-connection fire-and-forget write overlap for #1547, with a second scratch/timeout slot so reuse remains safe. Also strengthened the pipelined timeout test to cover two concurrent sends. Validation: unit build passed; focused BrokerSender send-loop tests passed; full unit suite passed (4458/4458).

PR #1555, opened 2026-07-08
fix(networking): abort connection when a frame write fails or is cancelled

thomhurst:
## Root cause (stress run 28907694063)

In the 1-broker producer stress tests, Dekaf sustained **1.52M msg/s (median)** vs Confluent's 1.43M — until ~10 minutes in, when throughput flatlined to ~0 for the rest of the run, dragging the reported average down to 956k. Both single-connection 1-broker producer jobs (fire-and-forget and acks-all) hit it at almost the same offset.

Broker log at the exact stall second:

```
InvalidRequestException: Error getting request for apiKey: PRODUCE, apiVersion: 11 ...
Caused by: java.lang.RuntimeException: Error reading byte array of 120 byte(s): only 0 byte(s) available
```

Sequence:
1. Broker stalls briefly (tmpfs retention burst) → TCP window fills → `_stream.WriteAsync` blocks.
2. The 30s request-timeout token cancels the write **mid-frame**. Part of the frame is already on the wire.
3. `WritePreSerializedToStreamAsync` threw `KafkaException("Flush timeout")` but left the connection in use. Every subsequent request was appended to a misaligned stream → broker misparses → closes socket.
4. Producer wedged in ~30s timeout/retry cycles (visible as 1-second throughput blips every ~30s), 2943 errors, 61s max latency, run overran 900s→953s. Broker disk dropping to ~0 was retention deleting segments because nothing was arriving — symptom, not cause.

## Fix

Any exception from the frame write (timeout, user cancellation, IO failure) now aborts the connection:

- `MarkDisposed()` — the `ConnectionPool` replaces it, and writers already queued on the write lock refuse to append another frame to the poisoned stream (new guard at the top of `WritePreSerializedToStreamAsync`).
- Dispose the socket — unblocks the read pump so the receive loop fails all in-flight requests promptly instead of each waiting out its own request timeout.

This matches the Java client's behavior of disconnecting on request timeout rather than reusing a stream whose alignment is unknown. Covers all request types (produce, fetch, metadata) since everything goes through this write path.

## Expected impact

With the healthy-period rate already above Confluent's, fixing the collapse should flip the 1-broker producer stress tables in Dekaf's favor.

Per instruction, no tests in this PR — the failure needs a >30s broker-side write stall to reproduce, which the existing 15-minute stress runs exercise naturally.

### DIFF
--- a/src/Dekaf/Networking/KafkaConnection.cs
+++ b/src/Dekaf/Networking/KafkaConnection.cs
@@ -199,6 +199,8 @@ public sealed partial class KafkaConnection : IKafkaConnection, IIdleTrackedKafk
     private long _receiveTimeoutDeadlineTimestamp;
     private int _receiveTimeoutExpired;
     private OAuthBearerTokenProvider? _ownedTokenProvider;
+    private readonly object _disposeGate = new();
+    private Task? _disposeTask;
     private int _disposed;
     private int _connected;
     private long _lastUsedTimestampMs = Dekaf.MonotonicClock.GetMilliseconds();
@@ -1014,7 +1016,7 @@ public sealed partial class KafkaConnection : IKafkaConnection, IIdleTrackedKafk
         // outgoing stream misaligned. Writers already queued on _writeLock must not append
         // another frame to a poisoned stream.
         if (Volatile.Read(ref _disposed) != 0)
-            throw new KafkaException($"Connection to broker {BrokerId} is closed");
+            throw new ObjectDisposedException(nameof(KafkaConnection));
 
 #if DEBUG
         System.Diagnostics.Debug.Assert(!callerOwnsTimeout || cancellationToken.CanBeCanceled,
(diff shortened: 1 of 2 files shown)


thomhurst:
Addressed the blocking review in ef61d8d2.

- `AbortAfterWriteFailure()` now starts the same full `DisposeAsync` teardown path instead of hand-disposing only the socket.
- Disposal is idempotent via a cached dispose task, so a prior `MarkDisposed()` no longer makes later cleanup a no-op.
- Added `WriteFailure_DisposeAsyncRunsFullStreamCleanup` to assert a write failure still allows full stream cleanup.
- Kept the conservative write-cancellation behavior intentionally: if cancellation happens during `WriteAsync`, the client cannot prove frame alignment is intact, so the broker connection is poisoned and replaced.

Local verification:
- `dotnet build tests/Dekaf.Tests.Unit/Dekaf.Tests.Unit.csproj --configuration Release`
- `Dekaf.Tests.Unit.exe --treenode-filter /*/*/KafkaConnectionTests/WriteFailure_DisposeAsyncRunsFullStreamCleanup`
- `Dekaf.Tests.Unit.exe --treenode-filter /*/*/KafkaConnectionTests/*`
- `dotnet format --verify-no-changes --verbosity minimal`
- full `Dekaf.Tests.Unit.exe` net10.0 suite passed on rerun; first run had one producer metrics test fail, then pass solo and pass in the full rerun.

PR #1562, opened 2026-07-08
fix(networking): never cancel a frame write mid-frame — abandon the wait instead

### LINKED ISSUE #1559: P0: Torn PRODUCE v11 frame still reproduces after #1555 — partial frame reaches wire
## Evidence

Stress run [28911368290](https://github.com/thomhurst/Dekaf/actions/runs/28911368290) (sha d042ac33, **includes #1555**), "Dekaf Producer" (FnF, 1 broker, single conn) job — broker log at 01:58:10, minute ~14 of 15:

```
org.apache.kafka.common.errors.InvalidRequestException: Error getting request for apiKey: PRODUCE, apiVersion: 11, connectionId: 172.17.0.2:9092-172.17.0.1:60080-1-1
Caused by: java.lang.RuntimeException: Error reading byte array of 120 byte(s): only 0 byte(s) available
ERROR Closing socket for 172.17.0.2:9092-172.17.0.1:60080-1-1 because of error (kafka.network.Processor)
```

Identical signature to the pre-#1555 corruption (run 28907694063).

## What #1555 did and didn't fix

- **Fixed:** reuse-after-torn-write. Client now aborts the connection and recovers — 1069 errors, 1.32M msg/s avg, no permanent wedge (previously producer wedged with throughput ~0).
- **Not fixed:** the torn frame itself. A partial PRODUCE frame still reaches the wire under sustained max-throughput load. Some write path still cancels or interleaves mid-frame before the abort lands.

## Impact

Single biggest producer throughput leak: pre-corruption sustained rate ~1.52M msg/s vs 1.32M avg in this run, plus 1069 failed deliveries (accepted-vs-delivered gap ~1071). Blocks Dekaf decisively beating Confluent in the 1-broker FnF table.

## Next steps

- Audit **all** socket write paths for cancellable mid-frame writes, not just `WritePreSerializedToStreamAsync` (the only path #1555 hardened).
- Make frame writes uncancellable except at frame boundaries — honor request timeout by abandoning the wait, not by cancelling `Stream.WriteAsync` mid-frame.
- Verify no second queued writer can flush bytes between a partial write and the abort completing (single-writer ordering during `AbortAfterWriteFailure`).

## Repro

Single-connection 1-broker producer stress ≥10 min at ~1.5M msg/s; grep broker log for `InvalidRequestException`.

thomhurst:
Fixes #1559

## Problem

#1555 stopped a torn frame from poisoning subsequent writes, but the torn frame itself still hit the wire: a request timeout cancelled `Stream.WriteAsync` mid-frame, so the broker always saw a partial PRODUCE frame (`InvalidRequestException: Error reading byte array of 120 byte(s): only 0 byte(s) available`, socket closed), and any transient broker stall cost the whole connection plus all in-flight requests. Reproduced in stress run 28911368290 with #1555 merged: 1069 errors at minute ~14, throughput dip on reconnect.

## Fix

Frame writes are no longer cancellable. `WriteFrameHoldingLockAsync` writes the complete frame with `CancellationToken.None` while holding the write lock, and owns lock release + pooled buffer return. Timeout/cancellation semantics move to `AwaitFrameWriteAsync`, which **abandons the wait instead of cancelling the write** (`Task.WaitAsync`):

- Write completes late → stream still frame-aligned, connection stays usable, zero errors. A transient stall (the stress-run case) no longer tears the connection down. Late responses are already dropped by `DispatchResponse` as unknown correlation ids.
- Write faults → `AbortAfterWriteFailure()` as before (partial frame may be on the wire).
- Write stuck past a further request-timeout grace period → background observer force-aborts the connection; disposing the socket unblocks the pending write on a dead broker.

Caller-visible exceptions are unchanged (`KafkaException` flush timeout / `OperationCanceledException`). The non-`callerOwnsTimeout` path drops the pooled CTS + registration in favour of `WaitAsync(timeout, token)`.

## Tests

- `WriteTimeout_AbandonedWriteCompletes_ConnectionStaysUsable` — deterministic (TCS-backed stream, no delays): caller times out, write must not be cancelled, connection must not be disposed once the write drains, write lock released exactly once.
- `WriteFailure_DisposeAsyncRunsFullStreamCleanup` — rewired to the new entry point; now also asserts the faulted write self-aborts the connection and releases the lock.

Behaviour change to note: a genuinely dead broker now costs one extra request timeout (grace period) before the connection is replaced; a transiently stalled broker costs nothing instead of a connection teardown.

### DIFF
--- a/src/Dekaf/Networking/KafkaConnection.cs
+++ b/src/Dekaf/Networking/KafkaConnection.cs
@@ -975,119 +975,165 @@ public sealed partial class KafkaConnection : IKafkaConnection, IIdleTrackedKafk
     {
         var (serializedArray, serializedLength) = PreSerializeRequest<TRequest, TResponse>(request, correlationId, apiVersion, headerVersion);
         var clearSerializedArray = KafkaMessageMetadata<TRequest, TResponse>.ApiKey == ApiKey.SaslAuthenticate;
-        byte[]? arrayToReturn = serializedArray;
         try
         {
             await SemaphoreHelper.AcquireOrThrowDisposedAsync(_writeLock, nameof(KafkaConnection), cancellationToken)
                 .ConfigureAwait(false);
-            try
-            {
-                await WritePreSerializedToStreamAsync(
-                        serializedArray,
-                        serializedLength,
-                        correlationId,
-                        cancellationToken,
-                        callerOwnsTimeout)
-                    .ConfigureAwait(false);
-            }
-            finally
-            {
-                SemaphoreHelper.ReleaseSafely(_writeLock);
-            }
+        }
+        catch
+        {
+            // Lock never acquired: no bytes written, so the buffer can be returned here.
+            DekafPools.SerializationBuffers.Return(serializedArray, clearArray: clearSerializedArray);
+            throw;
+        }
+
+        // From here the frame write owns both the lock and the buffer; WriteFrameHoldingLockAsync
+        // releases them itself so a timed-out caller can abandon the wait without cancelling the
+        // socket write mid-frame (see AwaitFrameWriteAsync).
+        var writeTask = WriteFrameHoldingLockAsync(serializedArray, serializedLength, clearSerializedArray);
+        await AwaitFrameWriteAsync(writeTask, correlationId, callerOwnsTimeout, cancellationToken).ConfigureAwait(false);
+    }
+
+    /// <summary>
+    /// Writes one complete frame while holding the write lock, then releases the lock and
+    /// returns the serialization buffer. The socket write is deliberately not cancellable:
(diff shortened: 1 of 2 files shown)


thomhurst:
Addressed the blocking review feedback: added coverage for the abandoned frame write stuck past its grace period, and kept the frame write semaphore available until abandoned writes unwind so the write lock is released after stream disposal unblocks the pending write.

Local validation:
- dotnet build tests/Dekaf.Tests.Unit --configuration Release
- tests/Dekaf.Tests.Unit/bin/Release/net8.0/Dekaf.Tests.Unit.exe --hangdump --hangdump-timeout 15m --log-level Debug
- tests/Dekaf.Tests.Unit/bin/Release/net10.0/Dekaf.Tests.Unit.exe --hangdump --hangdump-timeout 15m --log-level Debug

PR #1564, opened 2026-07-08
feat(stress): report median interval throughput

### LINKED ISSUE #1561: Stress test reporting: 15-min mean is runner-noise sensitive — publish median-of-intervals and lead with CPU μs/msg
## Problem

Run-over-run stress numbers swing far more than the code changes between them. Between the 2026-07-07 and 2026-07-08 runs, **Confluent** (unchanged binary) dropped 13-24% in every producer scenario (FnF 1.41M→1.07M msg/s), and Dekaf (3conn) FnF showed an apparent 1.73M→970k crater with 0 errors. These are shared-CI runner variance, not regressions — but the published tables can't distinguish the two, so every run triggers a false regression hunt.

CPU μs/msg is much stabler across runs and already tells the true story (Dekaf better in every producer scenario in both runs: 0.76-1.40 vs 1.33-1.94).

## Proposal

- Record per-interval throughput during the run (likely already sampled) and publish **median of intervals** alongside/instead of the 15-min mean — a late-run stall (e.g. #1559) then shows as a visible tail event instead of silently dragging the average.
- Lead the headline comparison with CPU μs/msg (or show both), since it is far less sensitive to runner contention.
- Optionally: flag runs where Confluent's own numbers deviate >10% from its trailing median as "noisy environment" in the generated doc.

## Why it matters

Weekly numbers are the project's public benchmark claim. Noise-robust reporting stops false alarms and makes real regressions (like the #1559 stall) stand out instead of blending into variance.

thomhurst:
## Summary
- serialize median sampled interval throughput in stress results
- show CPU microseconds per message first while keeping mean, median, accepted rate, errors, and throughput ratio visible
- update generated docs callouts to prefer CPU efficiency when CPU data exists

## Test plan
- [x] dotnet build tools/Dekaf.StressTests --configuration Release
- [x] python -m py_compile .github/scripts/stress_report.py
- [x] synthetic C# report generation from sample stress results
- [x] synthetic Python report generation from sample stress results

Closes #1561

### DIFF
--- a/.github/scripts/stress_report.py
+++ b/.github/scripts/stress_report.py
@@ -1,6 +1,8 @@
 """Shared utilities for stress test result reporting."""
 
 from collections import defaultdict
+from math import isfinite
+from statistics import median
 
 SCENARIO_TITLES = {
     'producer': 'Producer (Fire-and-Forget)',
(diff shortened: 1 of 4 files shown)


thomhurst:
Addressed the review nit in 60af9242 by removing the unused MedianIntervalMegabytesPerSecond property. Verified with dotnet build tools/Dekaf.StressTests/Dekaf.StressTests.csproj -c Release and git diff --check.

PR #1567, opened 2026-07-08
fix(consumer): cache repeated string values

### LINKED ISSUE #1560: Consumer efficiency gap vs Confluent: 2.31 KB/msg allocated, heavy Gen1/Gen2, 20% worse CPU per message
## Evidence

Stress run [28911368290](https://github.com/thomhurst/Dekaf/actions/runs/28911368290) (2026-07-08), 15-min consumer scenario, 1000B messages:

| Metric | Dekaf | Confluent |
|---|---|---|
| Messages/sec | 1,262,271 | 1,095,235 |
| CPU μs/msg | **1.22** | **1.02** |
| Cores used | **1.54** | **1.11** |
| Alloc/msg | 2.31 KB | 2.38 KB (managed side only) |
| Gen0 / Gen1 / Gen2 | 56,245 / **32,897** / **1,556** | 474,970 / 1 / 1 |

Dekaf wins raw throughput (1.15x) but loses efficiency: ~20% more CPU per message and 40% more cores. The Gen1/Gen2 profile is pathological — 1,556 Gen2 collections in 15 minutes vs Confluent's 1, meaning allocations are surviving Gen0 (long-lived buffers or batch objects held across collections), not just churning.

Contrast with Dekaf's own raw-bytes path, which proves the transport is fine:

- Consumer (Raw Bytes): 76 B/msg, 182 Gen2, 0.58 CPU μs/msg
- Consumer (Raw Batch): 76 B/msg, 140 Gen2, 0.51 CPU μs/msg

So the 2.31 KB/msg and the Gen1/Gen2 pressure come from the deserialization/materialization layer between raw fetch and typed `ConsumeResult`.

## Why it matters

This is the only scenario where Confluent is genuinely competitive with Dekaf on efficiency. Closing it makes Dekaf the clear winner across every table. Known weak spot from the 2026-07 audit (consumer GC anomaly).

## Next steps

- Profile the typed-consumer path (perf-trace stress run) to attribute the 2.25 KB/msg delta between raw and typed paths.
- Identify why allocations survive to Gen1/Gen2 — suspect pooled/held buffers or per-message result objects rooted by batch lifetime.
- Target: typed consumer CPU ≤ 1.0 μs/msg and Gen2 count in the same order as the raw path.

thomhurst:
## Summary
- Generalize the built-in string key cache into a bounded string deserializer.
- Keep string value caching opt-in through `WithCachedStringValues()` so default string consumers do not pay hash/lookup cost.
- Opt the typed consumer stress scenarios into cached string values because those scenarios replay a fixed pre-seeded payload.
- Add regression coverage plus repeated-payload and saturated-cache consumer hot-path benchmarks.

## Benchmark
`dotnet run --project tools/Dekaf.Benchmarks/Dekaf.Benchmarks.csproj -c Release -- --filter "*ConsumerHotPathBenchmarks*" --artifacts "artifacts/benchmarks/issue-1560-typed-opt-in"`

| Case | Mean | Allocated |
| --- | ---: | ---: |
| Typed batch string deserialize | 45.033 ns | 80 B |
| Typed batch string cached deserialize (saturated) | 51.270 ns | 80 B |
| Typed batch repeated 1KB string deserialize | 161.634 ns | 2064 B |
| Typed batch repeated 1KB string cached deserialize | 126.766 ns | 40 B |

## Tests
- `dotnet build src/Dekaf/Dekaf.csproj -c Release`
- `dotnet build tools/Dekaf.Benchmarks/Dekaf.Benchmarks.csproj -c Release`
- `dotnet build tools/Dekaf.StressTests/Dekaf.StressTests.csproj -c Release`
- `dotnet build tests/Dekaf.Tests.Unit/Dekaf.Tests.Unit.csproj -c Release -f net10.0`
- `dotnet build tests/Dekaf.Tests.Unit/Dekaf.Tests.Unit.csproj -c Release -f net8.0`
- `tests/Dekaf.Tests.Unit/bin/Release/net10.0/Dekaf.Tests.Unit.exe --no-ansi --progress off --reflection --timeout 4m` (4478 passed)
- `tests/Dekaf.Tests.Unit/bin/Release/net8.0/Dekaf.Tests.Unit.exe --no-ansi --progress off --reflection --timeout 10m --maximum-parallel-tests 1` (4472 passed)
- changed `CachingStringDeserializerTests` on net8.0 by UID loop (10 passed)

Closes #1560

### DIFF
--- a/src/Dekaf.SchemaRegistry/SubjectSchemaIdCache.cs
+++ b/src/Dekaf.SchemaRegistry/SubjectSchemaIdCache.cs
@@ -4,7 +4,7 @@ namespace Dekaf.SchemaRegistry;
 
 internal sealed class SubjectSchemaIdCache
 {
-    // Match CachingStringKeyDeserializer: fixed topic sets stay cached,
+    // Match CachingStringDeserializer: fixed topic sets stay cached,
     // dynamic topic names cannot grow without bound.
     internal const int MaxCachedEntries = 16_384;
 
--- a/src/Dekaf/Builders.cs
+++ b/src/Dekaf/Builders.cs
@@ -1296,6 +1296,9 @@ public sealed class ProducerBuilder<TKey, TValue>
 /// <typeparam name="TValue">Value type.</typeparam>
 public sealed class ConsumerBuilder<TKey, TValue>
 {
+    private const int DefaultMaxCachedStringValueBytes = 4 * 1024;
+    private const int DefaultMaxCachedStringValueEntries = 128;
+
     private readonly KafkaClientInfrastructure? _clientInfrastructure;
     private IReadOnlyList<string> _bootstrapServers = [];
     private string? _clientId;
@@ -1332,6 +1335,9 @@ public sealed class ConsumerBuilder<TKey, TValue>
     private AwsMskIamConfig? _awsMskIamConfig;
     private IDeserializer<TKey>? _keyDeserializer;
     private IDeserializer<TValue>? _valueDeserializer;
+    private bool _cacheStringValues;
+    private int _maxCachedStringValueBytes = DefaultMaxCachedStringValueBytes;
+    private int _maxCachedStringValueEntries = DefaultMaxCachedStringValueEntries;
     private IRebalanceListener? _rebalanceListener;
     private Microsoft.Extensions.Logging.ILoggerFactory? _loggerFactory;
     private bool _enablePartitionEof;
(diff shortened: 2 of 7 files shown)


thomhurst:
Addressed the blocking review in d5069638:

- value caching is no longer default; `Kafka.CreateConsumer<TKey, string>()` keeps the plain built-in string deserializer unless the caller opts in with `WithCachedStringValues()`.
- added a default-builder test that proves repeated values are not cached by default, plus an opt-in test that proves bounded caching is applied when requested.
- added the requested regression benchmark for the non-repeating/saturated-cache shape: default string deserialize is 45.033 ns / 80 B; saturated cached deserialize is 51.270 ns / 80 B. Since caching is now opt-in, default consumers avoid that overhead.
- typed stress scenarios now call `WithCachedStringValues()` because they intentionally replay a fixed pre-seeded payload. Repeated 1KB values drop from 2064 B/msg to 40 B/msg in the benchmark.

Verified builds for Dekaf, benchmarks, stress tests, unit tests net10/net8, full net10 unit, full net8 unit with `--maximum-parallel-tests 1`, changed cache tests on net8, and the updated BenchmarkDotNet run.

PR #1571, opened 2026-07-08
fix(producer): validate batch generation before acting on pooled ReadyBatch references

### LINKED ISSUE #1570: Producer batch recycle race: stale retry references pass IsReturnedToPool after re-rent, causing framing-guard failures and message loss
## Summary

Stress run [28939906458](https://github.com/thomhurst/Dekaf/actions/runs/28939906458/job/85859279944) (main @ `bc02bafd`, includes the #1559 framing guard from #1568) failed the FnF 1-broker producer job on the message-loss gate:

```
MESSAGE LOSS DETECTED — failing the run:
  Dekaf producer: 56 errors
  Dekaf producer: 48,704 messages accepted but never delivered
```

48,704 / 56 = ~870 msgs per batch — exactly 56 whole batches. The wire stayed clean (broker log has zero errors, no torn frames, client sustained 1.16M msg/s for the full 15 minutes). The #1559 guard fired 56 times and discarded the requests before they reached the socket — corruption converted to local loss, as designed. The underlying race it guards against is still live.

## Root cause

A pooled batch is returned to the pool and re-rented while a stale reference still treats it as owned:

1. `_preEncodedRecords` is only assigned in `PartitionBatch.Complete()` on a **freshly rented** `RecordBatch` (`RecordAccumulator.cs`). The size pass and write pass in `ProduceRequestPartitionData.Write` are back-to-back in one synchronous call — a declared/emitted length mismatch is only possible if the same pooled object was recycled and re-rented mid-serialization (pool `Reset` clears `_preEncodedRecords`; the new renter installs a different-length slice).
2. Stale `ReadyBatch` references survive in retry paths (`_sendFailedRetries`, partition carry-over). The staleness guards (`BrokerSender.CoalesceBatch` and the send-failure retry loop) check `batch.IsReturnedToPool` — but `RecordBatch.RentFromPool()` **resets `_returnedToPoolFlag` to 0**. A stale holder inspecting a re-rented object sees "not returned" and coalesces/serializes a batch now owned by a different partition. Two owners, one object.
3. The guard's `CorruptMessage` exception lands in the generic send-failure catch, which queues the batch for retry with stolen buffers. Retries can never succeed, so each batch eventually fails on delivery timeout → 56 errors.

## Fix direction

`IsReturnedToPool` is fundamentally unsound for staleness detection because the flag is reset on re-rent. Replace it with a monotonic **generation stamp**: increment on every rent, have any holder that can outlive ownership capture the generation at enqueue time and compare before use. Alternatively (or additionally), hard-remove a batch from every retry/carry-over structure before `Cleanup()` recycles it.

## Lineage

- #1516 / earlier: BufferMemory refund leak (fixed)
- #1555: poison connection on torn write (insufficient)
- #1562: never cancel a frame write mid-frame (ruled out mid-write tear)
- #1559 / #1568: framing guard — fails PRODUCE locally on length mismatch (working; converted wire corruption into bounded local loss)
- This issue: the actual pooled-object recycle race

thomhurst:
Fixes #1570

## Problem

Stress run [28939906458](https://github.com/thomhurst/Dekaf/actions/runs/28939906458) failed with 56 errors and 48,704 accepted-but-never-delivered messages — exactly 56 batches. Each error was the #1568 framing guard firing (`declared records length != emitted bytes`), which is only possible when a pooled `RecordBatch` is recycled and re-rented mid-serialization.

The staleness defense for pooled `ReadyBatch` references was `IsReturnedToPool` — but `Initialize()` resets that flag on re-rent. A stale reference sitting in `_sendFailedRetries` (or held by a send task racing a disposal/completion) that inspects a re-rented object sees "not returned" and acts on a batch now owned by a different partition: re-sending it (framing mismatch, previously wire corruption), completing it (foreign batch falsely acknowledged), or failing it (foreign batch falsely failed → message loss).

## Fix

Extend the generation-stamp pattern already used by `PendingResponse.IsSameIncarnation` to every place a `ReadyBatch` reference can outlive ownership:

- **`ReadyBatch.IsCurrentIncarnation(int)`** — combines the pool flag with the monotonic generation stamp. Read order (flag, then generation) pairs with the new write order in `Initialize()` (generation increment **before** flag clear) so a reader can never observe a cleared flag with a stale generation.
- **`SendCoalescedAsync`** snapshots generations before its first await, while ownership handed over by the coalesce step is still exclusive. All later touches validate against the snapshot: fire-and-forget completion/cleanup loops, the shutdown fail path, and the send-failure retry path.
- **`_sendFailedRetries`** entries now carry the captured generation; both dequeue sites (retry pickup and disposal drain) validate before acting.
- **`PendingResponse.Create`** now receives the send-start snapshot instead of re-capturing after the send — a later capture can stamp a foreign incarnation and then complete it when the response arrives.

Stale entries are skipped with a warning log (`LogStaleRetryBatchSkipped` / `LogStaleBatchInSendSkipped`) so future stress runs surface any remaining double-ownership path by name instead of by corruption.

## Testing

- New `ReadyBatchIncarnationTests` (5 tests, passing) pin the regression: a returned-and-re-rented batch passes the old `IsReturnedToPool` check but fails `IsCurrentIncarnation`, and `Initialize()` ordering keeps the flag/generation observable states consistent.
- Full test suites + stress validation left to CI per request. The real validation is the next scheduled stress run: the framing guard + these skip logs should report zero occurrences.

## Notes

- Allocation impact: one pooled `int[]` rent/return per coalesced send (per-batch-group, not per-message); the rent that previously happened in `PendingResponse.Create` moved to send start, so the acks!=None path count is unchanged.

### DIFF
--- a/src/Dekaf/Producer/BrokerSender.cs
+++ b/src/Dekaf/Producer/BrokerSender.cs
@@ -161,7 +161,8 @@ internal sealed partial class BrokerSender : IAsyncDisposable
         /// that was present when this PendingResponse was created.
         /// </summary>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public bool IsSameIncarnation(int i) => Batches[i] is not null && Batches[i].Generation == BatchGenerations[i];
+        public bool IsSameIncarnation(int i)
+            => Batches[i] is not null && Batches[i].IsCurrentIncarnation(BatchGenerations[i]);
 
         /// <summary>
         /// Clears batch references and returns pooled arrays to <see cref="ArrayPool{T}.Shared"/>.
@@ -181,19 +182,38 @@ internal sealed partial class BrokerSender : IAsyncDisposable
         Unmute
     }
 
+    private readonly struct BatchReference
+    {
+        public readonly ReadyBatch Batch;
+        public readonly int Generation;
+
+        public BatchReference(ReadyBatch batch, int generation)
+        {
+            Batch = batch;
+            Generation = generation;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool IsCurrentIncarnation() => Batch.IsCurrentIncarnation(Generation);
+    }
+
     [StructLayout(LayoutKind.Auto)]
     private readonly struct SendLoopEvent
     {
         public readonly SendLoopEventType Type;
         public readonly ReadyBatch? Batch;
+        public readonly int BatchGeneration;
 
-        private SendLoopEvent(SendLoopEventType type, ReadyBatch? batch = null)
+        private SendLoopEvent(SendLoopEventType type, ReadyBatch? batch = null, int batchGeneration = 0)
         {
             Type = type;
             Batch = batch;
+            BatchGeneration = batchGeneration;
         }
 
-        public static SendLoopEvent NewBatch(ReadyBatch batch) => new(SendLoopEventType.NewBatch, batch);
+        public static SendLoopEvent NewBatch(ReadyBatch batch) => new(SendLoopEventType.NewBatch, batch, batch.Generation);
+
+        public BatchReference GetBatchReference() => new(Batch!, BatchGeneration);
 
         public static SendLoopEvent ResponseReady() => new(SendLoopEventType.ResponseReady);
 
@@ -207,11 +227,13 @@ internal sealed partial class BrokerSender : IAsyncDisposable
     private struct ConnectionBucket
     {
         public ReadyBatch[] Batches;
+        public int[] Generations;
         public int Count;
 
         public void Clear()
         {
             Array.Clear(Batches, 0, Count);
+            Array.Clear(Generations, 0, Count);
             Count = 0;
         }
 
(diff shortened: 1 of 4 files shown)


thomhurst:
Addressed the blocking review items in e95b99f:
- kept generation snapshots aligned with ProduceRequestScratch's in-place topic sort
- made PendingResponse use ReadyBatch.IsCurrentIncarnation so returned-to-pool batches fail liveness checks
- guarded post-await send/release paths and carried generations through coalesce, carry-over, retry, and pre-serialization refs
- added regression coverage for paired sort, PendingResponse returned-to-pool detection, and stale coalesce refs

Validation:
- dotnet build tests/Dekaf.Tests.Unit/Dekaf.Tests.Unit.csproj --configuration Release
- ./tests/Dekaf.Tests.Unit/bin/Release/net10.0/Dekaf.Tests.Unit
- dotnet build tests/Dekaf.Tests.Integration/Dekaf.Tests.Integration.csproj --configuration Release
- ./tests/Dekaf.Tests.Integration/bin/Release/net10.0/Dekaf.Tests.Integration --treenode-filter /*/*/ConcurrentAccessPatternTests/MultipleTopicProducers_SharedBaseProducer_IndependentTopics --maximum-parallel-tests 1 --log-level Debug
- same failing integration on net8.0

PR #1576, opened 2026-07-08
Fix producer batch recycling during produce writes

### LINKED ISSUE #1574: producer: batch recycle race persists after #1571 — stress runs still lose ~25-40 batches to CorruptMessage guard
## Evidence

Run [28946418064](https://github.com/thomhurst/Dekaf/actions/runs/28946418064) (main@9f7421d3, **includes #1571**) failed the loss gate on both producer jobs:

- **Acks All (1 broker):** 25 errors, 828,932,744 accepted vs 828,906,816 delivered → 25,928 lost = exactly ~25 batches × ~1037 msgs
- **3 Brokers:** 41,332 accepted-never-delivered (~40 batches), zero sampled errors

Identical signature to run 28939906458 (56 hits of the #1568 framing guard): size-pass vs write-pass mismatch in `ProduceRequestPartitionData.Write` → local `KafkaException(CorruptMessage)` at `ProduceRequest.cs:214` → batch fails, wire stays clean. The guard is doing its job; the underlying recycle race (#1559) is not fixed.

No `LogStaleBatchInSendSkipped`/`LogStaleRetryBatchSkipped` warnings visible — but the stress harness wires no ILogger (#1573), so this is inconclusive.

## Why #1571's generation stamps may be insufficient

#1571 validates `ReadyBatch.IsCurrentIncarnation` at coalesce, retry-dequeue, and PendingResponse sites. Three structural gaps remain:

1. **Detect-only, not preventive.** A generation check passes at time T; nothing prevents another thread recycling the batch at T+ε while serialization is mid-flight. The guard trip is the *detection* of exactly this.
2. **`ReadyBatch.Generation` does not protect the underlying `RecordBatch`/`BatchArena`.** They are pooled separately and returned in `ReadyBatch.Cleanup()` (`RecordAccumulator.cs:6226`): `BatchArena.ReturnToPool(_arena)` + `_recordBatch.ReturnToPool()`. Once the arena is re-rented, a producer thread appends into memory a stale serializer is still reading — generation on the wrapper can't see it.
3. **Asymmetric pre-serialization wait.** `FailCore` calls `WaitForPreSerializationIfStarted()` before cleanup; `CompleteSend` (`RecordAccumulator.cs:6040`) runs `Cleanup()` in its `finally` **without** waiting. The pre-serialization/PreCompress task (`PrepareBatchForPublish`, `RecordAccumulator.cs:3251`) runs on the thread pool with only bracketing generation checks around `PreCompress` — a concurrent `CompleteSend → Cleanup → arena re-rent` between those checks mutates bytes under the compressor/serializer.

## Fix direction

Detection can't win this race; ownership must. Options (roughly in order of preference):

- **Refcount/pin:** send path takes a pin on the ReadyBatch (covering RecordBatch+arena) before serialization; `Cleanup()` defers pool return until pins drop to zero.
- **Symmetric wait:** make `CompleteSend` also `WaitForPreSerializationIfStarted()` — closes the PreCompress window but not a send-path serialization overlap, so likely necessary-not-sufficient.
- **Single-owner recycle:** only the send loop that last serialized the batch may trigger pool return, i.e. hard-remove from every queue/pending structure *before* `Cleanup()` is reachable.

## Repro

Producer stress ≥15 min at ~1M msg/s (both 1-broker acks-all and 3-broker). Land #1573 first so the run names the racing path via stale-skip warnings and error samples.

Refs: #1559, #1568, #1570, #1571.

thomhurst:
## Summary
- keep producer request scratch data live until pipelined produce writes finish
- pin ReadyBatch pooled resources while pre-serialization and produce request serialization read RecordBatch/BatchArena data
- defer ReadyBatch resource cleanup until active pins release

## Test plan
- [x] dotnet build tests/Dekaf.Tests.Unit --configuration Release
- [x] ./tests/Dekaf.Tests.Unit/bin/Release/net10.0/Dekaf.Tests.Unit --treenode-filter "/*/*/ReadyBatchIncarnationTests/*"
- [x] ./tests/Dekaf.Tests.Unit/bin/Release/net10.0/Dekaf.Tests.Unit --treenode-filter "/*/*/BrokerSenderSendLoopTests/*"
- [x] ./tests/Dekaf.Tests.Unit/bin/Release/net10.0/Dekaf.Tests.Unit --treenode-filter "/*/*/KafkaConnectionPipelinedTests/*"
- [x] ./tests/Dekaf.Tests.Unit/bin/Release/net10.0/Dekaf.Tests.Unit

Closes #1574

### DIFF
--- a/src/Dekaf/Networking/IKafkaConnection.cs
+++ b/src/Dekaf/Networking/IKafkaConnection.cs
@@ -101,3 +101,20 @@ public interface IKafkaConnection : IAsyncDisposable
     /// </summary>
     ValueTask ConnectAsync(CancellationToken cancellationToken = default);
 }
+
+internal interface IKafkaPipelinedWriteCompletionConnection
+{
+    ValueTask<Task<TResponse>> SendPipelinedAfterWriteAsync<TRequest, TResponse>(
+        TRequest request,
+        short apiVersion,
+        CancellationToken cancellationToken = default)
+        where TRequest : IKafkaRequest<TResponse>
+        where TResponse : IKafkaResponse;
+
+    ValueTask<Task<TResponse>> SendPipelinedWithCallerTimeoutAfterWriteAsync<TRequest, TResponse>(
+        TRequest request,
+        short apiVersion,
+        CancellationToken cancellationToken = default)
+        where TRequest : IKafkaRequest<TResponse>
+        where TResponse : IKafkaResponse;
+}
--- a/src/Dekaf/Networking/KafkaConnection.cs
+++ b/src/Dekaf/Networking/KafkaConnection.cs
@@ -131,7 +131,10 @@ internal static class ConnectionHelper
 /// <summary>
 /// A multiplexed connection to a Kafka broker using System.IO.Pipelines.
 /// </summary>
-public sealed partial class KafkaConnection : IKafkaConnection, IIdleTrackedKafkaConnection
+public sealed partial class KafkaConnection :
+    IKafkaConnection,
+    IIdleTrackedKafkaConnection,
+    IKafkaPipelinedWriteCompletionConnection
 {
     private readonly string _host;
     private readonly int _port;
@@ -686,7 +689,11 @@ public sealed partial class KafkaConnection : IKafkaConnection, IIdleTrackedKafk
         CancellationToken cancellationToken = default)
         where TRequest : IKafkaRequest<TResponse>
         where TResponse : IKafkaResponse
-        => SendPipelinedCoreAsync<TRequest, TResponse>(request, apiVersion, callerOwnsTimeout: false, cancellationToken);
+        => SendPipelinedCoreAsync<TRequest, TResponse>(
+            request,
+            apiVersion,
+            callerOwnsTimeout: false,
+            cancellationToken);
 
     /// <summary>
     /// Pipelined overload that skips the per-write CancellationTokenSource rent and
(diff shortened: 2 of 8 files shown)


thomhurst:
Addressed the blocking review on the latest head (`10f89d9`):

- Replaced the silent pipelined fallback with a hard failure when a connection does not implement `IKafkaPipelinedWriteCompletionConnection`, so produce sends cannot silently downgrade to the unsafe pre-fix behavior.
- Reworked producer send-loop tests to use `TestKafkaConnection : IKafkaConnection, IKafkaPipelinedWriteCompletionConnection` and assert the old `SendPipelinedAsync` / caller-timeout paths are not used for pipelined produce sends.
- Added `SendLoop_PipelinedProduce_ReleasesBufferMemoryOnlyAfterWriteCompletes`, which gates write completion and proves buffer memory stays unreleased until the write-completion gate opens.
- Added a comment in `WaitForCleanupIfStarted()` explaining why reset must wait for deferred resource cleanup behind active pins.

Local validation:

- `Dekaf.Tests.Unit.exe --treenode-filter "/*/*/BrokerSenderSendLoopTests/*"` (13 passed)
- `Dekaf.Tests.Unit.exe --treenode-filter "/*/*/BrokerSenderMuteOrderingTests/*"` (11 passed)
- `Dekaf.Tests.Unit.exe --treenode-filter "/*/*/ReadyBatchIncarnationTests/*"` (9 passed)

PR #1648, opened 2026-07-09
test: add transactional crash-recovery harness

### LINKED ISSUE #1621: Transactional client crash-recovery harness and coverage
Part of #1597.

## Scope
Add a child-process integration harness that can be terminated mid-transaction, then verify recovery using the same `transactional.id`.

## Acceptance criteria
- A helper client signals only after its transactional write is broker-acknowledged and the transaction remains open.
- Test force-terminates that helper process without graceful disposal, restarts a producer with the same `transactional.id`, and verifies the dangling transaction is aborted.
- Replacement producer commits a new transaction successfully.
- `read_committed` verification proves no duplicates, no loss, and no visibility of the crashed client's dangling records.
- Child process lifecycle and diagnostics are bounded and cleaned up per test.

thomhurst:
## Summary
- add a child-process transactional producer that signals only after broker acknowledgment
- force-kill the client with its transaction open, then reuse the same `transactional.id`
- verify replacement commit recovery and exact `read_committed` visibility
- bound process waits, captured diagnostics, and signal-file cleanup

## Tests
- `dotnet build Dekaf.sln --configuration Release`
- `Dekaf.Tests.Unit` net10.0: 4,509 passed
- `TransactionCrashRecoveryTests` net10.0: 3 passed (Kafka 4.0/4.1/4.2)
- `TransactionCrashRecoveryTests` net8.0: 3 passed (Kafka 4.0/4.1/4.2)
- NativeAOT publish reached the platform linker; local Visual C++ linker is unavailable

Closes #1621

### DIFF
--- a/tests/Dekaf.Tests.Integration/TransactionCrashRecoveryTests.cs
+++ b/tests/Dekaf.Tests.Integration/TransactionCrashRecoveryTests.cs
@@ -1,0 +1,77 @@
+using Dekaf.Consumer;
+using Dekaf.Producer;
+using Dekaf.Protocol.Messages;
+
+namespace Dekaf.Tests.Integration;
+
+[Category("Transaction")]
+public sealed class TransactionCrashRecoveryTests(KafkaTestContainer kafka) : KafkaIntegrationTest(kafka)
+{
+    [Test]
+    public async Task ForceTerminatedProducer_ReplacementAbortsDanglingTransaction()
+    {
+        var topic = await KafkaContainer.CreateTestTopicAsync();
+        var transactionId = $"crash-recovery-{Guid.NewGuid():N}";
+        var crashedValue = $"crashed-{Guid.NewGuid():N}";
+        string[] committedValues =
+        [
+            $"committed-{Guid.NewGuid():N}",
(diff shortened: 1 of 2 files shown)


thomhurst:
Follow-up on the TOCTOU observation: commit 196031bf adds deterministic coverage of the exact interleaving (the child exits after `HasExited` and before `Kill`). `TerminateAsync` completes on both net8 and net10, with all 3 Kafka fixtures passing on each target.

The runtime deliberately makes this race idempotent: Windows `Kill` uses `throwIfExited: false` and returns for an invalid handle; Unix `Kill` returns when the process already exited or reports `ESRCH`, and `KillTree` also ignores `ESRCH`. Therefore no `InvalidOperationException` escapes, and replacing the successful no-op with a custom exception would misstate the outcome.

Runtime sources:
- https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.Process/src/System/Diagnostics/Process.Windows.cs
- https://github.com/dotnet/runtime/blob/main/src/libraries/System.Diagnostics.Process/src/System/Diagnostics/Process.Unix.cs

PR #1653, opened 2026-07-09
test(records): cover null and empty boundaries

### LINKED ISSUE #1627: Boundary tests: null and empty record fields and headers
Part of #1603.

## Scope

Pin Kafka wire semantics for null versus zero-length keys, values, and header values, including header-count edges.

## Acceptance criteria

- Null and empty keys remain distinguishable after produce/consume.
- Null and empty values remain distinguishable after produce/consume.
- Null and empty header values remain distinguishable after produce/consume.
- Zero headers, one header, duplicate header keys, and a high-but-valid header count round-trip in order without loss.
- Unit serialization coverage and a broker-backed integration path both exercise the cases.

thomhurst:
## Summary
- pin null versus zero-length key and value wire behavior
- cover null/empty header values plus zero, one, duplicate, and 1,024-header boundaries
- verify header order and values through a broker-backed produce/consume path

## Tests
- `Dekaf.Tests.Unit --treenode-filter '/*/*/RecordBatchTests/*'` (38 passed)
- `dotnet build tests/Dekaf.Tests.Integration --configuration Release`
- focused `RecordBoundaryRoundTripTests` (passed)
- `HeaderRoundTripTests` (15 passed)
- `TombstoneAndNullValueTests` (18 passed)

Closes #1627

### DIFF
--- a/tests/Dekaf.Tests.Integration/RecordBoundaryRoundTripTests.cs
+++ b/tests/Dekaf.Tests.Integration/RecordBoundaryRoundTripTests.cs
@@ -11,11 +11,13 @@ public sealed class RecordBoundaryRoundTripTests(KafkaTestContainer kafka) : Kaf
     public async Task ProduceConsume_PreservesNullEmptyAndHeaderBoundaries()
     {
         const int highHeaderCount = 1024;
+        // Reuse one key so this count boundary does not fill the process-global header-key cache.
+        const string repeatedHeaderKey = "boundary-high";
         var topic = await KafkaContainer.CreateTestTopicAsync();
         var highHeaders = new Headers(highHeaderCount);
         for (var i = 0; i < highHeaderCount; i++)
         {
-            highHeaders.Add($"boundary-high-{i:D4}", [(byte)(i % 251)]);
+            highHeaders.Add(repeatedHeaderKey, [(byte)(i % 251)]);
         }
 
         var messages = new ProducerMessage<string?, string?>[]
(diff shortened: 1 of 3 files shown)


thomhurst:
Fixed both failing unit jobs in 7d755966. The 1,024-header boundary cases now reuse one header key, preserving count/order coverage without filling the process-global 128-entry header-key cache and racing `Read_RepeatedKeyBytes_ReusesCachedString`. Full unit suites pass on net8.0 (4,504/4,504) and net10.0 before rebase; post-rebase net10.0 passes 4,515/4,515. The three Kafka-version integration cases pass.

thomhurst:
Addressed the review design suggestion in 5e8a3611. Split the bundled broker round trip into four independent tests (null fields, empty fields, duplicate headers, and 1,024 headers), so each scenario reports separately. Kept them together in RecordBoundaryRoundTripTests because each combines record-field and header boundaries; shared setup now lives in one round-trip helper. Header values are copied before pooled consumer memory is released. Validation: 12/12 boundary cases on net8.0 and net10.0, plus 27/27 net10.0 boundary/header integration cases.

thomhurst:
Addressed the cache-overflow coverage observation in 7affe294. Added a deterministic local Utf8StringInternCache regression at a one-entry limit: the cached key retains reference identity, while the next key still decodes correctly and remains uncached on repeated reads. This exercises the bounded fallback without polluting the process-global header cache. Both target-framework builds pass; focused test passes 1/1 on net8.0 and net10.0; Internal suites pass 78/78 on each.

PR #1658, opened 2026-07-09
test(producer): isolate activity listener assertions

### LINKED ISSUE #1609: Flaky: ProducerAsyncPreparerTests activity-status assertions fail intermittently under parallel test runs
## Symptom

`ProduceAsync_PreparerThrowsSynchronously_StopsAndErrorTagsActivity` and `ProduceAsync_PreparerFaultsAsynchronously_StopsAndErrorTagsActivity` (tests/Dekaf.Tests.Unit/Producer/ProducerAsyncPreparerTests.cs) fail intermittently:

```
AssertionException: Expected to be equal to Error
but received Ok
at Assert.That(started[0].Status).IsEqualTo(ActivityStatusCode.Error)
```

## Repro evidence (2026-07-09, current main)

- Full producer-namespace run (`--treenode-filter "/*/Dekaf.Tests.Unit.Producer/*/*"`, 714 tests, parallel): failed in ~2 of 5 runs — reproduced on an unmodified checkout, so this is pre-existing on main, not introduced by in-flight work
- Same class run in isolation 10×: 0 failures

## Analysis

Fails only under parallel execution → almost certainly shared `ActivityListener`/`ActivitySource` state between concurrently running tests: another test's listener or activity interleaves, and `started[0]` picks up an activity whose status was set by a different code path (or the listener races the status write).

Per project policy, flaky = real bug: either the test needs isolation (`[NotInParallel]` group for activity-listener tests, filter activities by source + operation name and by the specific topic used) or there is a genuine race in how the producer stamps activity status on preparer failure. Determine which before suppressing with NotInParallel.

thomhurst:
## Summary
- filter preparer-test activity callbacks to the operation under test
- add a deterministic regression proving unrelated Dekaf activities are ignored

## Test plan
- [x] `dotnet build tests/Dekaf.Tests.Unit --configuration Release`
- [x] `ProducerAsyncPreparerTests` on net8.0 and net10.0 (5/5 each)
- [x] full producer namespace on net8.0 and net10.0 (713/713 each)

Closes #1609

### DIFF
--- a/tests/Dekaf.Tests.Unit/Producer/ProducerAsyncPreparerTests.cs
+++ b/tests/Dekaf.Tests.Unit/Producer/ProducerAsyncPreparerTests.cs
@@ -9,11 +9,29 @@ namespace Dekaf.Tests.Unit.Producer;
 
 // Drives the real KafkaProducer async produce path with a serializer that implements
 // IAsyncSerializerPreparer<T>, exercising the preparer gate in ProduceAsyncCore. Uses a
-// process-wide ActivityListener, so it must not run alongside other producing tests.
+// process-wide ActivityListener, so it must not run alongside other listener tests.
 [NotInParallel("ActivityListener")]
 public class ProducerAsyncPreparerTests
 {
     private const string Topic = "prepare-topic";
+    private const string PublishActivityName = Topic + " publish";
+
+    [Test]
+    public async Task ActivityListener_IgnoresOtherDekafOperations()
+    {
+        var (started, stopped, listener) = ListenForActivities();
+        using (listener)
+        {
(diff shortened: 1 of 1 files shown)


thomhurst:
Addressed the concurrency-safety observation in ff215fe2. Activity callbacks now enqueue into ConcurrentQueue<Activity>; a new 10,000-operation parallel regression reproduced the old List loss (8,974/10,000) and passes with the fix. Validation: both-TFM Release build clean; ProducerAsyncPreparerTests 6/6 on net8/net10; producer namespace 714/714 on both; concurrent regression 10/10 repeated. Full net8 reached 4,504/4,505 with only the known unrelated RecordBatch pool race tracked by #1665/#1674, so it was not rerun.

PR #1684, opened 2026-07-10
Add golden response reader snapshots

### LINKED ISSUE #1656: Snapshot Kafka response decoding from canned bytes
Part of #1600.

Add golden Verify.TUnit snapshots for Kafka response readers, building on the shared snapshot conventions from the request-writer child.

Acceptance criteria:
- Keep deterministic canned response bytes for every concrete Kafka response type at every supported API version.
- Decode each fixture through the production reader and snapshot the resulting object graph.
- Normalize pooled/disposable/runtime-only state so snapshots remain stable while retaining protocol fields.
- Include a coverage guard so new response types or supported versions cannot silently lack fixtures.
- Focused and full unit tests pass.

thomhurst:
## Summary
- add 125 checked-in Kafka response body fixtures covering all 64 concrete response types and supported versions
- decode fixtures through `KafkaMessageMetadata<TRequest, TResponse>.ReadResponse` and snapshot stable protocol object graphs plus body hex
- normalize pooled `ProduceResponse` and `FetchResponse` state and guard fixture coverage for future response types or versions

## Testing
- [x] focused response snapshots: net10 126/126
- [x] focused response snapshots: net8 126/126
- [x] full unit suite: net10 4,812/4,812
- [x] full unit suite: net8 4,790/4,790 (full rerun after one unrelated timeout; exact failed test passed independently)

Closes #1656


thomhurst:
Addressed the blocking review in `6a83b756`.

- Replaced seven degenerate fixtures with deterministic populated responses for Metadata, Fetch, Produce, DescribeConfigs, OffsetFetch, ListOffsets, and DescribeGroups. Coverage includes multiple topics/partitions, nested arrays, non-`None` errors, strings/UUIDs, KIP-951 leader/endpoint tags, unknown-tag skipping, and a real two-record Fetch `RecordBatch`.
- Added an independent response encoder plus an exact byte-for-byte resource test, so checked-in `.bin` fixtures remain reproducible without using the readers under test.
- Extracted `GetRequestInterface`, `GetStaticProperty<T>`, and `HexDump` into shared `WireFormatSnapshotSupport` used by request and response suites.
- Scoped response snapshots to the existing `FetchResponsePool` isolation key and explicitly release parsed Fetch record resources.

Validation: snapshot suites 253/253 on net8.0/net10.0; protocol suites 934/934 net8.0 and 940/940 net10.0; full unit suites 4,791/4,791 net8.0 and 4,813/4,813 net10.0.

PR #1728, opened 2026-07-11
Prevent partial record batches crossing partitions

### LINKED ISSUE #1716: Consumer permanently stalls: RecordBatch.Read overreads partition boundary on partial-batch stub <61 bytes, desyncing the whole fetch response
## Summary

`RecordBatch.Read` reads the full 61-byte v2 batch header **before** validating that `availableBytes` can hold it. When a fetch response partition's records slice ends in a partial-batch stub of **1–60 bytes** — which the broker legally produces, since `max.partition.fetch.bytes` is a byte cap, not batch-aligned — the header read walks **past the partition's records boundary into the next partition's bytes** and then throws. Because `KafkaProtocolReader` is forward-only, the boundary-restore code in `FetchResponse` cannot rewind: `remaining` is negative, the `if (remaining > 0) reader.Skip(remaining)` guard silently skips correction, and the reader stays misaligned. Every subsequent partition in the same response is parsed from garbage.

The consumer then discards the entire response (see linked issues), never advances `_fetchPositions`, and refetches the identical byte range — which reproduces the identical stub. Result: **permanent, silent consumer stall with continuous full-bandwidth fetching**.

This is the root cause of the `Consumer (Paired)` failure in stress run [29131579448](https://github.com/thomhurst/Dekaf/actions/runs/29131579448/job/86487804177) (324,306 messages delivered in ~1s, then zero delivery for 5 minutes; watchdog kill), and with high confidence also the `Dekaf Transactional EOS (3 Brokers)` failure in the same run (read_committed verifier delivered 22,574 of 284,900 committed messages then stalled to its 15-minute timeout).

## Affected code

- `src/Dekaf/Protocol/Records/RecordBatch.cs:989-1019` — header fields read unconditionally (lines 991–1009) before the truncation check at line 1014 (`maxRecordsLength = Math.Max(0, availableBytes - TotalBatchHeaderSize)`). A stub shorter than `TotalBatchHeaderSize` (61) overreads by up to 60 bytes before any guard fires.
- `src/Dekaf/Protocol/Messages/FetchResponse.cs:444-471` — the batch-read loop's catch/`break` paths are followed by:
  ```csharp
  var remaining = (int)(recordsEndPosition - reader.Consumed);
  if (remaining > 0) reader.Skip(remaining);
  ```
  which cannot correct an overshoot (negative `remaining`), leaving the reader desynced for all following partitions in the response.

## Reproduction (protocol-layer, real library code)

A standalone harness built one real batch via `RecordBatch.Write`, appended a truncated copy as the trailing stub inside a correctly-encoded v13 fetch partition, followed by sentinel bytes standing in for the next partition:

```
stub=100: parsed, batches=1, consumed=1217, expected=1217 -> OK (aligned)
stub= 61: parsed, batches=1, consumed=1178, expected=1178 -> OK (aligned)
stub= 60: THREW MalformedProtocolDataException: Malformed variable-length integer (consumed=1182, expected=1177)
stub= 30: THREW MalformedProtocolDataException: Malformed variable-length integer (consumed=1182, expected=1147)
stub= 10: THREW MalformedProtocolDataException: Malformed variable-length integer (consumed=1138, expected=1127)
```

`consumed > expected` = the reader ended past the partition boundary.

## Why it wedges permanently in the consumer

1. The overshoot makes `FetchResponse.Read` either throw (whole response discarded by the prefetch error handler as "transient", including partitions that parsed fine) or produce partitions with empty `Records`.
2. Empty/discarded partitions never reach `WritePrefetchedItemsAsync`, so `UpdateFetchPositionsFromPrefetch` never runs → `_fetchPositions` frozen.
3. The next fetch requests the same offsets with the same size → broker returns the byte-identical slice → identical stub → loop. No error surfaces to the application; heartbeats keep the group member alive indefinitely.

All six partitions freeze simultaneously because one broker serves all six in a single response: the first partition that overshoots poisons the rest, and uniform record sizes make the stub residue roll together across partitions/adaptive-size steps.

## Evidence from the failed run

- Watchdog stack captures (job artifact `results-consumer-1brokers`) show `MessageCount: 324306` frozen across both the 30s and 301s captures while a thread continuously executes `FetchResponse.Read → RecordBatch.Read` and `SocketPipe.ReadPumpAsync` stays busy.
- Gen2 GC ~85 collections/sec during the stall (LOH churn from re-parsed multi-MB responses; amplified by the fetch-branch buffer leak, filed separately).
- Broker logs show a perfectly stable group: joined at epoch 2, no leave, no session expiry, no errors — confirming the stall is entirely client-side.
- The same seeded topic was consumed flawlessly 495 times by the Confluent client in the same process immediately before, ruling out on-broker data corruption.

## Fix direction (Java-client parity)

In the batch-read loop (`FetchResponse.cs:444`), peek/bail **before** consuming when `recordsEndPosition - reader.Consumed < RecordBatch.TotalBatchHeaderSize` (and ideally also peek `batchLength` and bail when the full batch isn't present — mirroring `ByteBufferLogInputStream` in the Java client). Additionally/alternatively, parse the records section through a sub-reader bounded to the records slice so an overread can never cross a partition boundary.

## Test gap

`tests/Dekaf.Tests.Unit/Protocol/RecordBatchTests.cs:1154-1299` only covers truncation with `availableBytes >= TotalBatchHeaderSize`. Add cases for stubs of 1–60 bytes, and a FetchResponse-level test asserting reader alignment (`consumed == recordsEndPosition`) and that subsequent partitions in the same response still parse.

thomhurst:
## Summary
- stop `RecordBatch.Read` before parsing headers that do not fit inside the partition records slice
- consume only the trailing partial-batch stub and preserve the next partition boundary
- cover every 1-60 byte header stub and a two-partition fetch-response regression

## Test plan
- [x] focused boundary tests (2 passed)
- [x] `RecordBatchTests` (41 passed)
- [x] complete unit suite (5,033 passed)

Closes #1716

### DIFF
--- a/src/Dekaf/Protocol/Records/RecordBatch.cs
+++ b/src/Dekaf/Protocol/Records/RecordBatch.cs
@@ -988,6 +988,16 @@ public sealed class RecordBatch : IDisposable
     /// </remarks>
     public static RecordBatch Read(ref KafkaProtocolReader reader, CompressionCodecRegistry? codecs = null, int availableBytes = int.MaxValue)
     {
+        if (availableBytes < TotalBatchHeaderSize)
+        {
+            if (availableBytes > 0)
+            {
+                reader.Skip(availableBytes);
+            }
+
+            throw new InsufficientDataException();
+        }
+
         var baseOffset = reader.ReadInt64();
         var batchLength = reader.ReadInt32();
         var partitionLeaderEpoch = reader.ReadInt32();
--- a/tests/Dekaf.Tests.Unit/Protocol/FetchResponseBatchBoundaryTests.cs
+++ b/tests/Dekaf.Tests.Unit/Protocol/FetchResponseBatchBoundaryTests.cs
@@ -1,0 +1,147 @@
+using System.Buffers;
+using Dekaf.Protocol;
+using Dekaf.Protocol.Messages;
+using Dekaf.Protocol.Records;
+
+namespace Dekaf.Tests.Unit.Protocol;
+
+public sealed class FetchResponseBatchBoundaryTests
+{
+    [Test]
+    public async Task RecordBatch_Read_HeaderStubs_DoNotCrossAvailableBoundary()
+    {
+        for (var stubLength = 1; stubLength < RecordBatch.TotalBatchHeaderSize; stubLength++)
+        {
+            var (threw, consumed) = ReadHeaderStub(stubLength);
+            await Assert.That(threw).IsTrue();
+            await Assert.That(consumed).IsEqualTo(stubLength);
+        }
+    }
+
+    [Test]
+    public async Task FetchResponse_Read_HeaderStub_PreservesNextPartitionAlignment()
+    {
+        var firstBatch = CreateBatchBytes(baseOffset: 10, "first");
+        var secondBatch = CreateBatchBytes(baseOffset: 20, "second");
+        var firstPartitionRecords = new byte[firstBatch.Length + 60];
+        firstBatch.CopyTo(firstPartitionRecords, 0);
+        firstBatch.AsSpan(0, 60).CopyTo(firstPartitionRecords.AsSpan(firstBatch.Length));
+
+        var buffer = new ArrayBufferWriter<byte>();
+        var writer = new KafkaProtocolWriter(buffer);
+        writer.WriteInt32(0);                     // throttle time
+        writer.WriteInt16((short)ErrorCode.None); // error code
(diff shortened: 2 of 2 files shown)


thomhurst:
Merge gate denied because the clear review summary contains a `### Correctness` heading that the actionable-finding heuristic flags. No merge bypass attempted; leaving for a later iteration after gate handling is corrected.

PR #1730, opened 2026-07-11
Surface malformed fetch record batches

### LINKED ISSUE #1717: FetchResponse.Read silently swallows record-batch parse exceptions (Trace.WriteLine only) — corruption is indistinguishable from empty data
## Summary

`FetchResponse.Read`'s per-partition record-batch loop converts **any** unexpected parse exception into an invisible no-op:

```csharp
// src/Dekaf/Protocol/Messages/FetchResponse.cs:457-463
catch (Exception ex)
{
    // Unexpected error parsing a record batch — log for visibility and skip remaining batches.
    Trace.WriteLine($"Dekaf: Unexpected exception parsing RecordBatch: {ex}");
    break;
}
```

`Trace.WriteLine` goes nowhere in production (and in the stress harness, which attaches no logger at all). The partition silently ends up with whatever batches parsed before the failure — possibly zero — and the caller cannot distinguish "no data" from "data present but unparseable".

This swallow is what turned the parse defect in #1716 into a **silent permanent hang** instead of a loud failure: the consumer saw empty records, kept its fetch position, and refetched the same bytes forever with no error, no log line, no metric.

## Why this is P0-adjacent

For a client library with zero-tolerance correctness goals, a batch that fails to parse is either (a) a client bug or (b) wire/broker corruption. Both must surface. Today the failure mode is indistinguishable from an idle topic.

## Fix direction

- Replace `Trace.WriteLine` with a real signal: propagate a typed error (e.g. `MalformedProtocolDataException` with topic/partition/offset context) or record it on the partition result so the consumer layer can raise `ConsumeException` / a metric.
- Keep the `InsufficientDataException` catch (legitimate trailing partial batch) but make it boundary-safe (see #1716 — currently the exception can fire *after* the reader has already crossed the partition boundary).
- Add a `DekafMetrics` counter for batch parse failures so CI/ops can alert on nonzero values.

## Related

- #1716 (root-cause overread this swallow masked)

thomhurst:
## Summary
- propagate unexpected record-batch parsing failures as typed `MalformedProtocolDataException`
- include partition context while preserving original exception as inner cause
- count failures with `messaging.consumer.batch.parse.errors`
- retain `InsufficientDataException` handling for legitimate truncated trailing batches

## Test plan
- [x] `dotnet build tests/Dekaf.Tests.Unit/Dekaf.Tests.Unit.csproj --configuration Release --no-restore`
- [x] malformed fetch response regression (typed error + metric)
- [x] `RecordBatchTests` (41 passed)
- [x] `ResponseWireFormatSnapshotTests` (127 passed)
- [x] `LeaderDiscoveryResponseTests` (2 passed)

Closes #1717

### DIFF
--- a/src/Dekaf/Consumer/KafkaConsumer.cs
+++ b/src/Dekaf/Consumer/KafkaConsumer.cs
@@ -2,6 +2,7 @@ using System.Buffers;
 using System.Collections.Concurrent;
 using System.Diagnostics;
 using System.Diagnostics.Metrics;
+using Dekaf.Errors;
 using System.Runtime.CompilerServices;
 using System.Runtime.ExceptionServices;
 using Dekaf.Compression;
@@ -82,6 +83,7 @@ internal sealed class PendingFetchData : IDisposable
     private int _headerGeneration;
     private bool _eagerParsed;
     private bool _hasBufferedCurrent;
+    private ConsumeException? _error;
 
     public string Topic { get; private set; } = null!;
     public int PartitionIndex { get; private set; }
@@ -167,6 +169,17 @@ internal sealed class PendingFetchData : IDisposable
             }
         }
 
+        return instance;
+    }
+
+    public static PendingFetchData CreateError(string topic, int partitionIndex, ConsumeException error)
+    {
+        var instance = Rent();
+        instance.Topic = topic;
+        instance.PartitionIndex = partitionIndex;
+        instance.TopicPartition = new TopicPartition(topic, partitionIndex);
+        instance._batches = Array.Empty<RecordBatch>();
+        instance._error = error;
         return instance;
     }
 
@@ -279,6 +292,9 @@ internal sealed class PendingFetchData : IDisposable
     /// </summary>
     public void EagerParseAll()
     {
+        if (Interlocked.Exchange(ref _error, null) is { } error)
+            throw error;
+
         if (_eagerParsed)
             return;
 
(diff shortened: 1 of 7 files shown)


thomhurst:
Addressed blocking review in d3edb4f7: malformed batches now stay partition-local, remaining response partitions still parse, pooled partial batches/lists are cleaned up, and KafkaConsumer surfaces a one-shot topic/partition-aware ConsumeException through all consume APIs. Added same-response healthy-sibling coverage plus consumer-facing typed-error coverage.

thomhurst:
Addressed latest blocking review: both streaming and single-message eager-parse paths now remove and dispose the same queued error item before rethrowing. Added regressions proving queue count drops immediately and the next poll consumes the following good fetch. Recovery tests: 13 passed; full unit suite: 5,034 passed.

thomhurst:
CI diagnosis: net10 job failed only because \ConsumerAssignmentFastPathTests.EnsureAssignmentForPollAsync_SlowPositionInitialization_DoesNotExpireMember\ timed out (~2.2s); 5,037 other unit tests passed. This is unrelated to the PR and reproduced as a timing flake on another branch locally. Retried failed job; no source change.

thomhurst:
Fixed the rerun's real pool-test race in 09b765f9. \FetchResponsePoolingTests\ asserted exact identity from global pools but used an isolated named parallel group, allowing unrelated pool users to interleave returns. It is now globally \NotInParallel\. The failing test passes on net8 and net10; both targets build cleanly.

PR #1740, opened 2026-07-11
Stop fabricating offset zero for unknown positions

### LINKED ISSUE #1723: Silent offset-0 fallbacks: ResolveOffsetAsync returns 0 when no connection lease available; BuildFetchResult defaults missing fetch positions to 0
## Summary

Two code paths silently substitute **offset 0** when they cannot determine the real fetch position. On a partition with existing data, fetching from 0 means mass re-delivery of already-consumed records (duplicates at scale) — or, for a consumer configured `AutoOffsetReset.Latest`, delivering the entire historical partition it explicitly asked to skip.

1. **`ResolveOffsetAsync` returns 0 when no connection lease is available** (`src/Dekaf/Consumer/KafkaConsumer.cs:4467-4468`):
   ```csharp
   var connectionLease = await GetPartitionLeaderControlConnectionAsync(partition, cancellationToken).ConfigureAwait(false);
   if (connectionLease is null)
       return 0;
   ```
   This runs inside `RetryHelper.WithRetryAsync`, but `return 0` is a *successful* result — the retry helper does not retry it. The caller (`:4450-4457`) then persists 0 into `_fetchPositions` and `SetPosition(...)` for a position that was the `-1`/`-2` auto-offset-reset sentinel. A transient lease shortage (pool contention, connection churn, adaptive scaling) permanently resolves `latest` to `0`.

2. **`BuildFetchResult` defaults missing positions to 0** (`KafkaConsumer.cs:5797`):
   ```csharp
   FetchOffset = fetchPositions.GetValueOrDefault(tp, 0),
   ```
   Any window where a partition is dispatched before its `_fetchPositions` entry exists (or after it was removed) issues a real broker fetch from offset 0. Interacting with `ResetOffsetOutOfRangeAsync` storing sentinel values into `_fetchPositions`/`_positions` (`:5441-5443`), this can also produce an offset-out-of-range ping-pong.

## Fix direction

- (1): treat "no lease available" as a retriable failure — throw a retriable exception inside the retry lambda instead of returning a fabricated offset; if retries exhaust, propagate an error rather than fabricate 0.
- (2): skip dispatching partitions with no known fetch position (or assert), rather than defaulting to 0. A missing entry is a state bug; fetching from 0 hides it as data duplication.

Found during the investigation of stress run 29131579448 (not the cause of that hang, but adjacent silent-fallback hazards on the same path).

thomhurst:
## Summary
- turn unavailable leader connections and missing `ListOffsets` partitions into retriable Kafka errors instead of successful offset `0`
- retry through existing metadata-refresh policy and propagate after exhaustion
- omit partitions with unknown fetch positions rather than dispatching broker fetches from offset `0`
- omit empty request topics and return their pooled partition lists immediately

## Test plan
- [x] build-result/pool tests (18 passed)
- [x] timing-sensitive unrelated failures verified isolated (2 passed)
- [x] complete unit suite (5,035 passed)
- [ ] integration tests (Docker daemon unavailable; `docker info` timed out)

Closes #1723

### DIFF
--- a/src/Dekaf/Consumer/KafkaConsumer.cs
+++ b/src/Dekaf/Consumer/KafkaConsumer.cs
@@ -2078,7 +2078,7 @@ public sealed partial class KafkaConsumer<TKey, TValue> :
                 }
                 catch (Exception ex)
                 {
-                    consecutiveErrors++;
+                    consecutiveErrors = PrefetchLoopControl.RecordConsecutiveError(consecutiveErrors, ex);
                     LogPrefetchLoopError(ex);
 
                     if (PrefetchLoopControl.ShouldBreakOnConsecutiveError(
--- a/src/Dekaf/Consumer/KafkaConsumer.cs
+++ b/src/Dekaf/Consumer/KafkaConsumer.cs
@@ -4256,7 +4256,9 @@ public sealed partial class KafkaConsumer<TKey, TValue> :
 
                 // Set equality alone is insufficient: the assignment can change away and back
                 // between polls. Unseen revocations require stale-fetch cleanup and position reset.
-                if (_assignment.SetEquals(coordinatorAssignment) && coordinatorRevocations is null)
+                if (_assignment.SetEquals(coordinatorAssignment)
+                    && coordinatorRevocations is null
+                    && HasInitializedFetchPositions(coordinatorAssignment))
                 {
                     Volatile.Write(ref _lastCoordinatorAssignmentVersion, coordinatorAssignmentVersion);
                     coordinator.AcknowledgeAssignmentSync(coordinatorAssignmentVersion);
@@ -4267,7 +4269,7 @@ public sealed partial class KafkaConsumer<TKey, TValue> :
                 List<TopicPartition>? newPartitions = null;
                 foreach (var partition in coordinatorAssignment)
                 {
-                    if (!_assignment.Contains(partition))
+                    if (!_assignment.Contains(partition) || !_fetchPositions.ContainsKey(partition))
                     {
                         newPartitions ??= new List<TopicPartition>();
                         newPartitions.Add(partition);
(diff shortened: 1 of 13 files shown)


thomhurst:
Addressed integration-test blocker in 1b5a77db. Added a NetworkPartition/Testcontainers scenario that pauses Kafka before first offset initialization, asserts the unavailable resolution propagates, then heals the broker and waits until \Latest\ resolves to offset 1 before producing. It verifies only the post-recovery record at offset 1 is consumed—fabricating offset 0 would replay the preexisting record and fail. Integration project builds cleanly; Docker unavailable locally, so CI will run the container test.

thomhurst:
CI follow-up in d7604ca8: paused broker surfaced the expected \ListOffsets\ failure as transport \TimeoutException\, not \KafkaException\; test now accepts either transport timeout or Kafka error, while still requiring post-heal \Latest\ position 1 and consumption of only offset 1. Integration project builds cleanly. Separate consume job exceeded its global job deadline after completed suites passed; fresh push reruns it.

thomhurst:
Addressed Codex P1 thread in 2fea7a19. Coordinator assignment fast path now requires initialized fetch positions; partitions already present in \_assignment\ but missing \_fetchPositions\ are re-added to the initialization set. Added regression proving an initial position failure leaves assignment published but the next unchanged-assignment poll retries and restores offset 10. net10 build and both new/unchanged fast-path tests pass.

thomhurst:
Addressed the current timeout-normalization P2 in 2877df9a at the source: KafkaConnection flush and receive timeouts now carry ErrorCode.RequestTimedOut, so RetryHelper consistently treats every ListOffsets transport-timeout path as retriable. Added assertions for both paths; build clean and KafkaConnection focused tests pass 46/46.

thomhurst:
CI retry reproduced EnsureAssignmentForPollAsync_SlowPositionInitialization_DoesNotExpireMember identically, so fixed its root race in 547abb77. The test now explicitly phases heartbeat assignment (partition 0, then 0+1) and installs the blocking OffsetFetch only after initial assignment, instead of depending on nondeterministic mock call counts. Target passes 10/10 fresh-process repetitions; build clean.

PR #1743, opened 2026-07-11
Verify consumed record batch CRC-32C

### LINKED ISSUE #1727: Consumer never verifies RecordBatch CRC — silent corruption is deserialized into user records
## Summary

`RecordBatch.Read` (`src/Dekaf/Protocol/Records/RecordBatch.cs`) reads and stores the batch CRC-32C but never verifies it against the batch contents. The Java client validates CRC on every consumed batch (`RecordBatch.isValid()` / `ensureValid()`); Dekaf currently accepts silently corrupted batches (bit flips in transit, broker storage faults, bugs like #1716 producing desynced reads) and deserializes garbage into user records.

For a client with production-stability goals this is a correctness gap: corruption should surface as a typed, per-partition `ConsumeException`, never as corrupt user data.

## Considerations

- CRC-32C over the post-header region is hardware-accelerated (`System.IO.Hashing.Crc32` / SSE4.2) — measure, but expected cost is small relative to decompression/deserialization; the zero-allocation constraint is preservable (span-based, no copies).
- Should be on by default with an opt-out (`CheckCrcs` option, mirroring `check.crcs` in the Java client/librdkafka) for users who want the last few % of throughput.
- Verification failure should follow the (fixed) error path from #1717/#1719: loud, typed, position-safe — not a silent empty partition.

## Related

- #1716 (a CRC check would have converted the desync into an immediate loud failure at the first poisoned batch)

thomhurst:
## Summary
- verify Kafka record-batch CRC-32C before decompression/deserialization
- enable checks by default and expose public WithCheckCrcs(false) opt-out
- carry the per-request flag through pooled pending requests so shared connections remain consumer-specific
- keep validation zero-allocation with stack header encoding and hardware-accelerated CRC32C
- cover valid, corrupt, opt-out, pooled reset, defaults, DI, and real-broker paths

## Verification
- Dekaf net10.0 + netstandard2.0 build: 0 warnings
- integration project build: 0 warnings
- focused unit tests: 159 passed
- CheckCrcs_DefaultAndOptOut_ConsumeValidBatch: passed against Testcontainers Kafka

## Dependency
- Typed partition-local surfacing of a CRC mismatch depends on #1730; this PR supplies the corruption signal consumed by that path.

Closes #1727

### DIFF
--- a/src/Dekaf/Builders.cs
+++ b/src/Dekaf/Builders.cs
@@ -1334,7 +1334,7 @@ public sealed class ConsumerBuilder<TKey, TValue>
     private int? _heartbeatIntervalMs;
     private int _rebalanceTimeoutMs = 60000;
     private int _requestTimeoutMs = 30000;
-    private bool _checkCrcs;
+    private bool _checkCrcs = true;
     private bool _useTls;
     private TlsConfig? _tlsConfig;
     private List<IConsumerInterceptor<TKey, TValue>>? _interceptors;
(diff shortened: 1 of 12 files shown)


thomhurst:
Dependency #1730 is merged. Rebased onto main and added explicit CRC-to-partition-error regression in f7940559: a CRC-corrupt batch under \CheckCrcs\ now proves \FetchResponsePartition.RecordParseError\ carries the inner \InvalidDataException\ and no records. net10 build and 112 focused tests pass.

PR #1744, opened 2026-07-11
Recover inconsistent fetch-clear markers

### LINKED ISSUE #1725: Hardening: marker-gate asymmetry in revoked-partitions pending-fetch-clear readers — an inconsistent entry would reproduce the silent consumer-stall shape
## Summary

Hardening: the revoked-partitions-pending-fetch-clear state is consulted asymmetrically across its three readers, so the machinery's safety depends on an invariant ("dict non-empty ⇒ marker == 1") that is maintained only by lock discipline, with no enforcement:

- `ShouldDropStaleFetchPartition` (`src/Dekaf/Consumer/KafkaConsumer.cs:3770`) checks `_coordinatorRevokedPartitionsPendingFetchClear.ContainsKey(partition)` **without** the marker gate.
- `HasPendingFetchClear` (`:3755-3757`) requires `MarkerPresent != 0` **and** dict membership.
- `ExcludePartitionsPendingFetchClear` (`:2203-2204`) skips filtering entirely when the marker is 0.

If an entry ever exists with `Pending == 0` and `Marker == 0` (e.g. a future edit to `TryQueueDivergingEpochResetLocked`/`CompleteDivergingEpochResets` sequencing, or an exception between staging and completion), the system enters exactly the silent-hang shape of #1716: dispatch keeps fetching the partition (not excluded), every response is dropped (`ShouldDropStaleFetchPartition` true), positions freeze, and the foreground clear never runs (`Pending == 0`). No current code path produces that state — this is prevention.

## Fix direction

- Make the three readers consistent (all gate on the marker, or none), or
- Assert/enforce the invariant where entries are added/removed (debug assert + defensive self-heal: if dict non-empty with marker 0 for N cycles, log + force-set the flags), and
- Add a unit test that exercises a synthetic stuck entry and asserts the consumer surfaces an error rather than hanging.

Found during the investigation of stress run 29131579448 (the audit that root-caused #1716); this specific state was verified *not* reachable today.

thomhurst:
## Summary
- centralize pending fetch-clear marker reads
- recover impossible dictionary-without-marker state under the existing coordination lock
- arm the foreground drain and emit a warning instead of silently excluding/dropping forever
- preserve the valid staged diverging-epoch window where marker is set before pending

## Validation
- `dotnet build tests/Dekaf.Tests.Unit/Dekaf.Tests.Unit.csproj -c Release --no-restore`
- `ConsumerAssignmentFastPathTests`: 23/23 passed on net10.0

No Testcontainers case is added because the inconsistent state is explicitly not reachable through current public behavior; the regression injects the synthetic internal state described by the issue.

Closes #1725

### DIFF
--- a/src/Dekaf/Consumer/KafkaConsumer.cs
+++ b/src/Dekaf/Consumer/KafkaConsumer.cs
@@ -752,6 +752,7 @@ public sealed partial class KafkaConsumer<TKey, TValue> :
     private readonly object _coordinatorRevokedPartitionsPendingFetchClearLock = new();
     private readonly ConcurrentDictionary<TopicPartition, byte> _coordinatorRevokedPartitionsPendingFetchClear = new();
     private readonly ConcurrentDictionary<TopicPartition, (long EndOffset, int Epoch)> _pendingDivergingEpochResets = new();
+    private int _stagedDivergingEpochResetBatches;
     private int _coordinatorRevokedPartitionsPendingFetchClearMarkerPresent;
     private int _coordinatorRevokedPartitionsPendingFetchClearPending;
 
@@ -2703,10 +2704,14 @@ public sealed partial class KafkaConsumer<TKey, TValue> :
                         if (partitionResponse.DivergingEpoch is not null)
                         {
                             _stuckFetchPositionTracker.Reset(tp);
-                            queuedDivergingEpochReset |= ResetToDivergingEpoch(
+                            if (ResetToDivergingEpoch(
                                 topic,
                                 partitionResponse,
-                                fetchBufferEpoch);
+                                fetchBufferEpoch,
+                                startsBatch: !queuedDivergingEpochReset))
+                            {
+                                queuedDivergingEpochReset = true;
+                            }
                             continue;
                         }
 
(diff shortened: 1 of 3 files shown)


thomhurst:
Addressed the unresolved stale-prefetch P1 in aef1b062. ShouldDropStaleFetchPartition now checks the pending-clear dictionary directly, so an inconsistent missing marker cannot let a completed prefetch advance positions before recovery. Marker-gated per-record predicates stay cheap. Updated the synthetic missing-marker regression to require the stale partition be dropped; net10 build and ConsumerAssignmentFastPathTests (23/23) pass.

thomhurst:
Addressed both current P2 threads in 8b6c4f17. Missing-marker recovery now runs only once per foreground poll (record/batch hot paths remain a single volatile pending read). Recovery restores visibility for staged diverging-epoch resets without publishing them drainable until CompleteDivergingEpochResets. Updated regression clears the staged marker and proves pending stays 0 until completion; build clean and 23/23 focused tests pass.

thomhurst:
Addressed marker-present/pending-missing P2 in 444b87d0. Recovery now independently restores each missing flag, while preserving the legitimate marker=1/pending=0 staged-divergence window. Added a non-staged marker-only regression that proves repair+drain; build clean and focused suite passes 24/24.

thomhurst:
Addressed recovery/drain race in 6cf14621. The foreground recovery wrapper now holds the existing reentrant coordination lock across marker repair and buffer drain, so a response cannot stage a new diverging reset between the scan and drain. Focused suite remains 24/24; build clean.

thomhurst:
Fixed CI regression in 5dc11d7e. ConsumerLeaderDiscoveryTests reflection helper now passes ResetToDivergingEpoch's startsBatch argument; the multi-partition same-response test marks only its first reset as batch-start. ConsumerLeaderDiscoveryTests pass 23/23 and ConsumerAssignmentFastPathTests pass 28/28; full unit project builds cleanly.

thomhurst:
Addressed all four blocking review findings in 70ffe56e. Background prefetch loop and partition exclusion now self-heal missing flags; foreground poll keeps a lock-free no-work fast path; recovery/drain use explicit locked helpers without redundant reentrant locks; ResetToDivergingEpoch reuses the staging wrapper instead of dead duplicate code. Updated missing-marker regression for background exclusion. Build clean; focused suites pass 51/51.

thomhurst:
Post-merge blocking review findings were shipped via #1748 (merged as aff3278a). Follow-up review was CLEAR with all checks green.

PR #1781, opened 2026-07-11
perf: share warm pipe pools across clients

### LINKED ISSUE #1765: perf(networking): per-consumer PipeMemoryPool starts cold — ~200KB of 64KB pipe segments allocated per client session
## Problem

Each `PipeMemoryPool` wraps a dedicated `ArrayPool<byte>.Create(4MB, 32)` (PipeMemoryPool.cs:72-75, deliberately not Shared to bound working set), and ConnectionPool creates a fresh one per instance (ConnectionPool.cs:114-116) — one per consumer/producer/admin client (KafkaConsumer.cs:932). So every client session starts with an empty pool: each concurrently-held 64KB read-pump segment (readBufferSize 65536, KafkaConnection.cs:497 / SocketPipe.cs:54) is a fresh `new byte[64K]`, then dies with the client. Trace-proven: ~200KB per ConsumeAll benchmark session, plus `ArrayPool.Create` bucket metadata per pool.

This is the largest component of Dekaf's ~528KB fixed per-session managed allocation (see #1753) and a real cost for apps that create short-lived clients.

## Fix

Share `PipeMemoryPool` process-wide, tiered by size config — exactly the `ResponseBufferPool.Create` pattern (KafkaConnection.cs:3459-3488: static default + s_sharedPools keyed by config, bounded maxArraysPerBucket). Bounded bucket depth preserves the working-set cap that motivated per-pool isolation. Consider also creating the pool lazily on first connection.

Part of the 2026-07-11 Confluent-gap audit.

thomhurst:
## Summary
- share `PipeMemoryPool` process-wide by bounded size configuration
- reuse warm 64KB pipe segments across short-lived client sessions
- keep distinct bucket capacities isolated and stop client disposal from killing global pools
- include pending pooled-list test cleanup needed for stable full-suite CI

## Test plan
- [x] Release build (`netstandard2.0`, `net8.0`, `net10.0` graph)
- [x] `ConnectionPoolTests` (38/38)
- [x] `PipeMemoryPoolTests` (12/12)
- [x] full net10 unit suite (5,090/5,090)

Closes #1765

### DIFF
--- a/src/Dekaf/Networking/ConnectionPool.cs
+++ b/src/Dekaf/Networking/ConnectionPool.cs
@@ -26,9 +26,9 @@ public sealed partial class ConnectionPool : IConnectionPool
     private Task? _idleReaperTask;
 
     /// <summary>
-    /// Shared memory pool for all connections. Bounds total retained memory to one set of
-    /// array buckets regardless of connection count, preventing multi-GB WorkingSet growth
-    /// in multi-broker scenarios with adaptive scaling. See <see cref="PipeMemoryPool"/>.
+    /// Process-shared memory pool for all clients with this size configuration. Bounds total
+    /// retained memory to one set of array buckets, preventing per-client cold starts and
+    /// multi-GB WorkingSet growth in multi-broker scenarios. See <see cref="PipeMemoryPool"/>.
     /// Bucket capacity is either provided by broker-aware producer sizing or falls back to
     /// scaling by <c>_connectionsPerBroker</c> for direct ConnectionPool use.
     /// </summary>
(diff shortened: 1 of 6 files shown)


thomhurst:
Addressed blocking cache-lifetime review in 7363b675: the process-wide shared pool cache is now capped at 16 configurations with FIFO eviction. Eviction drops only the cache reference, so active connections remain safe and retired pools become GC-eligible after release. Added a cache-bound regression test and isolated the shared-static fixture. Release build clean; focused networking/protocol tests 201/201; full net8 suite 5,020/5,020.

Issue #670, opened 2026-03-28
bug: Receive timeout after 30000ms on broker -1 in MultiPartition_CoalescedSend_OrderingPreserved


thomhurst:
## Summary

`MultiPartition_CoalescedSend_OrderingPreserved` in Producer integration tests consistently fails with receive timeouts on the bootstrap connection (broker -1) and subsequent broker connections. This occurs on multiple PR branches **and on main**, indicating a pre-existing bug.

## Root Cause

**Missing `IsCanceled` check in `ReceiveLoopAsync` after DuplexPipe migration (PR #458).**

PR #458 switched from `PipeReader.Create(Stream)` to `Pipe.Reader` (via SocketPipe/DuplexPipe). This changed the cancellation behavior:

- **Before PR #458:** `PipeReader.Create(Stream).ReadAsync(token)` throws `OperationCanceledException` when the token fires
- **After PR #458:** `Pipe.Reader.ReadAsync(token)` returns `ReadResult` with `IsCanceled = true` instead of throwing

The `ReceiveLoopAsync` (KafkaConnection.cs:856-871) only handles the throwing case. When `IsCanceled = true` is returned:
1. The catch block is never entered
2. `_disposed` is never set to 1
3. `IsConnected` remains true
4. The connection is never marked as failed
5. The ConnectionPool keeps returning the same unresponsive connection
6. All subsequent requests on this connection also time out silently
7. The loop resets the CTS and waits another 30s, indefinitely

## Fix

Add an `IsCanceled` check after `ReadAsync` returns, before processing the buffer:

```csharp
result = await _reader!.ReadAsync(timeoutCts.Token).ConfigureAwait(false);

if (result.IsCanceled && timeoutCts.IsCancellationRequested && !cancellationToken.IsCancellationRequested)
{
    LogReceiveTimeout(_options.RequestTimeout.TotalMilliseconds, BrokerId);
    Volatile.Write(ref _disposed, 1);
    throw new KafkaException(
        $"Receive timeout after {(int)_options.RequestTimeout.TotalMilliseconds}ms - connection to broker {BrokerId} failed");
}
```

## Key Files

- `src/Dekaf/Networking/KafkaConnection.cs` — lines 856-871 (missing `IsCanceled` check)
- `src/Dekaf/Networking/SocketPipe.cs` — line 34 (returns `Pipe.Reader`)
- `src/Dekaf/Networking/DuplexPipe.cs` — line 33 (returns `Pipe.Reader`)

[...]

PR #1783, opened 2026-07-11
perf(networking): read response frames directly into pooled arrays (one-copy receive path)

### LINKED ISSUE #1757: perf(networking): TryReadResponse copies every response payload a second time (pipe -> pooled array) — librdkafka parses in place
## Problem

The receive path copies every response frame from the pipe's buffers into a rented response array before parsing (KafkaConnection.cs ~1836 multi-segment path, ~1871 single-segment fast path). The socket pump already did the kernel->user copy into the Pipe (SocketPipe.cs:47-79); this second full-payload memcpy has no librdkafka equivalent — it reads the frame body into its response buffer once and parses in place.

At consumer stress throughput (~1.05GB/s on a 2-core VM) this is an estimated **0.10-0.18 CPU-μs/msg — the single largest Dekaf-only per-byte cost**, the top item in the 1.25-vs-0.91 CPU/msg gap vs Confluent (stress run 29144532789). The producer path pays it too on every response.

## Fix directions

- Parse fetch responses directly from the pipe `ReadOnlySequence<byte>`: with a 64KB+ pipe buffer most frames are single-segment, so the existing zero-copy parse can run against the pipe segment and only fall back to a copy for multi-segment frames. Requires holding the pipe read (not advancing) until batch memory is sliced/retained — fetch responses already retain pooled memory for zero-copy record slices, so the lifetime machinery exists.
- Alternative: read the frame body straight into the rented response array (skip the Pipe for the body once the 4-byte length is known), keeping exactly one copy like librdkafka.

Needs a BenchmarkDotNet + stress before/after; protocol-layer change so extra scrutiny per CLAUDE.md.

Part of the 2026-07-11 Confluent-gap audit; ranked #1 of the remaining consumer CPU costs.

thomhurst:
Closes #1757

## What

Replaces the pipe/pump receive path (`SocketPipe`/`DuplexPipe` → `Pipe` → `TryReadResponse` copy) with a new `ResponseFrameReader` that receives each response frame body **straight from the socket (or `SslStream` for TLS) into the pooled response array**. This removes the second full-payload memcpy on every response — per the issue, the single largest Dekaf-only per-byte cost on the consumer path (~0.10–0.18 CPU-µs/msg at stress throughput) — keeping exactly one user-space copy, like librdkafka.

## Design

- **Framing**: small frames (producer acks, heartbeats) still batch through a 64 KB receive buffer so one recv carries many frames; once a frame's 4-byte size prefix shows it extends past the buffered bytes, the remainder is received directly into the frame's pooled array. The internal buffer is fully drained before a direct fill starts and direct fills never request more than the frame's remaining bytes, so the source is never over-read.
- **Downstream unchanged**: the reader produces the same `(correlationId, PooledResponseBuffer)` the old path did — dispatch, awaiter-side parsing, and zero-copy fetch record slicing are untouched.
- **Wake mechanism**: reads are not cancellable (mirroring the old pump, which was also woken by socket close on teardown). `ResponseFrameReader.Abort()` (socket `Close(0)`) is the wake for both the receive-timeout timer (#670 semantics preserved: timer only fires with pending requests past the deadline, then the loop tears the connection down with the same `Receive timeout` error) and disposal.
- **Timeout refresh got cheaper**: byte progress now extends the deadline with a single volatile store; the timer re-arms itself for the remainder when it fires early, instead of a lock + `Timer.Change` per recv.
- **Deleted**: `SocketPipe`, `DuplexPipe`, the partial-frame assembly machinery, pipeline pause/resume threshold calculation. `PipeMemoryPool` stays (it now serves the one long-lived receive buffer per connection — see follow-ups).
- Both TFMs supported (`net10.0` + `netstandard2.0`); pooled async value-task builders are applied on modern .NET only.

## Benchmarks (BenchmarkDotNet, ShortRun, local)

New `ResponseFrameReaderBenchmarks` (steady state, construction outside the measured op):

| Scenario | Mean | Allocated |
|---|---|---|
| 10,000 × 100 B frames | 498 µs (~20 M frames/s) | 144 B/op (~0 B/frame) |
| 64 × 1 MB frames | 6.05 ms (~10.5 GB/s) | 144 B/op |

The isolated cost of the copy this PR removes is documented by the pre-existing `ReceiveLoopCopyBenchmark` (a full second memcpy of every response payload). Stress before/after should be validated on CI runners per the issue; local Windows runs are NAT-capped and not representative.

## Testing

- Unit: full suite green on both TFMs (`net10.0`: 5085/5085, `net8.0`/netstandard asset: 5014/5014). New `ResponseFrameReaderTests` cover multi-frame chunks, split prefixes/bodies, direct-fill integrity with a 16 B receive buffer, EOF (clean/mid-prefix/mid-body), invalid sizes, callback accounting, and abort. Receive-loop tests rewritten against the new reader (timeout kill, disposal vs stale timeout flag, EOF failure propagation, dispatch).
- Integration (local Docker, full 794-test suite): 791/794 on this branch. The 3 failures (`Producer_WithMutualTls*` handshake, `NonIdempotentMultiInflight`, `HighPartitionCount_CoalescedBrokerGrouping`) also fail on unmodified `main` in the same environment — a clean-main baseline run failed 17/794 including these exact tests, so they are pre-existing local-env flakiness, not from this change. Each also passes in isolation locally.

## Follow-ups (out of scope here)

- `PipeMemoryPool` + the ConnectionPool shared-pool plumbing now exist only to serve one ~64 KB receive buffer per connection; it could be retired in favour of `ArrayPool<byte>.Shared` and deleted wholesale.
- Pre-existing flaky unit tests observed under load while validating (not touched by this diff): `EnsureAssignmentForPollAsync_SlowPositionInitialization_DoesNotExpireMember` (timing), `FetchResponsePartition_ReturnAndRent_ClearsAllFields` / `FetchResponse_Read_HeaderStub_PreservesNextPartitionAlignment` (static-pool identity assertions race with parallel tests).


thomhurst:
Fixed net8 CI failure in e5d34831. Root cause was test cleanup double-returning pooled `List<RecordBatch>` instances: after manual `ReturnRecordBatchList`, the cleared list remained attached and response cleanup returned it again. Cleared ownership in both helpers. Release build clean; net8 boundary 2/2 and response snapshot 127/127 pass.

PR #1831, opened 2026-07-12
Preserve all release matrix results

### LINKED ISSUE #1829: Release gate: fail-fast:true cancels ~30 matrix legs on first failure - release sweep should report all legs and allow partial re-run
## Summary

In release gate run [29175004159](https://github.com/thomhurst/Dekaf/actions/runs/29175004159), 3 genuinely failing jobs (2 deterministic AOT, 1 transient Docker Hub 502) caused **~30 sibling matrix jobs to be cancelled** because every matrix uses `fail-fast: true` (`integration-groups.yml:28`, plus the release matrices in `ci.yml:256/271/285`).

## Why this is wrong for the release gate

- The release gate exists to *sweep* 4 broker versions x 3 frameworks before publishing. Cancelling the sweep on the first failure means one flaky leg leaves ~30 legs unknown — we learn nothing about whether 4.1.2/4.2.1 would have passed, and "Re-run failed jobs" must re-run cancelled legs too, repeating most of the matrix.
- In this run the 502 (transient, environmental) cancelled legs that would have surfaced — or ruled out — real per-version problems in the same pass. Diagnosing release readiness now requires a full re-dispatch.
- Fail-fast saves compute on PR CI, where any red fails the PR anyway. On a manually-dispatched release gate, complete information per dispatch is worth more than the saved minutes.

## Fix

- Set `fail-fast: false` for the release-path matrices: the `Release Matrix (...)` / `Release Matrix AOT (...)` jobs in `ci.yml` and the group matrix in `integration-groups.yml` (parameterize via a `fail-fast` workflow-call input defaulting to `true` so PR CI keeps current behavior, and pass `false` from the release path).
- The existing `CI Passed` aggregation job already ANDs all legs, so overall gating semantics are unchanged — the gate still fails; it just fails with a complete picture and supports re-running only the genuinely failed legs.

## Acceptance

A release dispatch with one failing leg reports pass/fail for every other leg; "Re-run failed jobs" re-executes only the failed ones; `publish_nuget` still blocked unless everything is green.

thomhurst:
## Summary
- keep PR/push integration matrices fail-fast
- disable fail-fast for both release caller matrices
- disable fail-fast for each release integration-group matrix

## Test plan
- [x] Parse both changed workflow files as YAML
- [x] Assert release caller and inner matrices use `fail-fast: false`
- [x] Assert PR/push caller retains `fail-fast: true`
- [x] `git diff --check`

Closes #1829

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -268,7 +268,7 @@ jobs:
     needs: [changes, build]
     if: github.event_name == 'workflow_dispatch' && inputs.publish_nuget == true
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
         framework: [net8.0, net10.0]
         kafka: ['4.0.2', '4.1.2', '4.2.1', '4.3.1']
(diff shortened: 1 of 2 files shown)


thomhurst:
Merge-Pr.ps1 gate passed, but merge aborted because `gh pr merge --delete-branch` attempted to delete the locally checked-out branch before the wrapper removed its worktree. No merge occurred; advancing per fail-closed loop rules.

Issue #1864, opened 2026-07-12
net8 unit suite: repeated timeout/cancellation failures under normal parallel run


thomhurst:
Found while validating #1858 on Windows x64, Release, net8.0.

## Evidence

Full net8 unit runner failed twice in the same clean worktree:

1. Concurrent with net10: 5,110 passed / 9 failed.
2. Net8 alone after a clean rebuild: 5,114 passed / 5 failed.

Repeated across both runs:
- DisposeAsync_OwnedDrainPastTimeout_WaitsForLease — task canceled after ~18–20s
- WriteTimeout_AbandonedWriteStuckPastGracePeriod_AbortsConnectionAndReleasesWriteLock — timeout/cancellation after ~10–14s

Additional net8-alone failures:
- DisposeAsync_SharedDrainTimeout_ContinuesWaitingForLease — task canceled (~18s)
- TryAppendFromSpansWithCompletion_MultipleAwaitedRecords_CountsPendingBatchOnce — FlushAsync canceled (~17s)
- SendLoop_ParallelSendFailures_RefreshMetadataOncePerTopic — 30s test timeout during BrokerSender.DisposeAsync (~45s wall)

The same commit's net10 full suite passed 5,200/5,200. Focused #1858 suites passed:
- BrokerSenderMuteOrderingTests: 17/17
- AdaptiveScaleDownTests: 28/28
- new scale-up regression: net10 + net8

## Impact

Net8 full-suite result is not stable under normal local parallel TUnit execution. Failures span connection retirement, write timeout, accumulator flush, and sender disposal; timing/thread-pool starvation or shared parallel test pressure is likely, but repeated failures remain real per project policy.

## Acceptance

- Reproduce the net8 full-suite failure signature under captured runner parallelism/thread-pool diagnostics.
- Remove arbitrary scheduling dependency; use deterministic synchronization and/or correct test isolation.
- Full net8 unit suite passes repeatedly without increasing individual timeouts.

PR #1865, opened 2026-07-12
fix(producer): preserve ordering across scaling

### LINKED ISSUE #1858: P0 correctness: non-idempotent producer reorders 982 same-partition messages across adaptive connection scale-up (round-trip stress fail)
Stress run: https://github.com/thomhurst/Dekaf/actions/runs/29190595109 — job "Producer → Consumer Round-Trip (Paired)" **failed** on this.

## Symptom

Dekaf round-trip validation failed with **982 out-of-order messages** (out of 250,000). Everything else was clean — this is pure reordering, not loss:

```json
"roundTripValidation": {
  "expectedMessages": 250000, "consumedMessages": 250000,
  "missingMessages": 0, "duplicateMessages": 0, "corruptMessages": 0,
  "outOfOrderMessages": 982, "mispartitionedMessages": 0,
  "unexpectedMessages": 0, "timedOut": false, "isSuccess": false
}
```

The validator (`RoundTripMessageCodec.cs:326-335`) tracks per-partition producer sequence and broker offset; `duplicates=0` rules out consumer-side replay, so messages with lower producer sequence were **written to the log at higher offsets** — the producer sent same-partition batches out of order. This is a non-idempotent producer (`idempotent: false`), where Kafka's per-partition ordering contract still applies.

## Smoking gun: mid-run connection scale-up

Delivery diagnostics captured exactly one adaptive scaling event, ~0.6s into a ~1.1s bounded run:

```json
"connectionScaleEvents": [{
  "occurredAtUtc": "2026-07-12T11:19:28.5684761+00:00",
  "brokerId": 1, "oldConnectionCount": 1, "newConnectionCount": 2,
  "direction": "up", "bufferUtilization": 0.0126,
  "bufferPressureDelta": 33, "sendLoopPressureDelta": 100
}]
```

982 reordered messages ≈ the in-flight window at the moment of the 1→2 transition. The scale-up re-routes partitions across connections while batches for the same partition are still unacked on the old connection; the new connection's batches can reach the broker first. #1698 ("preserve non-idempotent ordering") evidently does not cover the scale-*up* handoff path.

Secondary observation: the scale-up fired at **1.26% buffer utilization** on `sendLoopPressureDelta=100` during a short bounded burst — trigger sensitivity worth revisiting while in there (related: #1824 convergence, #1837 transition stalls, #1851 ordering test timeout).

## Fix plan

1. On connection scale-up (and any partition→connection re-routing), **quiesce per partition**: do not dispatch a partition's batches on the new route until every unacked batch for that partition on the old route is acked (or failed and re-queued *ahead of* newer batches). Same invariant scale-*down* already needed (#1602/#1698) — unify into one routing-transition barrier.
2. Alternatively (cheaper): pin a partition to its current connection until its unacked count reaches zero, and only then migrate — natural drain, no new synchronization on the hot path.
3. Add a stress/integration test: non-idempotent producer, force a connection scale event mid-stream, assert per-partition sequence monotonicity by consuming back (the round-trip harness already detects this — a deterministic unit-level version is what's missing, see #1851's timeout for prior art on how not to write it).

## Acceptance

- Round-trip stress green with adaptive scaling enabled and non-idempotent producer.
- `outOfOrderMessages == 0` across scale-up and scale-down transitions.

thomhurst:
## Summary
- fence non-idempotent partitions to their old connection during adaptive scale-up until pending requests acknowledge
- key migration state by full TopicPartition so same-numbered partitions across topics cannot interfere
- require all acknowledged producer requests to drain before scale-down changes modulo routing
- add deterministic scale-up and retained-slot scale-down regressions

## Test plan
- [x] dotnet build tests/Dekaf.Tests.Unit --configuration Release
- [x] net10 BrokerSenderMuteOrderingTests + AdaptiveScaleDownTests (45/45)
- [x] net8 scale-up regression (1/1)
- [x] net10 full unit suite (5,200/5,200)
- [x] net10 ProducerOrderingTests + MultiInflightProducerTests integration (23/23)
- [x] producer-roundtrip stress, 250,000 messages, real 1→2 scale event: 0 missing, duplicates, corrupt, out-of-order, mispartitioned, unexpected
- [ ] net8 full unit suite: unrelated repeated timeout/cancellation failures tracked in #1864

Closes #1858

### DIFF
--- a/src/Dekaf/Producer/BrokerSender.cs
+++ b/src/Dekaf/Producer/BrokerSender.cs
@@ -461,7 +461,7 @@ internal sealed partial class BrokerSender : IAsyncDisposable
     // Single-threaded send loop — no locks needed.
     private readonly HashSet<TopicPartition> _partitionsNeedingSequenceReset = new();
 
-    // Partition-affined connections: each partition pins to _pinnedConnections[GetConnectionForPartition(partition)].
+    // Partition-affined connections: each partition pins to _pinnedConnections[GetConnectionForPartition(topicPartition)].
     // For single-connection mode (_connectionCount == 1), degenerates to the original pinned behavior.
     // All producers use partition affinity (partition % N) for CPU cache locality.
     // Idempotent producers additionally require affinity for per-partition sequence ordering;
@@ -505,14 +505,14 @@ internal sealed partial class BrokerSender : IAsyncDisposable
     private volatile bool _loopExited;
     private volatile bool _redeliverAfterLoopExit;
 
-    // Per-partition migration fencing for idempotent producers: during a scale-UP,
+    // Per-partition migration fencing for all acknowledged producers: during a scale-UP,
     // partitions whose connection assignment changes (partition % oldCount != partition % newCount)
     // are fenced here if they have in-flight batches on the old connection. The partition
     // continues routing to the old connection until its in-flight clears, then is removed
     // from this dictionary and routes via the new partition % _connectionCount.
     // Scale-down never fences (its drain gates guarantee no in-flight) and clears any
     // stale entries at initiation. Send-loop owned (single-threaded) — no synchronization needed.
-    private readonly Dictionary<int, int> _migratingPartitions = new();
+    private readonly Dictionary<TopicPartition, int> _migratingPartitions = new();
 
     // Scaling thresholds
     // Also the per-connection unit of pressure for step estimation (step = delta / threshold),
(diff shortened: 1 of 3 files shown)


thomhurst:
Addressed both blocking findings in 1d0a17eb:

1. CompleteMigrations again scans only _migratingPartitions, not lifetime-wide _knownPartitions. TopicPartition cannot use stackalloc because it contains a managed string, so the fix uses a reusable List<TopicPartition> removal buffer. Buffer capacity grows during scale-up; response completion remains allocation-free and O(active migrations).
2. The scale-up regression now includes other-topic partition 1 alongside test-topic partition 1. Only the in-flight topic stays on connection 0; the same-numbered partition on the other topic routes to connection 1.

Verification: build clean; net10 BrokerSenderMuteOrderingTests + AdaptiveScaleDownTests 45/45; net8 cross-topic scale-up regression 1/1.

Issue #1866, opened 2026-07-12
consumer unit: foreground poll activity can expire member under full-suite contention


thomhurst:
## Summary

While validating #1859, the full net10 unit suite intermittently failed `RecordPollAsync_ActiveForegroundPollActivity_DoesNotExpireMember`: coordinator state became `Unjoined` instead of remaining `Stable` while foreground poll activity was registered.

## Evidence

```text
Expected Stable but received Unjoined
ConsumerCoordinatorKip848Tests.cs:474
```

Full-suite result: 5,203 passed, 1 failed. The exact test passed in 302ms on an isolated clean-`main` worktree at `7db60620`, and the #1859 product change executes only in `CommitOffsetsAsync`, which this test does not call.

## Acceptance

- Reproduce under suite-level parallel/thread-pool contention or deterministic scheduling.
- Foreground poll activity prevents max-poll expiry for its full registered lifetime.
- Test remains stable in repeated full-suite runs without timing sleeps.

PR #1867, opened 2026-07-12
fix(consumer): refresh coordinator before commit

### LINKED ISSUE #1859: Consumer commit-on-close fails with 'Unknown broker ID: -1'; retries never re-resolve coordinator, offsets dropped
Seen in stress run https://github.com/thomhurst/Dekaf/actions/runs/29190595109, job "Producer → Consumer Round-Trip (Paired)", during Dekaf consumer close:

```
[11:19:29.40] Warning Dekaf.Consumer.KafkaConsumer Failed to commit offsets during close (attempt 1/3)
System.InvalidOperationException: Unknown broker ID: -1
[11:19:29.61] Warning ... (attempt 2/3)  System.InvalidOperationException: Unknown broker ID: -1
[11:19:29.81] Warning ... (attempt 3/3)  System.InvalidOperationException: Unknown broker ID: -1
```

## Analysis

- The exception comes from `ConnectionPool` (`src/Dekaf/Networking/ConnectionPool.cs:253/266/299/451`) when asked for broker ID **-1** — i.e. the commit path ran with an unresolved coordinator ID (-1 is the "no coordinator" sentinel).
- The close path (`KafkaConsumer.cs:6977-7002`) guards on `_coordinator is not null`, so the coordinator *object* existed but its cached node ID was -1/invalidated.
- The retry loop (3 attempts, 200ms apart) retries the exact same call **without re-resolving the coordinator** (no FindCoordinator refresh), so all three attempts fail identically. The retries are dead weight for this failure class.

## Impact

Final offsets are silently dropped at close (warnings only). Next consumer generation re-processes from the last successful auto-commit → duplicate consumption after every close that races a coordinator change/disconnect. In the round-trip stress this is benign noise, but for real consumers it defeats commit-on-close semantics.

## Fix plan

1. In `CommitPendingOffsetsAsync` (or the close retry loop), treat `Unknown broker ID` / coordinator-unavailable as a **coordinator refresh trigger**: run FindCoordinator, then retry the commit against the new coordinator. This mirrors the AdminClient retry-context lesson (#1341): a retry that cannot change the failing input is not a retry.
2. Never enter the commit with -1: if the coordinator ID is unresolved, resolve first or fail fast with a precise error.
3. Unit test: close a consumer whose coordinator connection was invalidated (coordinator ID reset to -1) and assert the close commit re-resolves and succeeds.

## Acceptance

- Commit-on-close succeeds after coordinator loss (with refresh), or surfaces a real, actionable error — not three identical `Unknown broker ID: -1` warnings.

thomhurst:
## Summary
- rediscover an invalid consumer-group coordinator before leasing the offset-commit connection
- snapshot and validate the resolved broker ID so concurrent invalidation never sends sentinel `-1` to the pool
- cover direct coordinator commit and `KafkaConsumer.CloseAsync` pending-offset commit paths

## Test plan
- [x] `ConsumerCoordinatorKip848Tests` (79 passed)
- [x] `ConsumerDirtyCommitTests` (13 passed)
- [x] real Kafka `AutoCommit_OnClose_CommitsFinalOffsets` integration test
- [x] full net10 suite: 5,203 passed; unrelated foreground-poll timing flake filed as #1866

Closes #1859

### DIFF
--- a/src/Dekaf/Consumer/ConsumerCoordinator.cs
+++ b/src/Dekaf/Consumer/ConsumerCoordinator.cs
@@ -776,10 +776,6 @@ public sealed partial class ConsumerCoordinator : IAsyncDisposable
         if (string.IsNullOrEmpty(_options.GroupId))
             return new Dictionary<TopicPartition, TopicPartitionOffset>();
 
-        // If coordinator hasn't been discovered yet, return empty (no committed offsets known)
-        if (_coordinatorId < 0)
-            return new Dictionary<TopicPartition, TopicPartitionOffset>();
-
         // Lock is intentionally held across retries to protect the shared _fetchTopicGroups dictionary,
         // which is reused across calls to avoid allocations. With up to 3 retries x ~600ms each,
         // the lock can be held for up to ~1.8 seconds. Concurrent callers will block for the full retry duration.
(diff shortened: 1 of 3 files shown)


thomhurst:
Addressed blocking FetchOffsetsAsync race in 8676d352. Fetch now discovers inside RetryHelper, snapshots _coordinatorId once, validates it, and throws retriable CoordinatorNotAvailable instead of passing -1 to the pool. Added FetchOffsetsAsync_UnknownCoordinator_RediscoversBeforeFetch. ConsumerCoordinatorKip848Tests + ConsumerDirtyCommitTests pass 93/93 on net10 and net8.

PR #1872, opened 2026-07-12
fix(stress): preserve per-pass allocation

### LINKED ISSUE #1860: Stress harness: gcStats.allocatedBytes null for plain Dekaf rows in paired jobs — alloc/msg invisible for default config
Stress run: https://github.com/thomhurst/Dekaf/actions/runs/29190595109 (merged results artifact `stress-test-results-20260712-120501.json`).

## Symptom

`gcStats.allocatedBytes` is **null** — and therefore `allocatedBytesPerMessage` is missing — for **every plain "Dekaf" client row in paired jobs**, while `gen0Collections` in the same `gcStats` object is populated:

| Scenario | Client | allocatedBytes | gen0Collections |
|---|---|---|---:|
| producer / 1b | Dekaf | **null** | 295 |
| producer / 1b | Dekaf (3conn) | 1,758,020,112 | 336 |
| producer / 3b | Dekaf | **null** | 206 |
| producer-idempotent / 1b | Dekaf | **null** | 442 |
| producer-idempotent / 3b | Dekaf | **null** | 459 |
| producer-acks-all / 1b & 3b | Dekaf | **null** | 415 / 245 |
| consumer / 1b | Dekaf | **null** | 1,851 |

All Confluent rows, all `Dekaf (3conn)` rows, and all solo-job Dekaf rows (consumer-batch, consumer-raw, consumer-raw-batch, roundtrip, transactional) have the value. The failing set is exactly "first Dekaf pass in a multi-client paired job", which points at the allocated-bytes capture (`GC.GetTotalAllocatedBytes` delta or the MeterListener path — cf. the MeterListener cctor ordering issue from #1633) being initialized only for the last-constructed client run, or the delta baseline being consumed by the subsequent 3conn pass.

## Impact

Allocation-per-message is one of Dekaf's headline claims and the stress suite's regression signal for it. Right now the **primary Dekaf configuration has no allocation data at all** in paired runs — an allocation regression in the default path would ship invisibly. (The trend gate also can't compare what isn't there.)

## Fix plan

1. Reproduce locally with any paired scenario; log the raw before/after `GC.GetTotalAllocatedBytes(precise)` values per client pass.
2. Fix the capture so each client pass snapshots its own baseline/final pair regardless of run order (per-pass snapshot struct, not shared mutable state).
3. Make the reporter fail loudly (or at least warn in the job summary) when `allocatedBytes` is null for a client that ran to completion, so this can't silently regress again.
4. Backfill note: history entries from affected runs have no alloc medians for these rows; the trend script should treat null as "no data", not zero (verify it already does).

thomhurst:
## Summary
- sample and accumulate managed allocations within each stress-test pass so runtime counter rebases cannot erase a later paired client
- dispose the per-pass sampler across every scenario and warn when a counter rebase is observed
- fail completed runs when managed allocation data is unavailable
- add regression coverage for normal deltas, counter rebases, repeated capture, and missing-measurement failure

## Test plan
- [x] dotnet build tests/Dekaf.Tests.Unit/Dekaf.Tests.Unit.csproj -c Release -f net10.0 --no-restore
- [x] net10 unit suite (5,206 passed)
- [x] 1-minute paired producer stress run with STRESS_CLIENT_ORDER=confluent-first (both allocation rows populated; zero errors)

Closes #1860

### DIFF
--- a/tests/Dekaf.Tests.Unit/Stress/RoundTripMessageCodecTests.cs
+++ b/tests/Dekaf.Tests.Unit/Stress/RoundTripMessageCodecTests.cs
@@ -457,13 +457,14 @@ public class RoundTripMessageCodecTests
         await Assert.That(snapshot).IsNotNull();
         await Assert.That(snapshot!.DiagnosticsEnabled).IsTrue();
 
+        using var gcStats = new GcStats();
         var result = RoundTripScenarioHelpers.CreateResult(
             "producer-roundtrip",
             "Dekaf",
             options,
             DateTime.UtcNow,
             new ThroughputTracker(),
-            new GcStats(),
+            gcStats,
             delivered: 0,
             new RoundTripValidationSnapshot
             {
--- a/tests/Dekaf.Tests.Unit/StressTests/GcStatsTests.cs
+++ b/tests/Dekaf.Tests.Unit/StressTests/GcStatsTests.cs
@@ -1,0 +1,44 @@
+using Dekaf.StressTests.Metrics;
+
+namespace Dekaf.Tests.Unit.StressTests;
+
+public sealed class GcStatsTests
+{
+    [Test]
+    public async Task Capture_ReportsAllocationDeltaForPass()
+    {
+        var readings = new Queue<long>([100, 160, 225]);
+        using var stats = new GcStats(readings.Dequeue, sampleInterval: null);
+
+        stats.SampleAllocation();
+        stats.Capture();
+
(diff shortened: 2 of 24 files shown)


thomhurst:
Resolved the remaining Codex thread as a runtime-behavior false positive. On this repository's actual net10.0 target (.NET 10.0.301), Timer.Change(TimeSpan, TimeSpan) after Dispose() returns false rather than throwing; Claude independently reproduced this and reviewed current head 0f6827ac as CLEAR.

PR #1877, opened 2026-07-12
test(producer): stabilize span append tests

### LINKED ISSUE #1870: net8: span-append pending-batch flush cancels under full-suite load
Part of #1864.

## Scope

Fix TryAppendFromSpansWithCompletion_MultipleAwaitedRecords_CountsPendingBatchOnce canceling in RecordAccumulator.FlushAsync under net8 full-suite contention.

## Acceptance

- Determine whether pending-batch accounting or test synchronization causes the cancellation.
- Preserve exact once-only pending-batch accounting.
- Test passes repeatedly in the net8 full suite without increasing timeouts.

thomhurst:
Closes #1870

## What changed
- run intentionally blocked span-append operations on dedicated threads, independent of full-suite ThreadPool pressure
- replace the pending-count test's asynchronous polling drainer with deterministic flush/seal/drain completion
- retain exact once-only pending-awaited-batch accounting assertions without increasing timeouts

## Validation
- RecordAccumulatorTests net8.0: 80/80
- RecordAccumulatorTests net10.0: 80/80
- full net8.0 unit suite, consecutive run 1: 5122/5122
- full net8.0 unit suite, consecutive run 2: 5122/5122

### DIFF
--- a/src/Dekaf/Producer/RecordAccumulator.cs
+++ b/src/Dekaf/Producer/RecordAccumulator.cs
@@ -1016,6 +1016,8 @@ public sealed partial class RecordAccumulator : IAsyncDisposable
     private int _disposed;
     private int _closed;
 
+    internal Action? PurgeAppendWaitObservedForTest;
+
     /// <summary>
     /// True after CloseAsync has been called. Used by the sender loop to know
     /// when to exit after draining remaining batches.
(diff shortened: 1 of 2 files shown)


thomhurst:
Addressed the unresolved purge-race review on current head 997ea0c9. Added a one-shot test seam exactly where Purge observes AppendInProgress under the partition lock; the test now waits for that signal before asserting Purge is blocked and releasing the append. RecordAccumulatorTests pass 80/80 on net8/net10; full net8 passes 5122/5122.

Issue #1887, opened 2026-07-12
net10: shared-infrastructure scale-down test observes zero connections under suite load


thomhurst:
Observed while validating #1857 on Windows/net10 Release with 16-way test concurrency.

Full suite result: 5,217 passed / 1 failed.

Failure:
```
SharedInfrastructure_DoesNotShrinkConnectionGroup
Expected connectionCount == 1, found 0
AdaptiveScaleDownTests.cs:382
```

The test passed immediately in isolation (1/1), and the full net8 suite passed 5,129/5,129. The #1857 patch only changes `BrokerUnackedByteBudget`, stress result metadata/reporting, and its direct tests; it does not touch shared infrastructure or scale-down code.

Likely suite-load scheduling race in connection-group setup/observation. Diagnose with deterministic synchronization; do not extend timeouts or retry.

PR #1888, opened 2026-07-12
fix(producer): base admission budget on min RTT

### LINKED ISSUE #1857: P0: delivery-latency p99 gate fails on 1-broker (2.14x/2.18x Confluent) — unacked-budget horizon uses loaded RTT, self-inflates
Stress run: https://github.com/thomhurst/Dekaf/actions/runs/29190595109

## Symptom

The delivery-latency gate (#1811, Dekaf/Confluent ratio <= 2.0) **fails on both 1-broker producer scenarios**:

| Scenario | Metric | Ratio | Status |
|---|---|---:|---|
| producer / 1b | p99 vs Confluent | **2.14** | Threshold breach (fail) |
| producer-acks-all / 1b | p99 vs Confluent | **2.18** | Threshold breach (fail) |

Absolute numbers show the tail is far above the 10ms `DeliveryLatencyTargetMs` #1843 was supposed to enforce:

| Scenario (1b) | Dekaf p50 | Dekaf p95 | Dekaf p99 | Confluent p95 | Confluent p99 |
|---|---:|---:|---:|---:|---:|
| producer | 4.7ms | **41.4ms** | 53.4ms | 12.4ms | 25.0ms |
| producer-acks-all | 5.1ms | **50.3ms** | 57.6ms | 14.1ms | 26.4ms |
| producer-idempotent | 5.0ms | 25.0ms | 49.1ms | 26.9ms | 43.9ms |

p50 is now fine everywhere (0.78–1.03x) and 3-broker p50 went from ~20x to 0.81x — #1843's core idea works. The problem is the tail on fast (low-base-RTT, high-rate) 1-broker links.

## Root cause: the RTT term is measured *loaded*, so it self-inflates

`BrokerSender.cs:2313` computes `ackRttTicks = now - pending.RequestStartTime` — the round-trip of a request through a pipeline that already holds up to 32MB of standing bytes. That RTT **includes the queueing delay the budget is supposed to eliminate**. The budget horizon is `max(target, 1.5 × ewmaRtt)` (`BrokerUnackedByteBudget.cs:187-189`), so:

more standing bytes → higher measured RTT → higher horizon → budget allows more standing bytes.

The loop converges to a standing queue of ~1.5× whatever the loaded RTT happens to be, not to the 10ms target. At 1b rates (~1.3 GB/s) with a loaded RTT of ~27ms, horizon ≈ 40ms → ~50MB budget → cap (32MB) binds and p95 sits at 41–50ms. The `RttSafetyMultiplier` comment (`BrokerUnackedByteBudget.cs:39-43`) guards against the *underfill* spiral but the estimator enables the *overfill* one. This is the textbook bufferbloat estimator error that BBR/CoDel address by tracking **min-RTT**, never smoothed loaded RTT.

## Fix plan

1. Replace the loaded-RTT EWMA with a **windowed min-RTT** (e.g. min over the last ~10s, refreshed by occasional brief budget dips, à la BBR ProbeRTT). Min-RTT approximates base propagation+broker-service time without queueing.
2. Horizon becomes `max(target, 1.5 × minRtt)` — on a LAN where minRtt ≈ 1–3ms the 10ms target actually binds, and the standing queue drops to ~rate×10ms (~13MB at 1.3GB/s) instead of pinning at the 32MB cap.
3. Keep the 1.5× BDP safety floor (it is correct once RTT is the *base* RTT).
4. Add a stress assertion on Dekaf absolute p95 vs `DeliveryLatencyTargetMs` (e.g. p95 <= 3× target under open-loop saturation) so the target is enforced, not just the Confluent-relative ratio.

Must be fixed together with the throughput-side estimator defects (windowed-max rate, probing) — same class, same file; see #1856.

## Acceptance

- Delivery latency p99 ratio <= 2.0 vs Confluent on all paired producer scenarios (gate green).
- Dekaf 1b producer p95 <= ~3× DeliveryLatencyTargetMs.
- No throughput give-back beyond #1856's acceptance criteria.

thomhurst:
Closes #1857

## Summary
- replace loaded-RTT EWMA horizon with a 10-second minimum-RTT estimate
- drain to the configured target for an initial and periodic ProbeRTT refresh (minimum 50ms, at least one observed RTT)
- suppress upward capacity probing while the base-RTT drain is active
- serialize the producer latency target and fail stress trends when p95 exceeds 3x target

## Why
Loaded acknowledgement RTT contains standing queue delay. Feeding its EWMA back into `max(target, 1.5 * RTT)` lets the allowed queue inflate itself. Minimum RTT approximates propagation plus broker service, while brief target-only drains refresh that estimate after path changes.

## Validation
- estimator + sender suites: 56/56 net10, 56/56 net8
- full net8 unit suite: 5,129/5,129
- full net10 unit suite: 5,217/5,218; unrelated `SharedInfrastructure_DoesNotShrinkConnectionGroup` failed under suite load and passed isolated, tracked in #1887
- Python stress/report tests: 52/52
- paired local one-broker stress completed with zero errors or delivery loss; throughput stayed within 1.6% (222.6k before initial ProbeRTT refinement, 219.1k after). This workstation was broker/host-limited and did not reproduce CI's 1.3GB/s budget-binding regime, so the high-rate p95/p99 acceptance remains for stress CI to verify.

### DIFF
--- a/.github/scripts/stress_report.py
+++ b/.github/scripts/stress_report.py
@@ -192,9 +192,34 @@ def paired_latency_thresholds(results):
         client = str(result.get('client', ''))
         if client.casefold() != 'dekaf':
             continue
-        baseline = baselines.get(_latency_identity(result))
         latency = result.get('latency')
-        if baseline is None or not isinstance(latency, dict):
+        if not isinstance(latency, dict):
+            continue
+
+        target_ms = result.get('deliveryLatencyTargetMs')
+        p95_us = latency.get('p95Us')
+        if _positive_finite_number(target_ms) and _positive_finite_number(p95_us):
+            ratio = p95_us / (target_ms * 1_000)
+            breached = ratio > 3.0
+            evaluations.append({
+                'scenario': _latency_scenario_label(result),
+                'metric': 'latencyP95TargetRatio',
+                'metricLabel': 'Delivery latency p95 / target',
+                'current': ratio,
+                'baselineCount': 0,
+                'median': None,
+                'mad': None,
+                'lower': None,
+                'upper': 3.0,
+                'status': 'regression' if breached else 'stable',
+                'repeatedRegression': False,
+                'thresholdBreach': breached,
+                'thresholdDirection': 'maximum',
+                'latencyThreshold': True,
+            })
+
+        baseline = baselines.get(_latency_identity(result))
+        if baseline is None:
             continue
 
         for field, metric, label in (
--- a/.github/scripts/stress_report.py
+++ b/.github/scripts/stress_report.py
@@ -190,7 +190,7 @@ def paired_latency_thresholds(results):
     evaluations = []
     for result in results:
         client = str(result.get('client', ''))
-        if client.casefold() != 'dekaf':
+        if not client.casefold().startswith('dekaf'):
             continue
         latency = result.get('latency')
         if not isinstance(latency, dict):
@@ -217,6 +217,11 @@ def paired_latency_thresholds(results):
                 'thresholdDirection': 'maximum',
                 'latencyThreshold': True,
             })
+
+        # Multi-connection Dekaf variants use their own throughput profile and have no
+        # like-for-like Confluent pair, but the absolute configured target still applies.
+        if client.casefold() != 'dekaf':
+            continue
 
         baseline = baselines.get(_latency_identity(result))
         if baseline is None:
--- a/.github/scripts/test_stress_report.py
+++ b/.github/scripts/test_stress_report.py
@@ -151,16 +151,20 @@ class StressReportTests(unittest.TestCase):
         self.assertEqual(3.0, evaluation["upper"])
         self.assertTrue(evaluation["thresholdBreach"])
 
-    def test_paired_latency_thresholds_ignore_multi_connection_dekaf_results(self):
+    def test_paired_latency_thresholds_enforce_target_without_pairing_multi_connection_results(self):
         confluent = stress_result("Confluent", effective_rate=1000)
         confluent["latency"] = {"p50Us": 10_000, "p99Us": 50_000}
         multi_connection = stress_result("Dekaf (3conn)", effective_rate=2000)
-        multi_connection["latency"] = {"p50Us": 50_000, "p99Us": 500_000}
+        multi_connection.update({
+            "latency": {"p50Us": 50_000, "p95Us": 31_000, "p99Us": 500_000},
+            "deliveryLatencyTargetMs": 10,
+        })
 
-        self.assertEqual(
-            [],
-            paired_latency_thresholds([confluent, multi_connection]),
-        )
+        evaluations = paired_latency_thresholds([confluent, multi_connection])
+
+        self.assertEqual(1, len(evaluations))
+        self.assertEqual("latencyP95TargetRatio", evaluations[0]["metric"])
+        self.assertTrue(evaluations[0]["thresholdBreach"])
 
     def test_paired_latency_thresholds_require_matching_roundtrip_bound(self):
         confluent = stress_result("Confluent", effective_rate=1000)
--- a/.github/scripts/test_stress_report.py
+++ b/.github/scripts/test_stress_report.py
@@ -135,6 +135,22 @@ class StressReportTests(unittest.TestCase):
 
         self.assertEqual([], paired_latency_thresholds([unpaired]))
 
+    def test_paired_latency_thresholds_enforce_absolute_p95_target(self):
+        dekaf = stress_result("Dekaf", effective_rate=1400)
+        dekaf.update({
+            "latency": {"p50Us": 5_000, "p95Us": 31_000, "p99Us": 40_000},
+            "deliveryLatencyTargetMs": 10,
+        })
+
+        evaluations = paired_latency_thresholds([dekaf])
+
+        self.assertEqual(1, len(evaluations))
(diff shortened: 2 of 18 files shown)


thomhurst:
Addressed all current-head review findings in bd5acafa:
- SetCap now receives the sender timestamp, expires overdue min-RTT and capacity probes before republishing, and retains prior min RTT across idle probe expiry.
- expired min-RTT refresh windows use the RTT that triggered the refresh, not the stale minimum.
- absolute p95/target gating applies to Dekaf-prefixed variants while paired Confluent ratios remain exact Dekaf only.
- stale EWMA ownership comment corrected.

Added deterministic tests for SetCap expiry, current-RTT probe duration, and Dekaf (3conn) target gating. Budget+sender suites pass 58/58 on net10/net8; Python stress tests 41/41; full units net10 5220/5220 and net8 5131/5131.

thomhurst:
Addressed the late-ACK probe ordering finding in c0dea6b1. Active min-RTT probes now complete expiration before considering the current sample; samples strictly inside the window contribute, and a late lower RTT can still improve through normal min logic while a late loaded RTT cannot inflate the base. Added MinimumRtt_LateAckAfterEmptyProbeRetainsPriorMinimum. Budget+sender suites pass 59/59 on net10/net8; full units net10 5221/5221 and net8 5132/5132.

thomhurst:
Addressed the current-head idle-probe admission finding in 573b2bd2. Appenders remain read-only: the sender precomputes and volatile-publishes post-probe budgets, while the admission slow path checks the probe deadline and restores the retained min-RTT floor after expiry. Stopwatch sampling occurs only when the current gate would block. Added Admission_ExpiredIdleMinRttProbeUsesSafetyFloorWithoutAck. Focused budget/admission/sender suites pass 74/74 on net10/net8; full units pass 5222/5222 on net10 and 5133/5133 on net8.

thomhurst:
@claude review

thomhurst:
Addressed the queued-append expiry finding in b47b8e82. Budget-blocked PendingAppend operations now reuse their pooled max.block timer for a probe-deadline admission recheck, so DrainPendingAppends runs without an ACK/release. PendingAppend continuations now run asynchronously; dump analysis found inline completion could re-enter DisposeAsync while the drain guard was held. Added both timer-mechanism and end-to-end no-ACK probe-expiry regressions. net10 affected suites 119/119 and full units 5223/5223; net8 affected suites 120/120. Full net8 remains blocked on this branch only by the separately fixed MPSC convoy in #1894.

thomhurst:
Addressed the current-head queued-append wake finding in b47b8e82, then simplified the regression coverage and preserved the admission hot-path short-circuits in 97cbe812/HEAD. Pending appends reuse their pooled max.block timer to re-drain at the min-RTT probe deadline; no per-append timer allocation is added. The end-to-end regression test verifies a budget-gated append completes after probe expiry without an ack. Full units passed 5224/5224 net10 and 5135/5135 net8 before redundant-test removal; focused current-head admission/accumulator/budget suite passes 119/119 net10.

thomhurst:
@claude review

thomhurst:
Addressed the sibling-cancellation finding in the latest commit. Drain now caches one admission recheck delay per blocked partition per scan (so it performs only one budget lookup/Stopwatch read), but arms every live sibling's existing pooled timer. Cancelling any one waiter therefore cannot remove the partition's only probe-expiry wake.\n\nAdded AsyncAppend_SiblingUnblocks_WhenFirstProbeWaiterIsCancelled with a deterministic 1-second probe: two same-partition appends queue, the first is cancelled, and the sibling still completes at probe expiry without an ack.\n\nValidation: focused admission/budget/accumulator suites 120/120 on net10 and net8; full units 5224/5224 net10 and 5135/5135 net8.

thomhurst:
@claude review

thomhurst:
Addressed both current-head review findings in b2441cc6 and updated onto current main. Post-probe admission budgets now use the minimum RTT observed during the active probe, and an admission timer whose drain attempt loses the CAS guard keeps a 1ms retry armed instead of falling back to max.block.ms. Added deterministic regressions for both paths. Validation on combined head: build 0 warnings/errors; affected producer suites 122/122 on net10 and 122/122 on net8; full units 5227/5227 net10 and 5138/5138 net8.

thomhurst:
@claude review

thomhurst:
Addressed the three round-9 blockers in f7a27696. Min-RTT probes now clamp target-only drain to 250ms so an outlier cannot disable the safety floor for seconds. PendingAppend scheduling now shares one absolute deadline per blocked partition per scan and CAS-retains any earlier deadline, eliminating repeated clock reads/Timer.Change calls while keeping every sibling independently armed. Added real-broker Producer integration coverage: a record is queued behind an active admission probe, wakes at expiry, reaches Kafka, and receives metadata. Validation: unit/integration builds 0 warnings/errors; affected units 124/124 on net10 and net8; full units 5229/5229 net10, 5140/5140 net8; exact real-broker integration 1/1.

thomhurst:
@claude review

thomhurst:
Fixed NativeAOT IL2075 in the latest commit. The integration helper now reflects from the exact typeof(BrokerUnackedByteBudget) instead of instance.GetType(), giving trimming analysis the concrete non-public-field metadata requirement. Managed integration build passes with 0 warnings/errors; CI AOT publish will verify the Linux toolchain. The failure log was deterministic analyzer output at ProducerTests.cs:77, not runner infrastructure.

PR #1889, opened 2026-07-12
test(producer): isolate shared scale check

### LINKED ISSUE #1887: net10: shared-infrastructure scale-down test observes zero connections under suite load

thomhurst:
Closes #1887

Blocks/finishes the remaining stability work in #1864.

## Summary
- disable live adaptive scaling in the direct `MaybeScaleConnections` state-machine test
- use existing reflection helpers for the test's controlled state
- leave production behavior unchanged

## Root cause
The test set `_connectionCount = 2` and invoked the private scale method twice while the live sender loop could invoke the same method concurrently. One writer could shrink 2→1 and the other 1→0 before the assertion. Production uses only the serialized send-loop writer and retains its minimum-count guard.

## Validation
- focused net8: 25/25 consecutive process runs
- full net8: 5,127/5,127 three consecutive times (15,381 total)
- full net10: 5,216/5,216
- Release build: 0 warnings/errors

### DIFF
--- a/tests/Dekaf.Tests.Unit/Producer/AdaptiveScaleDownTests.cs
+++ b/tests/Dekaf.Tests.Unit/Producer/AdaptiveScaleDownTests.cs
@@ -360,20 +360,17 @@ public sealed class AdaptiveScaleDownTests
 
         try
         {
-            typeof(BrokerSender).GetField(
-                "_connectionCount",
-                BindingFlags.Instance | BindingFlags.NonPublic)!
-                .SetValue(sender, 2);
+            // This test drives MaybeScaleConnections directly. Keep the live send loop
+            // from becoming a second writer between the two deterministic passes.
+            SetField(sender, "_adaptiveScalingEnabled", false);
+            SetField(sender, "_connectionCount", 2);
 
             // First pass starts low-utilization tracking; zero sustained window makes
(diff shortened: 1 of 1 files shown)


thomhurst:
Addressed the remaining race in 5a8b2d04: the test now calls RequestCancellation and awaits the actual _sendLoopTask before mutating _connectionCount or directly invoking MaybeScaleConnections. A loop that already passed the scaling gate therefore cannot remain a second writer. AdaptiveScaleDownTests pass 28/28 on net10/net8; exact regression 20/20 fresh net10 processes; full net10 5216/5216.

PR #1951, opened 2026-07-13
Reduce consumer LOH churn

### LINKED ISSUE #1942: P1: Consumer LOH churn: fetch-frame pool retains 4 arrays/bucket vs ~11+ live frames -> 1211 Gen2 GCs per 15min (Confluent: 1)
## Summary

Dekaf consumers trigger **hundreds of Gen2 collections per 15-minute run** (stress run [29213011704](https://github.com/thomhurst/Dekaf/actions/runs/29213011704)) while Confluent triggers 1. The dominant cause: multi-MB fetch-frame arrays continuously falling out of the `ResponseBufferPool`, whose retention cap (4 arrays/bucket) is far below the ~11+ frames the prefetch pipeline is budgeted to keep alive. LOH allocations only die in Gen2 → constant LOH drop-then-reallocate churn. Producer delivery-latency outliers in the same run attribute 1.1s stalls to "GC pause … Gen2 +6 / pause +27.1ms", so this also feeds tail latency.

| Scenario | msg/s | alloc B/msg | Gen0 | Gen1 | Gen2 | bytes per Gen2 |
|---|---|---|---|---|---|---|
| consumer | 2.79M | 455 | 1890 | 1842 | **1211** | ~0.94 GB |
| consumer-batch | 3.00M | 400 | 1851 | 1812 | **1151** | ~0.94 GB |
| consumer-raw | 3.85M | 102 | 1438 | 1290 | 416 | ~0.85 GB |
| consumer-raw-batch | 4.11M | 24 | 1726 | 1280 | 75 | ~1.19 GB |
| Confluent consumer | 1.01M | 2400 | 440283 | 40 | **1** | — |

Bytes-per-Gen2 is ~constant (~1 GB) across a 13x allocation range — the signature of one LOH allocation budget driving Gen2. Confluent allocates 5x more per message but it's all SOH (Gen0-collected). Note also "Gen1 ≈ Gen0" is mostly counter semantics: every Gen2 includes Gen0/Gen1, so ~1211 of those counts *are* the Gen2s; the real ephemeral profile is only ~679 GCs.

## Allocation sites (verified)

1. **Fetch frame rent — primary.** `ResponseFrameReader.cs:114` `_responseBufferPool.Pool.Rent(frameSize)`; frames are ~4–48MB. Pool created with `maxArraysPerBucket: 4` (`KafkaConnection.cs:3484-3494`, deliberately reduced from 32/8). But frame lifetime is long: ownership transfers via `PooledResponseBuffer.TransferOwnership()` → `FetchResponse.PooledMemoryOwner` → `RefCountedMemoryOwner` shared by all partitions' `PendingFetchData`, returned only after **every record of the fetch is consumed** (`KafkaConsumer.cs:534`). The prefetch budget allows `(PrefetchPipelineDepth+1) × min(partitions × partitionFetchBytes, fetchMaxBytes)` = 6 × 24MB = **144MB (~11 frames) live at once** (`KafkaConsumer.cs:2712-2730`), growing toward ~576MB as `AdaptiveFetchSizer` raises partition bytes 4→16MB. 11+ live vs 4 retained ⇒ every fluctuation converts to a fresh LOH alloc + a dropped array. Adaptive resizing also migrates frames across pow2 buckets, stranding retained arrays in dead buckets.
2. **Parsed-record slab — secondary.** `KafkaConsumer.cs:191` `ArrayPool<Record>.Shared.Rent(totalRecordCount)` ≈ 295KB per partition-fetch (4MB fetch of 1KB msgs), plus full `AsSpan().Clear()`, returned each cycle.
3. **Consumer-read `RecordBatch` objects — Gen1 residual.** Deliberately not pooled (`RecordBatch.cs:1296-1300`), ~175K/s at 16KB wire batches, live ms–tens-of-ms through parse → `MpscFetchBuffer` → consume — classic mid-life objects. The codebase already documents and fixes this exact failure mode on the producer side (`RecordBatch.cs:769-772`).
4. (Not exercised in this run, real for compressed topics): decompression `DetachableBufferWriter(pool, recordsLength * 4)` (`RecordBatch.cs:1267-1274`) — 4x estimate + doubling growth → LOH overshoot.

## Ranked fixes

1. **Size the fetch-frame pool to the prefetch pipeline** (targets ~90% of allocated bytes and nearly all Gen2s): dedicated ring of frame buffers sized `(PrefetchPipelineDepth + 2) × fetchConnections`, or derive the consumer-tier `maxArraysPerBucket` from consumer config (~16) instead of flat 4. Retention is bounded by the same prefetch budget that already governs live memory, so the old unbounded-retention concern behind cap=4 doesn't apply. Also **quantize adaptive fetch sizes to pow2 bucket boundaries** so resizing doesn't strand retained arrays.
2. **Pool consumer-read `RecordBatch`** via the existing `RecordBatchPool` (`RecordBatch.cs:779-830`), resolving the documented double-dispose ambiguity with a generation/lease token as `PendingFetchData._headerGeneration` already does (`KafkaConsumer.cs:548-553`). Kills ~175K mid-life allocations/s.
3. **Retain the `Record[]` slab inside pooled `PendingFetchData`** (grow-only field) instead of per-fetch `Rent/Return + Clear`.

Validation: traced consumer stress run (perf-trace-analyzer) confirming (a) LOH alloc rate ≈1.2 GB/s with stacks in `ResponseFrameReader.ReadFrameAsync`, (b) Gen2 trigger = AllocLarge; post-fix expect Gen2 in single digits and the alloc/msg for `consumer` to drop from 455B toward the ~24B floor.

Expected side benefit: removes the "GC pause / Gen2" delivery-latency outliers on producer scenarios sharing the process in round-trip, and the consumer-raw steady-state drift (-8.4%, steady/peak 0.82 breach this run) if churn-driven.

thomhurst:
## Summary
- size consumer response-buffer pools for the prefetch pipeline working set
- quantize adaptive fetch sizes to ArrayPool power-of-two tiers
- reuse consumer RecordBatch objects and retain parsed-record slabs across fetches
- preserve generation-safe ownership and failed-parse cleanup

## Test plan
- [x] `Dekaf.Tests.Unit.exe` (5,302 passed)
- [x] `Dekaf.Tests.Integration.exe --treenode-filter '/*/*/ConsumerTests/*'` (30 passed)

Closes #1942

### DIFF
--- a/src/Dekaf/Consumer/AdaptiveFetchSizer.cs
+++ b/src/Dekaf/Consumer/AdaptiveFetchSizer.cs
@@ -192,24 +192,58 @@ internal sealed class AdaptiveFetchSizer
 
     private void Grow()
     {
-        _currentPartitionFetchBytes = Math.Min(
-            (int)Math.Min((long)(_currentPartitionFetchBytes * _growthFactor), int.MaxValue),
+        _currentPartitionFetchBytes = QuantizeGrowth(
+            _currentPartitionFetchBytes,
+            _growthFactor,
             _maxPartitionFetchBytes);
 
-        _currentFetchMaxBytes = Math.Min(
-            (int)Math.Min((long)(_currentFetchMaxBytes * _growthFactor), int.MaxValue),
+        _currentFetchMaxBytes = QuantizeGrowth(
+            _currentFetchMaxBytes,
+            _growthFactor,
             _maxFetchMaxBytes);
     }
 
     private void Shrink()
     {
-        _currentPartitionFetchBytes = Math.Max(
-            (int)(_currentPartitionFetchBytes * _shrinkFactor),
+        _currentPartitionFetchBytes = QuantizeShrink(
+            _currentPartitionFetchBytes,
+            _shrinkFactor,
             _minPartitionFetchBytes);
 
-        _currentFetchMaxBytes = Math.Max(
-            (int)(_currentFetchMaxBytes * _shrinkFactor),
+        _currentFetchMaxBytes = QuantizeShrink(
+            _currentFetchMaxBytes,
+            _shrinkFactor,
             _minFetchMaxBytes);
+    }
+
+    private static int QuantizeGrowth(int current, double factor, int maximum)
+    {
+        var candidate = Math.Min((long)(current * factor), maximum);
+        return Math.Min(RoundUpToPowerOfTwo(candidate), maximum);
+    }
+
+    private static int QuantizeShrink(int current, double factor, int minimum)
+    {
+        var candidate = Math.Max((long)(current * factor), minimum);
+        return Math.Max(RoundDownToPowerOfTwo(candidate), minimum);
+    }
+
+    private static int RoundUpToPowerOfTwo(long value)
+    {
+        var result = 1L;
+        while (result < value && result <= int.MaxValue)
+            result <<= 1;
+
+        return result > int.MaxValue ? int.MaxValue : (int)result;
+    }
+
+    private static int RoundDownToPowerOfTwo(long value)
+    {
+        var result = 1L;
+        while ((result << 1) <= value && result <= int.MaxValue / 2L)
+            result <<= 1;
+
+        return (int)result;
     }
 }
 
--- a/src/Dekaf/Consumer/KafkaConsumer.cs
+++ b/src/Dekaf/Consumer/KafkaConsumer.cs
@@ -1028,7 +1028,7 @@ public sealed partial class KafkaConsumer<TKey, TValue> :
                 PoolSizing.ForConsumerResponseBuffers(
                     options.BootstrapServers.Count,
                     options.PrefetchPipelineDepth,
-                    options.ConnectionsPerBroker)),
+                    options.MaxConnectionsPerBroker)),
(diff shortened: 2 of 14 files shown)


thomhurst:
Addressed all review findings in `79f7389e`:

1. Response-buffer retention now sizes from `MaxConnectionsPerBroker`; added a call-site test that reflects the actual consumer pool and distinguishes baseline 1 from adaptive ceiling 4.
2. Malformed mid-fetch cleanup now uses `ReturnRecordBatchListAfterFailedParse`, returning every successfully parsed pooled batch.
3. Share-consumer parsed batches now use `DisposeAndReturnUnownedConsumerBatch()` in `finally`, including deserializer exception paths.

Validation: focused sizing/parse/pooling/share tests 68/68; full unit suite 5303/5303.

thomhurst:
Addressed current-head review at b36b982a825148d08527b3e0cf3d0cea5cd1ed5e: added a ShareConsumer regression proving a batch returns to RecordBatch pool when value deserialization throws, and corrected Dispose() XML docs to state that it does not return the object. Verified build (0 warnings), regression 5/5, full unit suite 5304/5304.

PR #1953, opened 2026-07-13
Isolate global MeterListener tests

### LINKED ISSUE #1949: P2: Isolate MeterListener tests from concurrent publication
## Summary

Full unit suite can fail `DekafDiagnosticsTests.MetricInstruments_HaveCorrectNames` because its `MeterListener` runs concurrently with other listener-based tests. The class is `[NotInParallel("ActivityListener")]`, but the metric test does not join the shared `MeterListener` exclusion used by producer/protocol/stress tests.

Observed while validating #1947:

```
Expected to contain messaging.consumer.lag
but the item was not found in the collection
```

The focused diagnostics tests pass; the 5,297-test suite produced 5,296 passes and this one failure. Other unisolated listeners also exist in `ConsumeOneFastPathTests` and `KafkaConsumerFetchMetricsTests`.

## Acceptance criteria

- Put every test using global `MeterListener` publication state in one shared non-parallel group.
- Preserve parallelism for tests that do not use global listener state.
- Run the full unit suite repeatedly enough to cover listener publication ordering.

Discovered during #1947.

thomhurst:
## Summary
- isolate every global `MeterListener` test in one shared non-parallel group
- move instrument publication coverage out of the activity-listener-isolated class
- retain existing class-wide isolation for consumer fixture tests while allowing unrelated tests to run in parallel

## Test plan
- [x] `dotnet build tests/Dekaf.Tests.Unit/Dekaf.Tests.Unit.csproj --configuration Release --no-restore`
- [x] three affected listener tests individually (all passed)
- [x] full unit suite twice consecutively (5294/5294 each)

Closes #1949

### DIFF
--- a/tests/Dekaf.Tests.Unit/Consumer/ConsumeOneFastPathTests.cs
+++ b/tests/Dekaf.Tests.Unit/Consumer/ConsumeOneFastPathTests.cs
@@ -11,7 +11,7 @@ using Microsoft.Extensions.Logging;
 
 namespace Dekaf.Tests.Unit.Consumer;
 
-[NotInParallel]
+[NotInParallel("MeterListener")]
 public sealed class ConsumeOneFastPathTests
 {
     private const string Topic = "test-topic";
--- a/tests/Dekaf.Tests.Unit/Consumer/ConsumeOneFastPathTests.cs
+++ b/tests/Dekaf.Tests.Unit/Consumer/ConsumeOneFastPathTests.cs
@@ -349,6 +349,7 @@ public sealed class ConsumeOneFastPathTests
     [Test]
     public async Task ConsumeOneAsync_EmitsFetchMetricsAsDeltas()
     {
+        var topic = $"metrics-test-{Guid.NewGuid():N}";
         var messagesReceived = new List<long>();
         var bytesReceived = new List<long>();
         using var listener = new MeterListener();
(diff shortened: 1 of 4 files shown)


thomhurst:
Addressed the blocking review at 77d61c8b60d478ac60d8d609fd68ebb9c0612504: metric callbacks now filter by a GUID-scoped topic or GUID-derived broker ID, so unrelated concurrent consumer activity cannot pollute exact assertions. Verified build (0 warnings), focused tests (27/27), and two full unit runs (5294/5294 each).

thomhurst:
Addressed current-head review at e42d2b0dac6678dcc78bcb069fe2883eaa81cf3e: moved listener isolation from class scope to the one ConsumeOne listener test and both fetch-metric tests. The disabled fetch test must remain isolated because it asserts process-global Instrument.Enabled; a focused parallel run proved this (cache count became 1), then passed after method isolation. This frees 23 unrelated fast-path tests. Rebased onto main; build 0 warnings, focused 27/27, full 5297/5297.

thomhurst:
Addressed current-head cleanup review at 7b93327513b0272424d211eb9748a7b94b672828: ConsumeOneFastPathTests now reuses AccumulatorTestHelpers.GetTag; deleted duplicate helper. Rebased onto current main. Verified build 0 warnings, focused 27/27, full unit 5310/5310.

thomhurst:
Correction: merge completed server-side at e8c23cadcd7b39414b9031958d99a526e762d7db. Wrapper returned nonzero only because the local branch was checked out in a linked worktree; clean worktree removed afterward.

PR #1955, opened 2026-07-13
Stabilize adaptive scale-down reap test

### LINKED ISSUE #1952: P2: Stabilize adaptive scale-down connection reap count
## Summary

Full unit-suite execution can fail `AdaptiveScaleDownTests.ScaleDown_ReleasesRemovedIndex_AndWidthOneRetainsGroupSlot` because two idle connections are observed as reaped where the test expects exactly one.

Observed while validating #1949:

```
Expected to be 1
but found 2
at AdaptiveScaleDownTests.cs:378
```

The test is timing-sensitive around scale-down and idle-reaper ownership; an isolated run should be used to distinguish cross-test shared state from an internal race, not as evidence that the suite failure is harmless.

## Acceptance criteria

- Identify why a full-suite run can reap two connections during the width-one retention case.
- Make connection ownership/reap expectations deterministic without relaxing the invariant.
- Verify with repeated focused runs and full unit suites.

Discovered during #1949.

thomhurst:
## Summary
- publish broker leader metadata before testing routed-slot retention
- prevent the background idle sender from legitimately releasing the slot before manual reap
- preserve the exact one-connection reap invariant

## Test plan
- [x] affected test repeated 5 times
- [x] full unit suite twice consecutively (5297/5297 each)

Closes #1952

### DIFF
--- a/tests/Dekaf.Tests.Unit/Producer/AdaptiveScaleDownTests.cs
+++ b/tests/Dekaf.Tests.Unit/Producer/AdaptiveScaleDownTests.cs
@@ -352,6 +352,8 @@ public sealed class AdaptiveScaleDownTests
                 await pool.GetConnectionAsync(1);
 
                 var senderType = typeof(BrokerSender);
+                GetField<MetadataManager>(sender, "_metadataManager")
+                    .Metadata.Update(CreateMetadata(leaderId: 1));
                 var getConnection = senderType.GetMethod(
                     "GetConnectionForPartition",
                     BindingFlags.Instance | BindingFlags.NonPublic)!;


thomhurst:
Merge gate passed (MERGEABLE/CLEAN, 19 checks green, no unresolved threads), but Merge-Pr.ps1 aborted because local branch issue-1952-scale-down-reap-race is still checked out at C:/git/Dekaf-worktrees/issue-1952-scale-down-reap-race, so gh could not delete it. Nothing merged or removed; defer to a later iteration after worktree ownership clears.

thomhurst:
Correction: authoritative GitHub state shows #1955 merged successfully at 2026-07-13T01:45:38Z (merge commit 663fb6d0). The wrapper's nonzero exit was solely the post-merge local branch deletion failure; the cleanup sweep has now removed the merged worktree.

PR #1954, opened 2026-07-13
Restore round-trip batching and phase diagnostics

thomhurst:
## Summary

- form one bounded cross-partition request wave after producer idle without stalling an active pipeline
- warm both round-trip producers before timing and keep Dekaf on the requested fixed connection profile
- report separate produce and consume rates instead of hiding the regressed phase in one blended number
- expose per-broker request rate, average coalesce width, and coalesce-width histogram
- make request-timeout tests deterministic through the injected monotonic clock

## Root cause

After non-idempotent mute-on-send was removed, narrow partition batches could be dispatched before sibling partitions reached the sender. The old 20-iteration spin covered only a few microseconds. The stress harness also timed producer construction, blended produce and consume phases, and let Dekaf adapt beyond the fixed connection count used by Confluent, obscuring both the regression and ordering behavior.

## Validation

- `dotnet build tests/Dekaf.Tests.Unit --configuration Release --no-restore` — 0 warnings, 0 errors
- net10 unit suite — 5,296 passed
- Kafka 4.3.1 round-trip stress, 250,000 × 1000-byte messages, 16KB batches, 5ms linger, both clients:
  - Dekaf: 250,000 consumed, 0 missing/duplicate/corrupt/out-of-order, 70,522 produce msg/s
  - Confluent: 250,000 consumed, 0 missing/duplicate/corrupt/out-of-order, 70,622 produce msg/s
  - blended round-trip: Dekaf 53,507 msg/s vs Confluent 44,815 msg/s (1.19x)

Closes #1940

PR #1956, opened 2026-07-13
Avoid unused producer connections

### LINKED ISSUE #1945: P2: Idle reaper reaps brokerId=-1 connection at 10min every run; 3conn config opens connections dense routing can never use
## Summary

Two related oddities in `producerDeliveryDiagnostics.connectionReapEvents` from stress run [29213011704](https://github.com/thomhurst/Dekaf/actions/runs/29213011704):

1. **Every producer scenario reports a reap of a connection with `brokerId: -1`, `connectionIndex: 0`, `idleDurationMs ≈ 599,000` at exactly the 10-minute mark.** A brokerId of -1 means the reaped connection was never associated with a real broker — most likely the initial bootstrap/metadata connection sitting idle since startup. Examples: FnF 1b `(-1, 0, 599044)`; idempotent 3b `(-1, 0, 599464)`. If it is the bootstrap connection, reaping is probably correct behavior but the event metadata is misleading (brokerId -1 with a connectionIndex suggests slot bookkeeping against a pseudo-broker); if it is a real data-path connection that was never routed to, that's a routing bug. Related history: #1859 (close-commit brokerId -1).

2. **On 3-broker runs with `ConnectionsPerBroker=3` (3conn variants), per-broker connections are reaped after sitting idle for the entire run**: idempotent 3b 3conn reaped `(1,2)`, `(2,1)`, `(3,1)` each at ~599.9s idle. With 6 partitions over 3 brokers = 2 partitions per broker, dense partition-affined routing (#1918) can only ever use 2 of the 3 configured connections per broker — the third is created, never receives traffic, and burns a socket + broker resources for 10 minutes until the reaper collects it.

## Asks

1. Fix the reap-event metadata so bootstrap/pseudo-broker connections are identifiable (or excluded) — diagnostics consumers currently can't distinguish "harmless bootstrap cleanup" from "data connection starved by routing" (#1913 diagnostics-gap follow-up).
2. Don't open more connections per broker than the partition-affinity routing can use (`min(ConnectionsPerBroker, partitionsOnBroker)` at group creation), or route round-robin overflow onto the extra connections. Creating-then-reaping is wasted work and skews connection-count diagnostics.
3. Confirm the 10-minute idle reap of the brokerId -1 connection can't disturb in-flight work at reap time (the earlier idle-reap producer-slot invalidation was #1910/#1917; the reap here lands at minute 10, after the acks-all dip at minutes 6–8 in #1944, so it's not that dip's cause — but the interaction surface remains).

thomhurst:
## Summary
- create indexed broker connections lazily as partition-affinity routes request each slot
- mark bootstrap connection reap diagnostics explicitly
- retire and drain active leases/operations before idle disposal

## Test plan
- [x] `Dekaf.Tests.Unit.exe` (5,297 passed)
- [x] focused connection/reaper tests (86 passed post-rebase)
- [x] `MultiConnectionProducerTests` integration (8 passed)

Closes #1945

### DIFF
--- a/src/Dekaf/Networking/ConnectionPool.cs
+++ b/src/Dekaf/Networking/ConnectionPool.cs
@@ -15,6 +15,7 @@ namespace Dekaf.Networking;
 public sealed partial class ConnectionPool : IConnectionPool
 {
     private const int MaxConnectionReapDiagnosticEvents = 256;
+    private static readonly TimeSpan DefaultIdleReapDrainTimeout = TimeSpan.FromSeconds(5);
     private readonly string? _clientId;
     private readonly ConnectionOptions _connectionOptions;
     private readonly ILoggerFactory? _loggerFactory;
@@ -22,6 +23,7 @@ public sealed partial class ConnectionPool : IConnectionPool
     private readonly int _connectionsPerBroker;
     private readonly ResponseBufferPool _responseBufferPool;
     private readonly ClientTelemetryMetricCollector? _telemetryMetricCollector;
+    private readonly TimeSpan _idleReapDrainTimeout;
     private readonly OAuthBearerTokenProvider? _sharedOAuthBearerTokenProvider;
     private CancellationTokenSource? _idleReaperCts;
     private Task? _idleReaperTask;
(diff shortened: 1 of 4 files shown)


thomhurst:
Addressed both blocking findings in 976638483b7d7046ce5edb5005ba1680cde35bae:

- Recheck races now return an explicit restore outcome, so a request arriving before retirement preserves and restores the healthy connection.
- Idle retirement drain is bounded (5s production default); timed-out drains continue detached with observed failures, so pool disposal cannot remain blocked under the disposal lock.

Regression coverage added for both paths. Validation: 46 focused unit tests, 5,302 full unit tests, 8 MultiConnectionProducer integration tests, and full Release solution build (0 warnings/errors).

thomhurst:
CI diagnosis: faults failed only in BufferMemoryStressTests.SustainedLoad_DoesNotCauseUnboundedMemoryGrowth with an OOM/8.25GB final growth; consume was then canceled by fail-fast. This is the same unrelated stress failure observed on #1954, while all unit/build/quality and other integration lanes pass and review is clear. Requested failed-lane rerun; no PR code change is supported by the evidence.

PR #1958, opened 2026-07-13
Coalesce partition-limit diagnostics

### LINKED ISSUE #1943: P2: Partition-limited no-op scale events spam diagnostics (~6-10/s all run); stress timeline truncates at ~55s hiding late-run activity
## Summary

When adaptive connection scaling is capped at the partition count, the scale governor records a **partition-limited no-op scale event (`6→6`) every governor pass (~every 100–300ms) for the entire run** — thousands of events per run. In stress run [29213011704](https://github.com/thomhurst/Dekaf/actions/runs/29213011704) every 1-broker producer timeline is flooded with `6→6 | 0/100` rows; the markdown reporter's row cap truncates the table at ~55s of a 15-minute run, so **all late-run scale activity is invisible** — exactly where the acks-all mid-run dip (minutes 6–8, see companion issue) happened.

## Mechanism

- `RecordSendLoopPressureIfScaleUseful()` (`src/Dekaf/Producer/BrokerSender.cs:2371-2389`) increments `_partitionLimitedPressureEvents` on every send-loop iteration that hits the in-flight ceiling while `connectionCount == knownPartitions.Count`.
- The governor (`BrokerSender.cs:4943-4960`) records a diagnostics event whenever the delta ≥ `ScalePressureDeltaThreshold` (100) and resets the snapshot — under sustained ceiling pressure that's every pass, forever. The event is pure bookkeeping (old == new == 6); nothing can change because scaling is partition-capped.
- Each event allocates/stores a diagnostics record in the accumulator and lands in `producerDeliveryDiagnostics`/stress artifacts (hundreds of KB per scenario artifact).

## Why it matters

1. **Signal drowned**: real scale transitions (1→2, 2→5, 5→6) are 3 rows; the other ~250+ rows are identical no-ops. The reporter cap then hides the rest of the run.
2. **The condition itself is the finding**: sustained partition-limited pressure means the producer *wants* more parallelism but the topology forbids it — that's a steady-state condition worth one summarized signal (count + duration), not an event stream.
3. Minor steady-state overhead in the send loop / governor and artifact bloat.

## Fix plan

1. Rate-limit / coalesce partition-limited events: emit on state entry, then aggregate (count, max delta, duration) and emit at most every N seconds or on state exit.
2. In the stress reporter (`tools/Dekaf.StressTests/Reporting/MarkdownReporter.cs`), when rows exceed the cap, sample across the whole run (first/last + decimated middle) instead of truncating the tail, and print an "N events omitted" note.
3. Consider surfacing "partition-limited for X% of run" as a first-class scenario metric — it directly explains ceilings like #1938.

thomhurst:
## Summary
- coalesce partition-limited pressure into bounded 15-second summaries
- record observation count, duration, and maximum pressure per summary
- evenly sample long scale timelines across first/middle/last rows
- report omitted event counts instead of silently hiding late-run behavior

## Test plan
- [x] `Dekaf.Tests.Unit.exe` (5,299 passed)
- [x] focused adaptive-scale/reporting diagnostics tests (52 passed)

Closes #1943

### DIFF
--- a/src/Dekaf/Producer/BrokerSender.cs
+++ b/src/Dekaf/Producer/BrokerSender.cs
@@ -555,6 +555,11 @@ internal sealed partial class BrokerSender : IAsyncDisposable
     private long _lastSendLoopPressureSnapshot;
     private long _partitionLimitedPressureEvents;
     private long _lastPartitionLimitedPressureSnapshot;
+    private long _lastPartitionLimitedDiagnosticMs;
+    private long _partitionLimitedDiagnosticObservationCount;
+    private long _partitionLimitedDiagnosticMaxBufferPressureDelta;
+    private long _partitionLimitedDiagnosticMaxSendLoopPressureDelta;
+    private double _partitionLimitedDiagnosticMaxBufferUtilization;
     private long _lastScaleTimeTicks;
     private Task<int>? _pendingScaleTask; // Background connection creation, polled by send loop
     private double _pendingScaleBufferUtilization;
@@ -597,6 +602,7 @@ internal sealed partial class BrokerSender : IAsyncDisposable
     // so changing this value affects both trigger sensitivity and per-step magnitude.
     private const long ScalePressureDeltaThreshold = 100;
     private const long ScaleCooldownMs = 5_000;
+    private const long PartitionLimitedDiagnosticIntervalMs = 15_000;
     private const double ScaleUtilizationThreshold = 0.7;
 
     private const int MaxScaleStep = 3; // Cap per scale-up to avoid over-provisioning from a brief spike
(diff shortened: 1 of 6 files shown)


thomhurst:
Addressed current-head review at 8c38b5ad546524562dcca205d048b9ed0eeade0f: send-loop shutdown now flushes any partial partition-limited diagnostic window, with a disposal regression preserving observation/max-pressure data. Also rebased onto current main (including #1955). Verified build (0 warnings), affected tests (38/38), full unit suite (5300/5300).

thomhurst:
Correction: merge completed successfully server-side at ec5d9d57375e21999ce0d99f946db79dc18d22e4. Wrapper returned nonzero only because the local branch remained checked out in a linked worktree; clean worktree removed afterward.

PR #1959, opened 2026-07-13
Restore delivery-coupled producer batching

### LINKED ISSUE #1938: P0: 1-broker producer throughput collapsed to 0.60-0.67x Confluent (compound batching regression: #1926 mute removal + #1929 micro-linger + #1920 permissive admission)
## Summary

Stress run [29213011704](https://github.com/thomhurst/Dekaf/actions/runs/29213011704) (commit 9e8cc648, first run on Kafka 4.3.1): **every 1-broker producer scenario collapsed ~40–49%** and Dekaf now loses to Confluent 0.60–0.67x on scenarios it won 1.13–1.44x in the previous run ([29202683474](https://github.com/thomhurst/Dekaf/actions/runs/29202683474), commit 0ee1abc4). Confluent *improved* 3–16% on the same run/VM, so the ratio flip is real, not VM noise.

| Scenario (1 broker, 1000B, 15m) | Dekaf prev → now | Confluent prev → now | Ratio prev → now |
|---|---|---|---|
| FnF | 1,685,818 → 1,017,571 (-40%) | 1,283,014 → 1,484,760 (+16%) | 1.29x → **0.67x** |
| FnF (3conn) | 1,858,785 → 941,749 (-49%) | — | 1.44x → **0.63x** |
| Acks All | 1,599,862 → 934,963 (-42%) | 1,380,230 → 1,422,679 (+3%) | 1.13x → **0.64x** |
| Idempotent | 1,426,631 → 909,101 (-36%) | 1,228,465 → 1,382,029 (+13%) | 1.15x → **0.64x** |
| Idempotent (3conn) | 1,608,872 → 833,256 (-48%) | — | 1.31x → **0.60x** |

All Dekaf producer results (1b *and* 3b) now cluster suspiciously at ~0.83–1.02M msg/s regardless of scenario — a common ceiling.

## Diagnostic evidence (from run artifacts + #1919 diagnostics)

- **Send loop pinned at the in-flight ceiling**: connection-scale timelines show constant partition-limited pressure events (`6→6`, delta exactly 100 = `ScalePressureDeltaThreshold`) every ~100–300ms for the entire run. These fire from `RecordSendLoopPressureIfScaleUseful()` when `pendingCount >= _totalMaxInFlight` (`src/Dekaf/Producer/BrokerSender.cs:1404-1446`) and scaling is capped at partition count (6).
- **Accumulator near-empty**: buffer utilization 0–2% all run; `brokerBudgets.admissionBlockCount = 1` total (admission gate never binds); sampled `unackedBytes` p50 = **0 MB** (p95 10.5MB) — the pipe is empty at most sample instants.
- **CPU idle**: 0.88–0.93 of 2 cores (previous run: 1.35). Not CPU-bound — wait-bound.
- **Latency dropped while throughput collapsed**: FnF p50 3.05ms → 1.85ms. Classic small-batch signature: throughput traded for latency nobody asked for.
- **Ceiling math**: with 6 connections, idempotent `_totalMaxInFlight` = 6 × 5 = 30 slots (`BrokerSender.cs:763`; `MaxInFlightRequestsPerConnection` idempotent default 5, `ProducerOptions.cs:18`). 30 slots × ~64KB requests / ~2ms loaded RTT ≈ **0.96M msg/s — exactly the observed plateau**. The per-connection *bytes* ceiling is only consulted when `_connectionCount == 1` (`BrokerSender.cs:1414-1415`), so only the count ceiling gates at 6 connections.

## Root cause: compound batching regression in 0ee1abc4..9e8cc648

Each ProduceRequest carries at most **one batch per partition** per coalesce pass, and with partition-affined bucketing (1 partition per connection at 6/6) that means **one batch per request** — so request size == sealed batch size. Three PRs in the window compounded to make sealed batches far smaller than the 1MB `BatchSize`:

1. **#1926 (04c645cd) "Pipeline non-idempotent same-partition producer sends"** — `_muteOnSend` flipped from `!_isIdempotent || _maxInFlight <= 1` (true for all non-idempotent) to `_maxInFlight <= 1 || Acks == Acks.None` (false for the stress acks=1/all scenarios) (`BrokerSender.cs:758`). Pre-#1926, a muted partition **deferred linger sealing** (`RecordAccumulator.cs:4261-4263` → `ShouldDeferPartialBatchSeal:3545-3553`), so batches filled toward 1MB during each in-flight RTT — a self-clocked full-batch pump. `_serializeBatchesPerPartition` also flipped true→false for non-idempotent (`RecordAccumulator.cs:1986`), removing the second defer condition. Post-#1926 nothing defers seals.
2. **#1929 (8e6295ef) "arm linger scheduler"** — removed the 1ms `PeriodicTimer` quantization; every awaited append signals the linger loop, and any batch containing a completion source seals at `min(1ms, LingerMs/10)` = **0.5ms** (`RecordAccumulator.cs:6180-6199`). The stress harness plants an awaited `ProduceAsync` every 1000 messages (`StressTestHelpers.cs:19`), i.e. ~1–1.9k sampled batches/sec, so statistically most batches per partition seal at ~0.5ms age instead of filling.
3. **#1920 (315e538b) "Fix pipelined admission rate estimation"** — the admission gate became strictly more permissive (zero block events all run vs the pre-#1920 low-biased gate that blocked appenders). Blocked appenders previously let backlog form so batches size-sealed at 1MB; with no backpressure the accumulator stays pinned near-empty and the linger clock rules batch size. This is the only in-range change besides #1929 that alters **idempotent** batch formation — important because #1926 didn't touch the idempotent path, yet idempotent collapsed too.

Possible amplifiers (need runtime data):
- **#1934 (9e8cc648) Kafka 4.3.1**: the workflow moved `apache/kafka:latest` → `4.3.1`; what `latest` resolved to in the healthy run is unknown. Higher per-request RTT multiplies directly into the 30-slot ceiling. Confluent improving doesn't exonerate this — librdkafka coalesces all partitions into fewer, larger requests and is far less sensitive to per-request cost.
- **#1927 (9214050f)**: plain producers now negotiate Produce **v12** (previously v11) — different broker-side code path for every produce. Cheap to A/B by capping to v11.

## Discriminating experiments

1. Log per-request `EncodedBytes` + ack RTT (`PendingResponse.RttStartTime`, `BrokerSender.cs:3148-3151`): "~60–500KB at 2–5ms" ⇒ client batching regression confirmed; "~1MB at 15–30ms" ⇒ broker RTT.
2. Rerun 9e8cc648 stress against the previous broker image (isolates #1934).
3. Revert #1929 alone (or raise `LatencySampleInterval`) and rerun FnF/idempotent.
4. A/B Produce v11 cap (isolates #1927).

## Fix plan (goal: restore >1.3x vs Confluent on all 1b scenarios)

1. **Restore delivery-coupled batch formation without re-serializing sends**: defer partial (non-size) seals for a partition while it has ≥N in-flight requests and the current batch is below a fill threshold — i.e., keep #1926's pipelining for full batches but re-create the "fill while in flight" behavior the mute provided for partial ones. Alternatively: a min-request-bytes wave barrier in the coalesce pass (don't dispatch a request far below `BatchSize` while more batches will seal within the current request cycle — cf. the 20-iteration micro-linger spin at `BrokerSender.cs:1359-1384`, which is time-bounded to a few μs and cannot bridge a 0.5ms seal cadence).
2. **Make awaited-append micro-linger (0.5ms) not apply under load**: a single sampled message in a hot batch shouldn't reduce that batch's linger 10x; scale the micro-linger by buffer/in-flight pressure or only apply it when the partition is otherwise idle.
3. **Reinstate effective admission pacing** once the estimator is fixed (tracked separately — see companion issue on `BrokerUnackedByteBudget`): the gate blocking appenders is what let batches size-seal.
4. Add per-request size + RTT histograms to `producerDeliveryDiagnostics` so this failure mode is visible directly instead of via inference (extends #1913/#1919).

Related history: #1856/#1857 (budget stuck-low), #1863/#1888 (BBR-style fix), #1911 (k/N rate bias), #1584 (send-loop core pegging).

thomhurst:
## Summary

- defer partial linger sealing while the same partition has an earlier batch in the delivery pipeline
- continue sealing size-full batches immediately, preserving non-idempotent and idempotent pipelining
- resume normal linger as soon as the predecessor exits, so sparse traffic does not wait for a full batch
- update zero-linger and purge coverage for delivery-coupled batching and contiguous idempotent sequences

## Root cause

Removing non-idempotent mute-on-send also removed the delivery clock that kept partial batches open during an in-flight request. Linger then sealed small batches even under sustained load. With partition-affined connections, each small batch became its own ProduceRequest and throughput hit the request-count ceiling.

The fix uses the existing per-partition pipeline-byte tracker as the delivery signal. A partial batch remains open while a predecessor exists; a full batch still seals and pipelines immediately. When traffic becomes sparse, the predecessor acknowledgement clears the signal and the next linger pass flushes normally.

## Validation

- `dotnet build tests/Dekaf.Tests.Unit --configuration Release --no-restore` — net8/net10, 0 warnings, 0 errors
- net10 unit suite — 5,303 passed
- focused accumulator/linger suite — 131 passed
- Kafka 4.3.1, 1-minute idempotent producer A/B:
  - no delivery coupling: 159,013 msg/s
  - full-batch delivery coupling: 570,825 msg/s (+259%)
  - paired Confluent result: 253,911 msg/s (Dekaf 2.25x)
  - 100% broker-confirmed delivery, 0 errors

Closes #1938

### DIFF
--- a/src/Dekaf/Producer/RecordAccumulator.cs
+++ b/src/Dekaf/Producer/RecordAccumulator.cs
@@ -3552,7 +3552,7 @@ public sealed partial class RecordAccumulator : IAsyncDisposable
             _serializeBatchesPerPartition,
             hasPipelineBatch,
             batch.EstimatedSize,
-            _options.BatchSize);
+            batch.EffectiveBatchSizeLimit);
     }
 
     // An earlier pipeline batch provides a short delivery clock under load. Keep the next
@@ -5772,7 +5772,7 @@ internal sealed class PartitionBatch
     public bool IsExactlyAtSizeLimit =>
         (long)RecordBatch.TotalBatchHeaderSize + _encodedRecordsLength == EffectiveBatchSizeLimit;
 
-    private int EffectiveBatchSizeLimit => _effectiveBatchSizeLimit;
+    internal int EffectiveBatchSizeLimit => _effectiveBatchSizeLimit;
 
     private static int GetEffectiveBatchSizeLimit(
         TopicPartition topicPartition,
--- a/src/Dekaf/Producer/RecordAccumulator.cs
+++ b/src/Dekaf/Producer/RecordAccumulator.cs
@@ -3538,19 +3538,37 @@ public sealed partial class RecordAccumulator : IAsyncDisposable
     private bool ShouldSealAppendedBatch(PartitionBatch batch)
         => batch.IsExactlyAtSizeLimit
             || (_options.LingerMs == 0
-                && !ShouldDeferPartialBatchSeal(batch.TopicPartition)
+                && !ShouldDeferPartialBatchSeal(batch)
                 && batch.ShouldFlush(Stopwatch.GetTimestamp(), _options.LingerMs));
 
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    private bool ShouldDeferPartialBatchSeal(TopicPartition topicPartition)
+    private bool ShouldDeferPartialBatchSeal(PartitionBatch batch)
     {
-        if (_mutedPartitions.ContainsKey(topicPartition))
-            return true;
-
-        return _serializeBatchesPerPartition
-            && _partitionQueueBytes.TryGetValue(topicPartition, out var queuedBytes)
+        var topicPartition = batch.TopicPartition;
+        var hasPipelineBatch = _partitionQueueBytes.TryGetValue(topicPartition, out var queuedBytes)
             && queuedBytes > 0;
+        return ShouldDeferPartialBatchSeal(
+            _mutedPartitions.ContainsKey(topicPartition),
+            _serializeBatchesPerPartition,
+            hasPipelineBatch,
+            batch.EstimatedSize,
+            _options.BatchSize);
     }
+
+    // An earlier pipeline batch provides a short delivery clock under load. Keep the next
+    // partial batch open until it fills; if traffic becomes sparse, the predecessor exits,
+    // queued bytes reach zero, and the next linger pass seals normally. Size-full batches
+    // bypass this method in ShouldSealAppendedBatch and continue pipelining immediately.
+    internal static bool ShouldDeferPartialBatchSeal(
+        bool isMuted,
+        bool serializeBatchesPerPartition,
+        bool hasPipelineBatch,
+        int currentBatchSize,
+        int maximumBatchSize) =>
+        isMuted
+        || (hasPipelineBatch
+            && (serializeBatchesPerPartition
+                || currentBatchSize < maximumBatchSize));
 
     private ReadyBatch? CompleteDetachedBatchAndEnqueue(PartitionDeque pd, PartitionBatch batchToComplete)
     {
@@ -4259,7 +4277,7 @@ public sealed partial class RecordAccumulator : IAsyncDisposable
                 // allocating one BatchSize arena per record while they wait. Size-full batches
                 // still seal in the append path, and FlushAsync must always seal.
                 else if (sealAll
-                    || (!ShouldDeferPartialBatchSeal(topicPartition)
+                    || (!ShouldDeferPartialBatchSeal(pd.CurrentBatch)
                         && pd.CurrentBatch.ShouldFlush(now, _options.LingerMs)))
                 {
                     ProducerDebugCounters.RecordBatchFlushedFromDictionary();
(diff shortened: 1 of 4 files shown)


thomhurst:
Investigated the net8 failure: MultiplePartitionsMuted_IndependentlyBlocked_OtherPartitionsProceed timed out at 30s under the full parallel suite. This PR does not modify BrokerSenderMuteOrderingTests or its implementation path; the exact net8 test passes locally 3/3 in ~0.36s each. Review verdict is clear. Requested a failed-job rerun; will re-evaluate after CI completes.

Issue #1911, opened 2026-07-12
P0 perf: admission-budget rate estimator biased low on pipelined lanes — idempotent 1-broker -13.4% (repeated-regression gate fail), slow ramp, late-run decay


thomhurst:
## Evidence (stress run 29202683474, commit 0ee1abc4 — first run with #1863 + #1888)

The #1863/#1888 budget rework fixed the 1-broker latency gates (p99 ratio 2.14x FAIL → 0.26x PASS) and acks-all 3b (big improvement), but three symptoms remain, all traceable to the same estimator:

1. `producer-idempotent / Dekaf / 1 broker`: 1,426,631 msg/s vs baseline 1,647,258 (band 1,512k-1,782k) — **repeated regression, gate FAIL**; CPU 0.85 vs 0.79 μs/msg — **gate FAIL**. History: 1,647k → 1,715k healthy (07-11 am), regressed to 1,466k when #1843 landed, **not recovered** by #1863/#1888. The 3conn idempotent variant is unaffected (1,609k, within band).
2. `producer / Dekaf / 1 broker`: intra-run drift **+12.52%** — throughput ramps up for minutes, so the run starts well below capacity.
3. `producer-acks-all / Dekaf / 1 broker`: steady-state/peak exactly 0.85 (borderline breach), drift -5.38% — budget collapses on transient broker slowdowns and recovers only by luck.

## Root cause

The per-response-pass rate sample is **structurally biased low by pipeline depth, and the bias is scale-invariant** — the +25% capacity probe can never correct it:

- Sample = `passAckedBytes / (ackNow - earliestRequestStartInPass)` (BrokerSender.cs:2324-2334 → `BrokerUnackedByteBudget.AddRateSample`, src/Dekaf/Producer/BrokerUnackedByteBudget.cs:283-308), fed to a 10-sample max filter (`RateWindowSamples = 10`, line 34).
- Idempotent single-connection keeps `MaxInFlightRequestsPerConnection = 5` requests pipelined (`_muteOnSend = false`, BrokerSender.cs:681). The broker services one connection serially, so a request's sojourn ≈ 5 × service_time + base RTT, while a pass typically acks k = 1-2 requests. Sample ≈ **(k/5) × true drain rate** — and the bias is independent of request size, so probing bigger budgets (bigger requests) grows numerator and denominator together. The ratchet-up channel is dead; this is the residual of the #1856 stuck-low failure mode.
[...]

Issue #1856, opened 2026-07-12
P0: #1843 unacked-byte budget regresses producer throughput 17-33%; Dekaf now loses to Confluent on 1b producer/idempotent and 3b acks-all


thomhurst:
Stress run: https://github.com/thomhurst/Dekaf/actions/runs/29190595109 (first run containing #1843 / aa44dc22, merged 2026-07-12 10:49 UTC; the trend baseline of the prior ~10 runs is entirely pre-#1843).

## Symptom

Producer throughput dropped across **every** producer scenario versus Dekaf's own trailing baseline, while paired Confluent rows on the same VMs *improved* — so this is a client-side change, not runner noise:

| Scenario | Current | Baseline median | Delta | Trend gate |
|---|---:|---:|---:|---|
| producer-idempotent / 1b | 1,192,908 | 1,647,258 | **-28%** | Repeated regression (fail) |
| producer-idempotent / 1b / 3conn | 1,084,176 | 1,621,440 | **-33%** | Repeated regression (fail) |
| producer / 1b | 1,344,029 | 1,621,118 | -17% | Regression + CPU µs/msg repeated fail (0.85 vs 0.79) |
| producer / 1b / 3conn | 1,347,803 | 1,649,315 | -18% | Regression + CPU µs/msg repeated fail |
| producer / 3b | 949,124 | 1,205,258 | -21% | Regression (warning) |
| producer / 3b / 3conn | 953,684 | 1,182,840 | -19% | Steady-state 0.81 repeated breach (fail) |
| producer-acks-all / 3b | 732,759 | 938,029 | **-22%** | Regression (warning) |
| producer-idempotent / 3b / 3conn | 969,419 | 1,100,268 | -12% | Repeated regression (fail) |

Consequences vs Confluent in the same run:

- **producer-idempotent 1b: Dekaf 1.19M vs Confluent 1.44M (-17%); 3conn 1.08M (-25%).** Pre-#1843 Dekaf led this scenario (~1.65M vs ~1.25M).
- **producer 1b: Dekaf 1.34M vs Confluent 1.40M (-4%).** Previously led.
- **producer-acks-all 3b: Dekaf 733k vs Confluent 838k (-13%).** Baseline 938k previously led.

There is also **within-run decay** (the budget ratchets down over the 15-minute run):

- producer-idempotent 1b 3conn: intra-run drift **-10.8%**, slope **-1.12%/min** (breach), steady-state/peak 0.825 (breach)
[...]

PR #1960, opened 2026-07-13
Fix unacked-budget rate/RTT estimation with BBR per-request delivery sampling

### LINKED ISSUE #1941: P1: BrokerUnackedByteBudget rate/RTT estimator produces TB/s and 2us samples -> budget pegged at cap -> DeliveryLatencyTargetMs unenforced (3-broker p99 281ms vs Confluent 20ms)
## Summary

`BrokerUnackedByteBudget`'s drain-rate and RTT measurements are physically impossible, so the computed budget saturates at its cap and **`DeliveryLatencyTargetMs` (default 10ms) is never enforced**. Result in stress run [29213011704](https://github.com/thomhurst/Dekaf/actions/runs/29213011704): 3-broker scenarios win throughput but fail every latency gate:

| Scenario (3 brokers) | Dekaf p50/p95/p99 | Confluent p50/p95/p99 | Gate |
|---|---|---|---|
| Idempotent | 13.25ms / **107.9ms** / **281ms** | 6.55ms / 14.5ms / 20.4ms | p95/target=10.79 FAIL, p99/Confluent=13.81 FAIL, p50/Confluent=2.02 FAIL |
| Idempotent (3conn) | 13.45ms / **115.6ms** / **364ms** | — | p95/target=11.55 FAIL |
| Acks All | — | — | p95/target=8.16 FAIL, p99/Confluent=3.15 FAIL |
| FnF | — | — | p95/target=3.31 FAIL, p99/Confluent=3.06 FAIL |

## Evidence (producerDeliveryDiagnostics, #1919)

FnF 1-broker end-of-run broker budget: `maxRateBytesPerSec = 1,067,807,339,449` (**1.07 TB/s**), `minRttMicros = 4`. Mid-run samples: median maxRate **806 GB/s**, peaks 1.6 TB/s; minRtt samples as low as **2μs**. Real link throughput is ~1 GB/s and localhost RTT ~100–400μs. Consequences:

- `budget = windowMaxRate × horizon` is astronomically large → always clamped to `_capBytes` = `BatchSize × 32 × connections` (`BrokerUnackedByteBudget.ComputeCap`, `CapBatchMultiplier=32`) = **192MB** at 6 connections / **67MB** per broker at 2 connections.
- Idempotent 3b observed standing `unackedBytes` up to **68MB per broker** (p95 45–55MB) with budget pinned at the 67MB cap — ~200ms of queueing at ~330MB/s per-broker drain. That is the p95=108ms / p99=281ms bufferbloat, 10–28x the 10ms target.
- On 1-broker scenarios the same brokenness makes the gate fully permissive (1 block event / 15 min), which removes the append backpressure that used to let batches size-seal — a contributing factor in #1938.

## Root cause

`OnAcked` (`src/Dekaf/Producer/BrokerUnackedByteBudget.cs:250-283`) computes `bytesPerSecond = ackedBytes × Stopwatch.Frequency / intervalTicks` with `intervalTicks = now - max(_lastAckTimestamp, requestStartTimestamp)`. When a pipelined response pass completes many requests back-to-back, consecutive `OnAcked` calls see near-zero intervals → rates in the TB/s range. `AddRateSample` then takes the **max** over a 2s window (`_windowMaxRate`), so a single degenerate sample poisons the budget for 2s — and under continuous pipelining, forever. Same anchoring problem yields 2–4μs "RTT" samples, defeating the min-RTT floor. #1920 ("Fix pipelined admission rate estimation") did not resolve this. Secondary concern: the occupancy floor `budget >= 1.5 × windowMaxOccupancy` (`RecomputeBudget`, line 544) is a positive feedback loop — higher occupancy raises the floor which admits more occupancy — bounded only by the cap it currently sits at anyway.

## Fix plan

1. **Measure delivery rate per response pass, not per ack**: aggregate all acked bytes of a pass and divide by the pass-to-pass interval, or accumulate bytes into the existing 100ms buckets and compute rate as `bucketBytesSum / bucketDuration` (windowed sum, not max-of-instantaneous). The BBR analogy wants max of *delivery-rate samples over full RTT windows*, not max of per-ack byte/interval quotients.
2. **Anchor RTT at request write completion** (per request, socket-write timestamp → response arrival), and discard sub-minimum samples that are artifacts of batch response processing.
3. **Cap or damp the occupancy floor** (e.g., only allow it to raise budget when the rate estimate is cold) to break the feedback loop.
4. Validation: rerun 3-broker idempotent stress; success = p95 ≤ 3× target (gate), p99/Confluent ≤ 2.0, no throughput loss vs current run. Watch `unackedBytes` p95 ≈ `rate × 10ms` (~3–4MB per broker), not 45–68MB.

Related: #1856/#1857 (original bufferbloat + horizon inflation), #1863/#1888 (BBR-style rework), #1920 (estimation fix attempt), #1938 (admission permissiveness contributing to 1b collapse).

thomhurst:
Fixes #1941. Refs #1938.

## Problem

The per-broker unacked-byte admission budget never enforced `DeliveryLatencyTargetMs` (10ms) because both estimator inputs were garbage (stress run [29213011704](https://github.com/thomhurst/Dekaf/actions/runs/29213011704)):

- **Rate samples of 806 GB/s median / 1.6 TB/s peak**: `OnAcked` divided pass-aggregated bytes by `now - max(lastAck, earliestRequestStart)`. The send loop hot-polls `ProcessCompletedResponses` while at the in-flight ceiling, so consecutive passes run microseconds apart and the interval collapses to the inter-pass gap. The 2s windowed MAX filter then retained the explosion permanently — budget pinned at cap (67–192MB), 3-broker idempotent p99 **281ms vs Confluent 20ms**, every 3b latency gate red.
- **minRtt samples of 2–4µs**: the RTT anchor was stamped *after* `await SendPipelinedAfterWriteAsync(...)` returned; on localhost the response can already be in, so the anchor measured poll-loop latency and disabled the 1.5×BDP RTT safety floor.
- **Occupancy-floor feedback**: `budget >= 1.5 × occupancy` where occupancy is itself admission-bounded by the budget — a positive feedback loop ratcheting toward cap.

This is the third estimator generation to fail by measuring through response-pass timing (#1856/#1857 → #1863/#1888 → #1920).

## Fix: BBR-style per-request delivery sampling

- Each request mints a `DeliverySnapshot` token (cumulative delivered bytes, delivery timestamp, **pre-write** send timestamp, app-limited flag) just before its socket write and redeems it on ack. The rate sample is the delivered-counter delta over `max(request sojourn, ack-axis interval)` — the denominator is never smaller than one real round trip, regardless of pass clustering or processing stalls. App-limited sends (pipe empty at send) fall back to the sojourn so idle gaps never deflate the estimate.
- The pre-write anchor makes 2–4µs RTT artifacts impossible by construction; its write-time inflation only biases individual samples high, which the min filter discards.
- The occupancy floor is bounded at `1.5 × rate budget`: fixed point becomes 1.5× the latency target instead of the cap, while #1911's headroom-above-demonstrated-demand protection survives (recovery beyond it is the capacity probes' job, which the sane rate samples make functional again).
- Min-RTT probe and capacity-probe mechanisms are unchanged; only their inputs are re-based.
- `AckedResponsePass` deleted; `OnAcked` is now called once per acked request.

## Diagnostics (for stress validation + #1938)

Per-request size and RTT log2 histograms on the budget, exposed via `ProducerBrokerBudgetDiagnostic` on live snapshots only (periodic 1s samples omit them to keep the 4096-entry ring compact). These directly answer #1938's discriminating question (request-size distribution) on the next stress run.

## Tests

- All 33 existing estimator tests re-pinned by hand-derivation (29 keep exact values; the pipelined-rate test re-derives under delta semantics; two occupancy tests re-pin to the damped bound; the expired-window test's final value changes 1,111→1,000 because pipelined credit now flows through the delivered counter, not a shrunken inter-ack interval).
- New regression tests: hot-poll back-to-back acks (the stress failure shape — would have read TB/s before), ack-axis stall spreading, app-limited idle non-deflation, occupancy-floor feedback convergence, histogram bucketing, and a sender-level test proving a synchronous response behind a ~20ms write can no longer produce microsecond min-RTT samples.
- Full unit sweep: **5302/5302 green**. Benchmark: **~20ns/op, 0 B allocated** per acked request (was 12.3ns/op per pass; per-batch scale).

## Expected stress outcome (validation gates)

- `maxRateBytesPerSec` ≈ 1e9, not 1e12; `minRttMicros` in [100, 1000]
- Budget ≈ 10–15MB at 1GB/s (1-broker) / 3–5MB per broker (3-broker) instead of cap
- `unackedBytes` p95 ≈ rate × 10ms; 3b idempotent p95/target ≤ 3, p99/Confluent ≤ 2, no 3b throughput loss
- 1b `admissionBlockCount` becomes non-trivial — the gate binding again restores append backpressure, which should partially restore batch size-sealing (#1938 interplay)

If 1b idempotent throughput dips post-merge, the escape knob is `OccupancyFloorHeadroom` 1.5 → 2.0 before any design revisit.

### DIFF
--- a/src/Dekaf/Producer/BrokerSender.cs
+++ b/src/Dekaf/Producer/BrokerSender.cs
@@ -167,6 +167,13 @@ internal sealed partial class BrokerSender : IAsyncDisposable
     // request is sent and removed by ProcessCompletedResponses when the response task completes.
     // HandleTimedOutRequests (Java pattern) checks request timeout centrally and fails all
     // pending responses on timeout, invalidating the connection.
+    /// <summary>
+    /// Wraps a pipelined response with the batch-generation snapshot the send captured
+    /// at send start (before its first await) — the generations detect batch object
+    /// recycling between send and response processing. Ownership of the rented
+    /// <paramref name="BatchGenerations"/> and <paramref name="Batches"/> arrays transfers
+    /// to this PendingResponse; they are returned by <see cref="ReturnBatchesArray"/>.
+    /// </summary>
     private readonly record struct PendingResponse(
         PipelinedResponse<ProduceResponse> ResponseTask,
         ReadyBatch[] Batches,
@@ -174,21 +181,8 @@ internal sealed partial class BrokerSender : IAsyncDisposable
         int Count,
         long EncodedBytes,
         long RequestStartTime,
-        long RttStartTime)
+        BrokerUnackedByteBudget.DeliverySnapshot DeliverySnapshotAtSend)
     {
-        /// <summary>
-        /// Wraps a pipelined response with the batch-generation snapshot the send captured
-        /// at send start (before its first await) — the generations detect batch object
-        /// recycling between send and response processing. Ownership of the rented
-        /// <paramref name="generations"/> array transfers to this PendingResponse;
-        /// it is returned by <see cref="ReturnBatchesArray"/>.
-        /// </summary>
-        public static PendingResponse Create(
-            PipelinedResponse<ProduceResponse> responseTask, ReadyBatch[] batches, int[] generations,
-            int count, long encodedBytes, long requestStartTime, long rttStartTime)
-            => new(responseTask, batches, generations, count, encodedBytes,
-                requestStartTime, rttStartTime);
-
         /// <summary>
         /// Returns true if the batch at index <paramref name="i"/> is the same incarnation
         /// that was present when this PendingResponse was created.
@@ -205,20 +199,6 @@ internal sealed partial class BrokerSender : IAsyncDisposable
         {
             ArrayPool<int>.Shared.Return(BatchGenerations);
             ArrayPool<ReadyBatch>.Shared.Return(Batches, clearArray: true);
-        }
-    }
-
-    internal struct AckedResponsePass
-    {
-        public long Bytes { get; private set; }
-        public long EarliestRequestStart { get; private set; }
-
-        public void Add(long bytes, long requestStart)
-        {
-            EarliestRequestStart = Bytes == 0
-                ? requestStart
-                : Math.Min(EarliestRequestStart, requestStart);
-            Bytes += bytes;
         }
     }
 
@@ -2595,10 +2575,6 @@ internal sealed partial class BrokerSender : IAsyncDisposable
         CancellationToken cancellationToken,
         Dictionary<(string Topic, int Partition), ProduceResponsePartitionData>? reusableResponseLookup = null)
     {
-        // Aggregate every successful request completed in this pass. Multiple connection
-        // lanes drain concurrently, so per-request samples understate broker-wide capacity.
-        var ackedPass = new AckedResponsePass();
-
         // CRITICAL: Process responses in FORWARD order (oldest request first).
         // With multi-inflight, R1 and R2 may both complete with errors for the same partition.
         // Forward iteration ensures R1's retry batches are added to carry-over before R2's,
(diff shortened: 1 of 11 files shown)


thomhurst:
Fixed failing NativeAOT publish at fa9b3de61a2b1b22030cc862d81dfab5ce716dcf: integration ProducerTests now passes the required DeliverySnapshot to OnAcked. Rebased onto current main. Verified Release integration build with warnings-as-errors (0 warnings), estimator-focused unit tests 73/73, full unit rerun 5305/5305. The initial full run hit a pre-existing reflection timeout race, which passed isolated and on full rerun.

thomhurst:
Addressed current-head review at 6d122ad32a05fa72f8c805b860c7543b79c12e2d: added end-to-end send-loop coverage with two outstanding requests, proving real pending snapshots are AppLimited=true then alse; updated stale constructor comment. Rebased onto current main including #1958/#1959. Verified new test 5/5 before rebase, post-rebase focused 123/123, full unit 5330/5330, build 0 warnings.

PR #1970, opened 2026-07-13
Fix round-trip progress watchdog lifetime

### LINKED ISSUE #1963: P0 bug: Round-trip stress scenario crashes 'Progress watchdog already tracks an active scenario' — introduced by #1954, kills entire round-trip job
## What happened

Stress run [29220954180](https://github.com/thomhurst/Dekaf/actions/runs/29220954180) (main @ 888b4ef1): the **Producer → Consumer Round-Trip (Paired)** job failed ~40s after scenario start, before any measurement:

```
Error: System.InvalidOperationException: Progress watchdog already tracks an active scenario.
##[error]Process completed with exit code 1.
```

Deterministic, not flaky. Consequences:
- Zero round-trip data this run (`results-producer-roundtrip-1brokers` artifact is 203 bytes — empty envelope). The #1940 regression (round-trip 0.87x Confluent) is **unverifiable** post-fix.
- `Publish Stress Test Results to Documentation` and `Fail on Stress Regressions` were both **skipped**, so the 6 trend-gate breaches present in this run's JSON were never evaluated.

## Root cause

`ProducerRoundTripStressTest.cs` registers two overlapping watchdog registrations on the **same** `ProgressWatchdog`:

- Line ~86: `using var watchdog = options.ProgressWatchdog.Track(throughput, Client, Name, ...)` — produce phase, scoped to the whole method.
- Line ~146: `using var consumeWatchdog = options.ProgressWatchdog.Track(consumeProgress, Client, $"{Name}-consume", ...)` — consume phase, while the first registration is still alive.

`ProgressWatchdog.Track` throws when `_activeRun is not null` (`ProgressWatchdog.cs:156`). The second `Track` therefore always throws once the produce loop finishes.

Introduced by #1954 ("Restore round-trip batching and phase diagnostics"), which added the phase-separated consume watchdog (per #1946/#1948) without releasing the produce registration.

`SoakStressTest` already solves this correctly for its concurrent phases: it creates a second watchdog via `options.ProgressWatchdog.CreateSibling()` (`SoakStressTest.cs:119`).

## Fix

Either:
1. Scope the produce watchdog so it is disposed before the consume phase starts (phases are sequential — a nested `using` block around the produce loop + flush suffices), or
2. Use `CreateSibling()` for the consume-phase watchdog, matching `SoakStressTest`.

Option 1 is cleaner since the phases never overlap. Also applies to the Confluent round-trip path if it shares the structure (its single `Track` at line ~326 appears fine, but verify).

## Acceptance

- Round-trip stress job completes and reports phase-separated produce/consume rates.
- Add a unit test that runs a scenario with two sequential phase `Track` registrations against one watchdog to lock in the pattern.

thomhurst:
## Summary
- release the produce-phase progress watchdog before round-trip consumption starts
- cover sequential round-trip phase registrations on one watchdog

## Test plan
- [x] focused `ProgressWatchdogTests` (4/4)
- [x] full net10 unit suite (5,335/5,335)
- [x] real Testcontainers round-trip: 1,000 produced and consumed; all validation counters clean; phase rates reported

Closes #1963

### DIFF
--- a/tests/Dekaf.Tests.Unit/StressTests/ProgressWatchdogTests.cs
+++ b/tests/Dekaf.Tests.Unit/StressTests/ProgressWatchdogTests.cs
@@ -8,6 +8,30 @@ namespace Dekaf.Tests.Unit.StressTests;
 public sealed class ProgressWatchdogTests
 {
     [Test]
+    public void Track_SequentialRoundTripPhases_ReusesWatchdog()
+    {
+        var outputDirectory = CreateOutputDirectory();
+        try
+        {
+            var produceProgress = new ThroughputTracker();
+            var consumeProgress = new ThroughputTracker();
+            using var watchdog = new ProgressWatchdog(outputDirectory);
+
+            using (watchdog.Track(produceProgress, "Dekaf", "producer-roundtrip"))
+            {
+            }
+
+            using (watchdog.Track(consumeProgress, "Dekaf", "producer-roundtrip-consume"))
+            {
+            }
+        }
+        finally
+        {
+            Directory.Delete(outputDirectory, recursive: true);
+        }
+    }
+
+    [Test]
     public async Task Track_Stall_CapturesStacksAndProducerDiagnosticsThenExits()
     {
         var outputDirectory = CreateOutputDirectory();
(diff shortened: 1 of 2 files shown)


thomhurst:
Closing as a duplicate of #1969, which was opened first with the same fix and regression coverage.

PR #1981, opened 2026-07-13
Eliminate consumer Gen2 churn

### LINKED ISSUE #1967: P2 perf: consumer residual Gen2 churn — 13-17 Gen2/15min vs Confluent's 1 (post-#1951), plus minutes 9-13 throughput dip to 0.81x of early-run rate
## Numbers (run [29220954180](https://github.com/thomhurst/Dekaf/actions/runs/29220954180), main @ 888b4ef1, 15min, 1KB msgs)

Dekaf consumers dominate Confluent on every headline metric — throughput 2.12x (2.56M vs 1.21M msg/s), CPU/msg 0.55x, allocated/msg 3B vs 2,440B. Two residuals:

### 1. Gen2 collections still ~16x Confluent

| Scenario | gen0 | gen1 | **gen2** | alloc total |
|---|---|---|---|---|
| consumer (Dekaf) | 39 | 29 | **16** | 6.9GB |
| consumer-batch | 48 | 22 | **17** | 5.3GB |
| consumer-raw | 73 | 19 | **15** | 4.4GB |
| consumer-raw-batch | 173 | 15 | **13** | 2.3GB |
| consumer (Confluent) | 515,838 | 1 | **1** | 2.5TB |

#1951 (in this run) reduced this from 1,211 → ~16, a 75x improvement, but ~1 Gen2/minute remains. With gen0 counts this low, a gen2:gen0 ratio of ~1:3 means these are still LOH/POH-triggered — most likely large fetch-frame or decompression buffers falling outside pool retention limits. Each Gen2 is a pause risk at 2.5–3.8M msg/s.

### 2. Mid-run throughput dip

`consumer` Dekaf minute series (M msg/s): `2.68 2.71 2.68 2.76 2.71 2.65 2.67 2.59 `**`2.18 2.19 2.34 2.43 2.37`**` 2.38 2.88` — minutes 9–13 run ~0.81x of the early-run rate (drift -6.0%, steadyStatePeakRatio 0.864, just above the 0.85 gate). `consumer-raw` shows a shallower version (3.9–4.0 → 3.46 at m12). Worth checking whether dips align with Gen2 timestamps or broker-side log segment rolls before chalking it up to the shared-VM noise band.

## Plan

1. Instrument LOH allocation source under consumer stress (GC ETW `GCAllocationTick` ≥85KB / dotnet-trace gc-verbose) and extend the #1951 pool sizing to cover whatever bucket is still leaking through (suspect: frames above the pool's max retained size, or the per-bucket retention count still below live working set at 3conn fetch depth).
2. Target: ≤2 Gen2 per 15-min run (parity-ish with Confluent) with no throughput loss.
3. While instrumented, correlate the m9–13 dip with GC/broker events; file separately if it's not GC.

Refs: #1942/#1951 (prior fix in this run).

thomhurst:
Closes #1967

## Root cause

Two independent cold-start retention paths forced full collections:

- Every pooled `PendingFetchData` retained its own 1–2 MB parsed `Record[]` slab. Deep prefetch touched roughly 100 instances before reuse, pinning about 186 MiB of slabs.
- High-throughput fetch responses warmed managed 32/64 MiB `byte[]` buckets. The initial LOH allocations accounted for about 388 MiB and dominated Gen2 churn.

## Fix

- Return parsed-record slabs to a dedicated, bounded `ArrayPool<Record>` when each `PendingFetchData` use ends. Clear every rented slab entry before return so record memory/header references are not retained.
- Store response frames at or above the 85,000-byte LOH threshold in bounded, power-of-two native-memory buckets. Small responses keep the existing managed `ArrayPool<byte>` path.
- Preserve contiguous zero-copy parsing, ownership transfer, slicing, direct socket reads, and the existing response-memory wrapper pool.

## Validation

- Full Windows net10 unit suite after review fixes: 5,352/5,352.
- Full Linux net8 unit suite against the netstandard2.0 asset: 5,255/5,255.
- Debug multi-target build (`netstandard2.0`, `net10.0`): 0 warnings, 0 errors.
- Release stress build: 0 warnings, 0 errors.
- Same-host 15-minute consumer stress, 500k seeded messages / 6 partitions:
  - Gen2: **1** (target ≤2; original issue reports 13–17).
  - Throughput: **668,340 msg/s** average, **679,059 msg/s** median.
  - Drift: **-0.8%**; slope: **-0.08%/min** (no minutes 9–13 collapse).
  - Allocated: **106.41 MB** over 601.5M messages.
- Original local 2-minute baseline: 587,556 msg/s, 9 Gen2, 983.39 MB allocated.
- Exact final code on isolated Testcontainers: 0 Gen2 and 13.23 MB allocated over two minutes.

### DIFF
--- a/src/Dekaf/Consumer/KafkaConsumer.cs
+++ b/src/Dekaf/Consumer/KafkaConsumer.cs
@@ -42,23 +42,31 @@ internal sealed class PendingFetchData : IDisposable
     // LockFreeStack avoids the ConcurrentStack node allocation on every return to the pool.
     private const int DefaultMaxPoolSize = 128;
     // Slabs outlive an individual PendingFetchData use, but not the pooled object itself.
-    // Bound retention per size bucket so deep prefetch cannot pin one large slab per item.
-    private const int MaxParsedRecordSlabsPerBucket = 16;
+    // Bound retention per size bucket so deep prefetch cannot pin one large slab per item;
+    // PoolSizing ratchets the depth for high-partition workloads.
+    private const int DefaultMaxParsedRecordSlabsPerBucket = 16;
     private static int s_maxPoolSize = DefaultMaxPoolSize;
     private static LockFreeStack<PendingFetchData> s_pool = new(DefaultMaxPoolSize);
-    private static readonly ArrayPool<Record> s_parsedRecordSlabPool = ArrayPool<Record>.Create(
+    private static int s_maxParsedRecordSlabsPerBucket = DefaultMaxParsedRecordSlabsPerBucket;
+    private static ArrayPool<Record> s_parsedRecordSlabPool = ArrayPool<Record>.Create(
         RecordBatch.MaxReasonableLazyRecordCount,
-        MaxParsedRecordSlabsPerBucket);
+        DefaultMaxParsedRecordSlabsPerBucket);
     private static readonly Lock s_resizeLock = new();
 
     internal static int MaxPoolSizeValue => Volatile.Read(ref s_maxPoolSize);
+    internal static int MaxParsedRecordSlabsPerBucketValue =>
+        Volatile.Read(ref s_maxParsedRecordSlabsPerBucket);
 
-    internal static void RatchetPoolSize(int newSize)
+    internal static void RatchetPoolSize(int newSize) =>
+        RatchetPoolSize(newSize, PoolSizing.ForConsumerParsedRecordSlabs(newSize));
+
+    internal static void RatchetPoolSize(int newSize, int desiredSlabDepth)
     {
         InterlockedHelper.RatchetUp(ref s_maxPoolSize, newSize);
 
         var currentPool = Volatile.Read(ref s_pool);
-        if (currentPool.Capacity < newSize)
+        if (currentPool.Capacity < newSize ||
+            Volatile.Read(ref s_maxParsedRecordSlabsPerBucket) < desiredSlabDepth)
         {
             lock (s_resizeLock)
             {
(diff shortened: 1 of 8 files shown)


thomhurst:
Addressed both blocking reviews in `1729b691`:

- Returned `Record[]` slabs now use `clearArray: true`, clearing the entire physical array; dead retained-slab/count logic was removed.
- Extended `Dispose_ReleasesParsedRecordSlabBeforePooledReuse` to verify the tail entry is cleared.
- Added `ReadFrameAsync_EofMidNativeFrame_ReturnsBufferToPool`, proving native address reuse after partial-frame EOF.
- The Linux net8 suite references Dekaf's `netstandard2.0` asset, so its large scripted-stream test exercises the conditional TLS staging path; the new loopback socket test separately preserves direct native socket fills.

Validation: Windows net10 5,347/5,347; Linux net8/netstandard2.0 5,250/5,250; zero build warnings.

thomhurst:
Addressed latest blocking verdict in `92364044` and `c98b1b54`:

- Parsed-record slab bucket depth now comes from `PoolSizing.ConsumerPoolSizes` and ratchets with the fetch-data working set.
- Retention remains bounded at 16–64 arrays per process-wide size bucket; the validated 64-partition default stays 16, while high-partition workloads scale upward.
- Each rented slab records its owning `ArrayPool<Record>`, so a rare ratchet returns in-flight arrays to the exact pool that rented them.
- Added formula, pathological-overflow, and live ratchet tests.
- The net8 large-stream test now asserts Dekaf's loaded assembly target is `.NETStandard,Version=v2.0` before exercising the conditional staging path.

Validation: Windows net10 5,352/5,352; Linux net8/netstandard2.0 5,255/5,255; zero build warnings.

PR #1995, opened 2026-07-13
perf(producer): cut send-loop estimator and coalesce-spin CPU

### LINKED ISSUE #1990: P1 perf: producer CPU/msg advantage vs Confluent eroded ~40% after 07-13 pipeline commits — wave-coalesce spin re-armed every send + per-acked-request budget recompute
## Evidence

Dekaf/Confluent CPU-per-message ratio (same-runner paired stress runs, lower = Dekaf better):

| Scenario | 07-11 → 07-12 runs | 07-13 06:54 | 07-13 09:09 ([29237954294](https://github.com/thomhurst/Dekaf/actions/runs/29237954294)) |
|---|---|---|---|
| producer FnF 3b | 0.63-0.75 | 0.91 | **1.05** (2.21 vs 2.10 μs/msg — Dekaf now costs MORE) |
| producer-idempotent 3b | 0.55-0.62 | 1.03 | 0.85 |
| producer FnF 1b | 0.56-0.62 | 0.65 | 0.73 |
| producer-acks-all 1b | 0.53-0.64 | 0.72 | 0.93 |

Absolute CPU/msg is noisy across runner VMs (both clients doubled on 07-13 — environmental), but the within-run ratio is controlled, and it stepped up in both 07-13 runs — i.e. after 888b4ef1 (#1960 BBR), 147737e0/ae996d6f (#1972 wave coalescing), 83dabd79 (delivery-age governor) landed. Throughput is still above Confluent; the CPU-efficiency lead is what eroded. Requests are well-formed (~900 msg/req), so this is per-request/per-wave overhead, not fragmentation.

## Root causes (audited via git diff of the 07-12/13 commits)

**1. Wave-coalesce busy-spin re-armed after every send + window widened 2-4x — explains fire-and-forget (which never touches the budget estimator).**
- `ae996d6f`: `BrokerSender.cs:1751-1752` re-arms `waveCoalesceArmed` after *every* successful send (previously only after an idle `WaitToReadAsync`, :1799) and dropped the `_totalPendingResponseCount == 0` guard. A saturated producer never idles, so every wave now enters the spin block (:1344-1397): `Thread.SpinWait(32)` (:1368) + `Stopwatch.GetTimestamp()` per iteration (:1361, :1379).
- `147737e0` widened quiet/max bounds to 1000μs/2000μs (:113-118). `26069b0b` restored 75/500μs only for `LingerMs==0` — the stress harness uses `LingerMs=5`, so it burns the wide window. Up to ~1ms of pause+rdtsc polling per wave × 3 broker send loops.

**2. Full budget recompute moved from once-per-response-pass to once-per-acked-request — explains idempotent/acks-all.**
- `888b4ef1`: `OnAcked` now called inside the per-request loop of `ProcessCompletedResponses` (`BrokerSender.cs:2689-2699`) with a fresh `Stopwatch.GetTimestamp()` per request (:2698). The old per-pass `AckedResponsePass` aggregation was deleted.
- Each `OnAcked` (`BrokerUnackedByteBudget.cs:364-406`) runs: min-RTT update, delivery-latency EWMA, request histograms, ~6 volatile writes, `Interlocked.Exchange` on `_pendingWrittenUnackedPeakBytes` (:399 — cache-line ping-pong vs appender threads), window advance (worst case 20-bucket loop + 2× `Array.Clear` at rollovers, :517-558), occupancy sample, capacity-probe eval, and `RecomputeBudget` which calls `ComputeBudget` 2-3x (:677-701). At the in-flight ceiling one pass drains many requests → ~N× frequency multiplier, ×3 brokers.

**3. Minor:** per-send O(batches) oldest-timestamp scan (`83dabd79`, `BrokerSender.cs:3484-3488`) — grows exactly when coalescing widens requests; per-request `RecordRequestHistograms` (2× Log2 + array increments) is pure diagnostics on the hot path.

Not the cause: `03cde08c` (#1979) removed a ThreadPool hop (CPU win); `d810bc0f` batch sealing is neutral; no new hot-path heap allocations found (`DeliverySnapshot` is a struct).

## Fix plan

1. Don't re-arm the coalesce spin unconditionally after every send: re-arm only when the previous wave coalesced below max width with partitions uncovered, and/or shrink the wide window under saturation; prefer a bounded `eventReader` wait over `SpinWait`+timestamp polling so quiet time yields the core.
2. Keep the cheap per-request BBR rate sample but defer `RecomputeBudget`, the `Interlocked.Exchange` peak-drain, and the timestamp read to once per response pass (published budget only needs freshness at admission).
3. Track oldest `StopwatchCreatedTicks` incrementally during coalescing instead of a second scan at send time; gate histogram recording behind diagnostics-enabled or sample every Nth ack.

Validation: paired stress with CPU/msg ratio back ≤ 0.7x Confluent in 1b and 3b FnF/idempotent without losing the #1972 throughput gains; `[MemoryDiagnoser]`/CPU benchmarks on the accumulator append + response-processing paths before/after.

Prior art: #1960, #1972, #1964 (throughput side, closed), #1988/#1989 (governor behavior).

thomhurst:
Fixes #1990.

## Problem

Two hot-path costs from the 07-13 pipeline work eroded Dekaf's CPU-per-message lead over Confluent (3-broker paired ratio 0.55-0.65x → 0.85-1.05x; fire-and-forget 3b now 2.21 vs 2.10 μs/msg in run 29237954294):

1. **Wave-coalesce spin re-armed after every send** (ae996d6f): on workloads whose requests are effectively single-partition — 99.99% of stress-run requests coalesce at width 1 — every wave burns up to a full 1ms quiet window in `Thread.SpinWait` + `Stopwatch.GetTimestamp()` polling, per broker send loop (~0.3 cores/broker at stress request rates).
2. **Per-acked-request budget recompute** (888b4ef1): `OnAcked` ran the full `RecomputeBudget` (2-3x `ComputeBudget`), the interlocked occupancy-peak drain, and every derived-field publish once per acked request; a hot-poll pass at the in-flight ceiling drains many requests, multiplying that cost by pass depth × broker count.

## Fix

1. **`WaveCoalesceGate`** (new struct): after 8 consecutive fruitless spins the spin is suppressed, with a probe spin every 64 sends so coalescible workloads re-open the gate within one interval. Fruitful spins keep the gate open, preserving the #1972 coalescing wins — the gate keys on the actual coalesce outcome, which static partition-count conditions cannot capture (many partitions + serial batch arrival should not spin; many partitions + co-temporal arrival should).
2. **`OnAcked` / `CompleteAckedPass` split**: `OnAcked` keeps the cheap per-request BBR rate/RTT/latency sampling; the new `CompleteAckedPass`, called once per response pass from `ProcessCompletedResponses` with the last acknowledgement's timestamp, drains the occupancy peak and republishes the budget. The admission gate only reads the budget between passes, so per-request publishing bought no precision.

## Numbers

`BrokerUnackedByteBudgetBenchmarks` (same machine, 0 B allocated):

| | main | this PR |
|---|---|---|
| per acked request | 15.5 ns | 7.9 ns (estimator feed) |
| amortized incl. per-pass publish (5-ack passes) | — | 9.6 ns |

Unit suite: 5359/5359 pass. New `WaveCoalesceGateTests` cover suppress/reprobe/reopen transitions.

## Validation gate (from #1990)

Paired stress run with CPU/msg ratio back ≤ 0.7x Confluent in 1b and 3b FnF/idempotent without losing the #1972 throughput gains — please eyeball the next scheduled stress run's CPU columns and the `coalesceWidthHistogram` before/after.

### DIFF
--- a/src/Dekaf/Producer/BrokerSender.cs
+++ b/src/Dekaf/Producer/BrokerSender.cs
@@ -1070,6 +1070,7 @@ internal sealed partial class BrokerSender : IAsyncDisposable
         var waveCoalesceMaxTicks = MicrosecondsToStopwatchTicks(waveCoalesceBounds.MaximumMicroseconds);
         var waveCoalesceQuietTicks = MicrosecondsToStopwatchTicks(waveCoalesceBounds.QuietMicroseconds);
         var waveCoalesceArmed = true;
+        var waveCoalesceGate = new WaveCoalesceGate();
 
         // Pre-allocate reusable response lookup dictionary to avoid per-response allocation.
         // Single-threaded: only accessed by the send loop, cleared after each use.
@@ -1346,6 +1347,7 @@ internal sealed partial class BrokerSender : IAsyncDisposable
                     && coalescedCount < maxCoalesce
                     && coalescedPartitions.Count < _knownPartitions.Count
                     && carryOver.Count == 0
+                    && waveCoalesceGate.ShouldSpin()
                     && ShouldMicroLinger(
                         coalescedBatches,
                         coalescedCount,
@@ -1353,6 +1355,7 @@ internal sealed partial class BrokerSender : IAsyncDisposable
                 {
                     waveCoalesceArmed = false;
                     _onWaveCoalesceStarted?.Invoke();
+                    var coalescedCountBeforeSpin = coalescedCount;
                     var started = Stopwatch.GetTimestamp();
                     var hardDeadline = started + waveCoalesceMaxTicks;
                     var quietDeadline = Math.Min(started + waveCoalesceQuietTicks, hardDeadline);
@@ -1394,6 +1397,8 @@ internal sealed partial class BrokerSender : IAsyncDisposable
                             break;
                         }
                     }
+
+                    waveCoalesceGate.OnSpinCompleted(coalescedCount > coalescedCountBeforeSpin);
                 }
 
                 // ── 6. Send or wait ──
@@ -1748,8 +1753,12 @@ internal sealed partial class BrokerSender : IAsyncDisposable
 
                 // Keep forming request waves while the pipeline is busy. Without rearming here,
                 // only the first request after an idle period can coalesce adjacent partitions.
+                // The gate suppresses the spin itself once it proves fruitless on this workload.
                 if (sentThisIteration)
+                {
                     waveCoalesceArmed = true;
+                    waveCoalesceGate.OnSent();
+                }
 
                 // ── 7. Compute timeout and wait ──
                 if (transactionEnrollmentReady)
@@ -2608,6 +2617,7 @@ internal sealed partial class BrokerSender : IAsyncDisposable
         // preserving per-partition FIFO ordering during the next coalescing pass.
         // Reverse iteration (swap-with-last O(1) removal) would process R2 before R1,
         // causing the newer batch to be coalesced first and violating ordering.
+        var ackedRequestsThisPass = 0;
         for (var connIdx = 0; connIdx < _pendingResponsesByConnection.Length; connIdx++)
         {
             var pendingList = _pendingResponsesByConnection[connIdx];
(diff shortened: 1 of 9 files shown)


thomhurst:
@claude review

Issue #1991, opened 2026-07-13
P2 ci: stress trend gate fails runs on Confluent-only baseline regressions — control-client drops signal environment, need Dekaf/Confluent ratio corroboration


thomhurst:
## Problem

Run [29237954294](https://github.com/thomhurst/Dekaf/actions/runs/29237954294) failed the `Fail on Stress Regressions` gate largely on environment noise: BOTH clients regressed vs their absolute baselines in nearly every paired scenario (e.g. producer FnF 1b: Dekaf -29%, Confluent -39%; acks-all 3b: Dekaf -19%, Confluent -32%; CPU/msg roughly doubled for both clients in all scenarios on 07-13). Within the run, Dekaf beat Confluent on throughput in every paired scenario (1.11x-3.22x). The failing rows include six Confluent-only "Repeated regression (fail)" entries.

## Root cause

`.github/scripts/stress_trend.py:254`: `should_fail = should_fail or repeated` for baseline metric regressions applies to **all clients including Confluent**. Only the intra-run threshold path excludes the control client (`not is_confluent`, :304). And Dekaf baseline regressions are judged against absolute msg/s / CPU baselines with no corroboration from the paired Confluent control — so a runner-class change (both clients drop proportionally) fails the gate even though the paired design exists precisely to normalize environment.

## Fix plan

1. Treat Confluent baseline regressions as warnings + environment markers, never `should_fail` (mirror the :304 exclusion at :254).
2. For scenarios with a same-run Confluent pairing, corroborate a Dekaf baseline regression with the Dekaf/Confluent ratio: fail only if the ratio also regresses beyond its own trailing band (ratio history is derivable from the existing `stress-history.json` observations). Unpaired Dekaf-only scenarios (consumer-batch/raw/raw-batch) keep current behavior.
3. When Confluent regresses beyond band, annotate the run "environment shift suspected" so baseline medians aren't silently poisoned.

This preserves the gate's sensitivity to real Dekaf regressions (a Dekaf-only drop still fails via ratio) while eliminating the recurring cross-VM-noise failures (per the ±30% runner variance documented in #1585 investigations).

Constraint: no new paid-runner jobs — this is purely a scripting change.

PR #1996, opened 2026-07-13
ci(stress): warn on Confluent regressions

thomhurst:
## What changed

- make repeated historical throughput and CPU regressions failure-eligible only for Dekaf results
- keep Confluent regressions in trend history and report them as warnings
- align intra-run eligibility with same Dekaf-only policy
- add regression coverage for consecutive Confluent throughput and CPU regressions

## Why

Confluent.Kafka is comparison baseline. Its performance degradation should not fail Dekaf stress regression gate. Dekaf performance regressions and Dekaf-to-Confluent latency thresholds remain gating; correctness failures still invalidate stress run.

## Validation

- `python -m unittest discover -s .github/scripts -p 'test_stress_*.py' -v` (42 passed)
- `git diff --check`


thomhurst:
Superseded by #1998, which covers the full #1991 scope: Confluent warning-only behavior, paired Dekaf/Confluent ratio corroboration, environment-shift marking/baseline exclusion, and warning-level annotation tests. Closing this narrower duplicate.

PR #1998, opened 2026-07-13
Corroborate stress regressions with Confluent control

### LINKED ISSUE #1991: P2 ci: stress trend gate fails runs on Confluent-only baseline regressions — control-client drops signal environment, need Dekaf/Confluent ratio corroboration

thomhurst:
## Summary
- treat Confluent baseline regressions as environment warnings, never gate failures
- require paired Dekaf regressions to be corroborated by a trailing Dekaf/Confluent ratio band
- mark suspected environment-shift runs and exclude them from absolute baselines
- preserve existing consecutive-regression behavior for unpaired scenarios

## Test plan
- [x] `python -m unittest test_stress_trend.py`
- [x] `python -m unittest discover -s .github/scripts -p 'test_stress_*.py'`
- [x] `python -m compileall -q .github/scripts`
- [x] replay stress run 29237954294 against committed history

Closes #1991

### DIFF
--- a/.github/scripts/stress_trend.py
+++ b/.github/scripts/stress_trend.py
@@ -132,10 +132,13 @@ def _matching_control_ratios(runs, result, metric):
     ratios = []
 
     for run in runs:
+        if run.get("environmentShiftSuspected", False):
+            continue
         matching_results = [
             item
             for item in run.get("results", [])
             if _pair_identity(item) == pair_key
+            and not item.get("environmentShiftSuspected", False)
         ]
         candidate = next(
             (item for item in matching_results if _identity(item) == candidate_key),
--- a/.github/scripts/stress_trend.py
+++ b/.github/scripts/stress_trend.py
@@ -92,6 +92,26 @@ def _identity(result):
     )
 
 
+def _is_confluent(result):
+    return str(result.get("client", "")).casefold().startswith("confluent")
+
+
+def _pair_identity(result):
+    """Match a Dekaf scenario to its same-run Confluent control.
+
+    Client and consumer connection count are intentionally excluded: each client uses its
+    own connection preset, while broker count and workload shape define the paired run.
+    """
+    return (
+        str(result.get("scenario", "unknown")).casefold(),
+        result.get("brokerCount", 1),
+        result.get("durationMinutes"),
+        result.get("messageSizeBytes"),
+        _consumer_seed_batch_size(result),
+        _roundtrip_messages(result),
+    )
+
+
 def _matching_observations(runs, result):
     key = _identity(result)
     matches = []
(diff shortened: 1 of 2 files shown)


thomhurst:
Confirmed: environment-shift exclusion is deliberately run-wide because Confluent is the runner/host control, not only a scenario control. Commit b9c27947 documents that invariant and adds a regression test proving one control regression marks every same-run observation. I left Markdown status wording and annotation severity/prefix classification separate: they are distinct output contracts, and consolidating them would broaden this focused gate fix.

thomhurst:
Addressed the ratio-baseline finding in commit `b6d01c49fa9bf684eb0312657e5d6f8a700fab7a`: run-level and observation-level environment shifts are now excluded from _matching_control_ratios; regression coverage verifies a skewed 5.0 ratio does not enter the clean 2.0 baseline. Python suite: 32/32; compileall: clean.

PR #1999, opened 2026-07-13
Add consumer fetch diagnostics timeline

### LINKED ISSUE #1992: P2 perf: consumer throughput steps down at minutes 6 and 9 and never recovers (steady/peak 0.77, -2.0%/min) — recurs across runs; Confluent decays too, needs client-vs-broker attribution
## Evidence (stress run [29237954294](https://github.com/thomhurst/Dekaf/actions/runs/29237954294), commit 98ffdc6a)

Consumer 1b per-minute throughput (k msg/s):
- Dekaf: `2623 2680 2698 2713 2792 2653 2322 2321 2404 2152 2033 2195 2134 2102 2068` — **-17% step at minute 6** (2792→2322), second step at minute 9 (2404→2152), never recovers. drift -20.4%, slope -2.00%/min, steady/peak 0.77.
- Confluent same job: `944 1054 1083 1107 1007 916 936 887 978 856 789 753 896 873 980` — also decays (-16%, slope -1.29) but gradually, partially recovering late.

History: Dekaf consumer steady/peak across runs = 0.80, 0.89, 0.86, 0.86, 0.93, **0.77** — the dip recurs (was the "minutes 9-13 dip to 0.81x" half of #1967, whose Gen2 half is fixed: Gen2 now 2/run post-#1981 ✅).

## Analysis

Both clients decaying in the same job points to a shared broker/host component (page-cache behavior as the seeded log ages, tmpfs memory pressure, cgroup throttling on the pinned cores). But the Dekaf-specific signatures need ruling out:

1. **Sharp step at minute 6 + minute 9** vs Confluent's gradual slope. Minute-9 steps have previously been idle-reaper artifacts (#1910, #1945, fixed by #1956 for producers) — verify the consumer connection path has no equivalent idle-reap/rebuffer trigger at ~540s.
2. Dekaf drives ~2.5x Confluent's fetch byte rate — if the broker-side effect is load-proportional (page-cache miss rate), Dekaf hits it harder; that would still count against us on the gate.

## Fix plan

1. Add consumer-side time-series diagnostics to stress output mirroring `brokerBudgetSamples`: fetch request rate, bytes/fetch, fetch RTT, queue depths, per-minute — so the step can be attributed (client stall vs broker slowdown).
2. Check for time-keyed client behavior at 360s/540s (connection reap, buffer pool trim, LOH compaction) in the consumer path.
3. If broker-side confirmed: adjust seeded-topic size/retention so the 15-minute window doesn't cross a broker state change, or normalize the gate on the paired Confluent control (see #1991).

Prior art: #1967 (Gen2 half fixed by #1981), #1910/#1945/#1956 (producer idle-reap steps), #1942/#1951 (LOH).

thomhurst:
## Summary
- capture minute-scale consumer fetch request rate, payload bytes per fetch, average RTT, and queue depths
- persist connection-reap events in consumer diagnostics and correlate samples with throughput
- render the timeline in stress JSON and Markdown for every Dekaf consumer scenario

## Test plan
- [x] `dotnet build tests/Dekaf.Tests.Unit/Dekaf.Tests.Unit.csproj --configuration Release --framework net10.0`
- [x] `Dekaf.Tests.Unit.exe` (5,363 passed)
- [x] `dotnet build src/Dekaf/Dekaf.csproj --configuration Release --framework netstandard2.0`
- [x] `dotnet build tools/Dekaf.StressTests/Dekaf.StressTests.csproj --configuration Release`
- [x] 1-minute real-Kafka consumer stress smoke (10.2M messages; populated fetch timeline; zero errors)

Closes #1992

### DIFF
--- a/src/Dekaf/Consumer/ConsumerDiagnosticSnapshot.cs
+++ b/src/Dekaf/Consumer/ConsumerDiagnosticSnapshot.cs
@@ -1,5 +1,7 @@
 namespace Dekaf.Consumer;
 
+using Dekaf.Networking;
+
 internal sealed class ConsumerDiagnosticSnapshot
 {
     public required DateTimeOffset CapturedAtUtc { get; init; }
@@ -18,6 +20,7 @@ internal sealed class ConsumerDiagnosticSnapshot
     public required ConsumerPartitionEpochDiagnostic[] MinimumFetchBufferEpochsByPartition { get; init; }
     public required int? AdaptivePartitionFetchBytes { get; init; }
     public required int? AdaptiveFetchMaxBytes { get; init; }
+    public ConnectionReapDiagnostic[] ConnectionReapEvents { get; init; } = [];
 }
 
 internal sealed record ConsumerTopicPartitionDiagnostic(string Topic, int Partition);
--- a/src/Dekaf/Consumer/KafkaConsumer.cs
+++ b/src/Dekaf/Consumer/KafkaConsumer.cs
@@ -1367,8 +1367,8 @@ public sealed partial class KafkaConsumer<TKey, TValue> :
             MinimumFetchBufferEpochsByPartition = minimumEpochs,
             AdaptivePartitionFetchBytes = _adaptiveFetchSizer?.CurrentPartitionFetchBytes,
             AdaptiveFetchMaxBytes = _adaptiveFetchSizer?.CurrentFetchMaxBytes,
-            ConnectionReapEvents = _connectionPool is ConnectionPool connectionPool
-                ? [.. connectionPool.GetConnectionReapDiagnosticsSnapshot()]
+            ConnectionReapEvents = _connectionPool is IConnectionPoolDiagnostics diagnostics
+                ? [.. diagnostics.GetConnectionReapDiagnosticsSnapshot()]
                 : []
         };
     }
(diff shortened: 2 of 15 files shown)


thomhurst:
Addressed review follow-ups in commit `ecc1eaa2869d5c48ad120a13127a7aab1383501e`: consumer fetch Markdown now samples at most 100 rows across the full timeline with an omission note; byte formatting now shares one binary-unit formatter and reports KiB/MiB/GiB consistently. Focused tests: 5/5; Release build: clean.

thomhurst:
Addressed the blocking re-review and periodic-loop concern in commit `f7930049`:

- preserved GC table `KB/MB/GB` convention while consumer diagnostics retain `KiB/MiB/GiB` through one shared formatter;
- added internal `IConnectionPoolDiagnostics`; producer and consumer now read reap snapshots through the capability interface;
- centralized best-effort `TryTakeSample`, so final diagnostics cannot abort any consumer scenario;
- consolidated periodic sampling loops in `StressTestHelpers.RunPeriodicAsync`;
- enforced one active metrics listener, preventing process-wide fetch RTT from being silently misattributed across concurrent trackers.

Timeline generators stay output-specific; the 100-row cap bounds nearest-sample/reap scans, so a sweep refactor adds complexity without material runtime benefit.

Verification: focused 8/8; full net10 unit suite 5,367/5,367; Release build and netstandard2.0 core build clean.

PR #2001, opened 2026-07-13
Stabilize producer budget recovery

thomhurst:
Closes #1989

## Summary

- floor loaded budgets on a serving-RTT EWMA instead of base RTT alone
- retain loaded delivery-rate peaks with a 60-second half-life, while clearing the retained floor after true application idle
- cap capacity-probe cadence at one second, evaluate three-RTT averages, and use stronger headroom when RTT-floor-bound
- retain one BDP during minimum-RTT probes instead of emptying the pipe
- rate-limit budget decay to 10%/second while new admission blocks show active gate pressure

Dense-affinity connection widening is intentionally excluded: each partition is deterministically assigned to one connection and width is capped by known partitions, so extra lanes would be idle or risk ordering changes. Recovery stays in the budget controller.

#1997 changes the same controller's latency-governor policy. This PR keeps that independent scope separate and may require a mechanical rebase after #1997 lands.

## Validation

- net10 unit suite: 5,367 passed
- net8 unit suite: 5,269 passed; one unrelated 30-second scheduling timeout passed alone in 302 ms
- focused budget suite: 52 passed
- BenchmarkDotNet: 10.29 ns acknowledgement, 13.00 ns acknowledgement pass, zero allocations

PR #1997, opened 2026-07-13
perf(producer): make latency target a cap

thomhurst:
## Summary
- make latency-budget growth one-sided and require observed admission pressure
- control seal-to-send queue latency instead of broker RTT, using rate/RTT-derived floors
- cap wave-coalesce quiet/max waits to 5%/10% of the delivery-latency target
- cover controller, floor, coalescing, and diagnostic behavior; update hot-path benchmark inputs

## Test plan
- [x] `dotnet build tests/Dekaf.Tests.Unit --configuration Release`
- [x] `tests/Dekaf.Tests.Unit/bin/Release/net10.0/Dekaf.Tests.Unit.exe` (5,362 passed)
- [x] `dotnet build tests/Dekaf.Tests.Integration --configuration Release`
- [x] `tests/Dekaf.Tests.Integration/bin/Release/net10.0/Dekaf.Tests.Integration.exe --treenode-filter /*/*/ProducerTests/*` (27 passed)
- [x] `dotnet run --project tools/Dekaf.Benchmarks --configuration Release -- --filter *BrokerUnackedByteBudgetBenchmarks*` (9.39 ns ack, 10.65 ns/pass, zero allocations)

Closes #1988

PR #2002, opened 2026-07-13
perf(producer): eliminate transactional continuation dispatch

### LINKED ISSUE #1993: P2 perf: transactional CPU/msg 1.44x Confluent (382 vs 266us) after #1979 — per-request overhead dominates at 1 msg/request
## Evidence (stress run [29237954294](https://github.com/thomhurst/Dekaf/actions/runs/29237954294), commit 98ffdc6a)

Producer Transactional EOS, 3 brokers, paired:

| Client | msg/s | CPU μs/msg | Cores used |
|---|---|---|---|
| Dekaf | 411 (3.22x ✅) | **382.26** | 0.21 |
| Confluent | 127 | 265.92 | 0.05 |

Throughput lead improved (2.4x → 3.2x, trend "improvement") and CPU/msg improved vs own baseline (400 → 382), but the ratio vs Confluent regressed from 1.32x (#1968, closed by #1979) to 1.44x. Dekaf burns 4x the cores for 3.2x the throughput — per-message CPU is the last dimension where Confluent wins this scenario.

Delivery diagnostics show the shape: 547 req/s at **1 msg/request** (coalesce width 1 across all 492,900 requests) — the txn scenario commits per message, so all per-request and per-txn-cycle overhead lands on each message.

## Where to look

Remaining candidates from the #1914/#1968 audits plus the 07-13 additions:
1. Per-acked-request `OnAcked` budget recompute (see #1990 item 2) — at 1 msg/req this is per-message.
2. Per-txn-cycle: EndTxn/AddPartitionsToTxn round-trips and scheduler transitions (#1921 KIP-890 implicit enrollment, #1922-#1925 filed earlier — verify which landed).
3. Wave-coalesce spin (#1990 item 1) — the txn send loop enters it per wave at width 1.

## Fix plan

1. Land #1990's per-pass budget recompute (direct win at 1 msg/req).
2. Adopt KIP-890 implicit partition enrollment on TV2 brokers (#1921) to drop the AddPartitionsToTxn round-trip per cycle.
3. Profile a 15s transactional stress slice with dotnet-trace and rank remaining per-cycle costs (commit path scheduler hops, per-message async state).

Validation gate: paired txn stress with CPU/msg ratio ≤ 1.0x Confluent while keeping ≥3x throughput.

Prior art: #1914, #1968 (closed by #1979), #1921-#1925.

thomhurst:
## Summary
- resume sequential transactional `ProduceAsync` continuations inline on the broker sender thread
- restore pooled completion sources to asynchronous continuation mode before reuse
- cover inline completion, pooled reuse, and sequential producer reentrancy
- add a focused continuation scheduling benchmark

## Performance
- paired 3-broker transactional stress: 842.93 us/msg Dekaf vs 889.42 us/msg Confluent (0.95x)
- median throughput: 63 vs 20 msg/s (3.18x)
- microbenchmark: 915.5 ns async vs 102.3 ns inline (0.11x), both 128 B
- transactional verification: zero loss, duplicates, aborted leaks, or unexpected records

## Validation
- `dotnet run --project tests/Dekaf.Tests.Unit -c Release -f net10.0 --no-build` (5,362 passed)
- `TransactionTests` integration filter (5 passed)
- `TransactionV2Tests` integration filter (3 passed)
- paired 1-minute `producer-transactional`, all clients, 3 brokers

Closes #1993

### DIFF
--- a/src/Dekaf/Diagnostics/DekafMetrics.cs
+++ b/src/Dekaf/Diagnostics/DekafMetrics.cs
@@ -40,6 +40,12 @@ internal static class DekafMetrics
             unit: "{retry}",
             description: "Number of produce retries.");
 
+    internal static readonly Counter<long> InlineContinuationExceptions =
+        DekafDiagnostics.Meter.CreateCounter<long>(
+            "dekaf.producer.inline_continuation.exceptions",
+            unit: "{exception}",
+            description: "Exceptions isolated from inline transactional produce continuations.");
+
     // Consumer metrics
     internal static readonly Counter<long> MessagesReceived =
         DekafDiagnostics.Meter.CreateCounter<long>(
--- a/src/Dekaf/Producer/IKafkaProducer.cs
+++ b/src/Dekaf/Producer/IKafkaProducer.cs
@@ -190,6 +190,12 @@ public interface IKafkaProducer<TKey, TValue> : IInitializableKafkaClient, IAsyn
     /// <summary>
     /// Begins a transaction (if transactional).
     /// </summary>
+    /// <remarks>
+    /// Transactional produce continuations run inline on a broker sender thread shared by
+    /// other partitions using that broker connection. Code after each transactional
+    /// <see cref="ITransaction{TKey, TValue}.ProduceAsync"/> call must remain non-blocking
+    /// and avoid long-running synchronous work.
+    /// </remarks>
     ITransaction<TKey, TValue> BeginTransaction();
 
     /// <summary>
@@ -469,6 +475,12 @@ public interface ITransaction<TKey, TValue> : IAsyncDisposable
     /// <summary>
     /// Produces a message within the transaction.
     /// </summary>
+    /// <remarks>
+    /// The continuation after awaiting this method runs inline on a broker sender thread
+    /// shared by other partitions using that broker connection. Keep code between this
+    /// call and the next transactional operation non-blocking and computationally cheap;
+    /// synchronous waits or long-running work delay broker-wide send and response processing.
+    /// </remarks>
     ValueTask<RecordMetadata> ProduceAsync(
         ProducerMessage<TKey, TValue> message,
         CancellationToken cancellationToken = default);
--- a/src/Dekaf/Producer/KafkaProducer.cs
+++ b/src/Dekaf/Producer/KafkaProducer.cs
@@ -970,9 +970,7 @@ public sealed partial class KafkaProducer<TKey, TValue> : IKafkaProducer<TKey, T
         }
 
         // All checks passed - we can proceed synchronously
-        completion = _valueTaskSourcePool.Rent();
-        if (!runContinuationsAsynchronously)
-            completion.SetRunContinuationsAsynchronously(false);
+        completion = RentCompletion(runContinuationsAsynchronously);
         var result = SyncProduceResult.Success;
         try
         {
@@ -1027,9 +1025,7 @@ public sealed partial class KafkaProducer<TKey, TValue> : IKafkaProducer<TKey, T
         }
 
         // Topic cache miss — fetch topic metadata inline and produce
-        var completion = _valueTaskSourcePool.Rent();
-        if (!runContinuationsAsynchronously)
-            completion.SetRunContinuationsAsynchronously(false);
+        var completion = RentCompletion(runContinuationsAsynchronously);
         try
         {
             await ProduceInternalAsync(message, completion, cancellationToken).ConfigureAwait(false);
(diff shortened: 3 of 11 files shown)


thomhurst:
Requested the single permitted CI rerun. `NonIdempotentProducer_MultipleInFlight_PipelinesSamePartitionBatches` timed out at 30s; this PR only changes the transactional produce path, while the non-transactional public path retains asynchronous continuations. The exact test passed 5/5 locally (0.36–0.45s each), and the full local unit suite previously passed 5,362/5,362.

thomhurst:
Addressed both blocking points in `4d71ad91`.

1. Ordinary produce no longer writes continuation mode at all; only transactional rents opt into `false`. The timed-out test directly constructs `BrokerSender`/`RecordAccumulator` and never constructs `KafkaProducer` or calls the new setter. Its only overlap is the pooled source's idempotent reset back to its existing `true` default after `GetResult`; that occurs after completion and cannot gate either `PipelinedSendLogger` signal. Exact test now passes 6/6 locally (latest 374 ms); full unit suite remains 5,362/5,362.
2. The real-broker transaction test is now an explicit, timeout-bounded `Transaction_ProduceThenCommit_InlineContinuationDoesNotDeadlock` regression. Its comment documents the load-bearing async flush boundary; the Transaction integration class passes 5/5.

thomhurst:
Addressed the latest review in `a05279e5`:

- ReadyBatch now isolates raw inline-continuation exceptions per source on both success and failure paths, so later completions and DoneTask cannot be stranded.
- Added a regression with two inline sources where the first continuation throws; the second source and DoneTask still complete.
- Added route-level coverage for transactional fast-path BufferFull: verifies the returned pooled source resumes asynchronously on its next rental.
- Factored completion rental into one aggressively-inlined helper.

Verification:
- Release build: 0 warnings/errors
- PooledValueTaskSourceTests: 42/42
- KafkaProducerFastPathTests: 9/9
- Full unit suite: 5364/5364
- TransactionTests integration: 5/5

thomhurst:
Addressed the remaining completion-isolation gaps in `177198a4`:

- Centralized safe pooled completion for result, exception, and cancellation.
- Applied it to PartitionBatch sealing failure, ReadyBatch orphan reset, append-worker failure/cancellation, enqueue rejection, and disposal channel drains.
- Expanded the regression to cover both successful and failed ReadyBatch completion when the first raw inline continuation throws.

Verification:
- Release unit/integration builds: 0 warnings/errors
- Focused regression: 2/2
- Full unit suite: 5365/5365
- TransactionTests integration: 5/5

thomhurst:
Addressed the API-surface warning in `b6466bff`:

- `BeginTransaction` and `ITransaction.ProduceAsync` now explicitly document that transactional continuations resume inline on the broker sender thread shared by other partitions, and require non-blocking, cheap synchronous work.
- Added a DEBUG-only counter for isolated raw inline-continuation exceptions, avoiding a silent diagnostic black hole with zero Release hot-path cost.

Verification: clean Release build; full unit suite 5365/5365.

thomhurst:
Addressed the carried Release-observability concern in `62f36dc3`.

Isolated raw inline-continuation exceptions now increment `dekaf.producer.inline_continuation.exceptions` through the standard `Dekaf` meter. The counter is touched only on the exceptional catch path, so normal message completion is unchanged; the existing DEBUG counter remains for richer local diagnostics. Consolidated result/exception/cancellation instrumentation through one helper.

Verification: Release build clean; metric registration/emission 2/2; `PooledValueTaskSourceTests` 43/43; full net10 unit suite 5,366/5,366.

thomhurst:
CI diagnosis for current head \62f36dc3\: \Unit Tests (net8.0)\ timed out only in \InFlightWait_MutedPartitionFilter_MovesCoalescedNormalBatchToCarryOver\ after 30s; 5,272/5,273 tests passed. This existing BrokerSender ordering test is outside the PR diff. Its exact net8 Release case passes 20/20 locally on this head (~0.9s/run), and net10 full suite already passed 5,366/5,366 locally. The PR's single allowed CI rerun was already consumed on the earlier-head timeout, so I am not rerunning again or adding an unrelated speculative change.

thomhurst:
Rebased cleanly onto current `main` after #1997 and #2001 in `53b391b6`; no conflicts.

Post-rebase verification:
- Release unit/integration builds: 0 warnings/errors
- Full net10 unit suite: 5391/5391
- Prior failing net8 test `InFlightWait_MutedPartitionFilter_MovesCoalescedNormalBatchToCarryOver`: 1/1 in 354ms
- TransactionTests integration: 5/5

This base refresh supersedes the unrelated net8 timeout run and validates the combined producer-controller state.

PR #2019, opened 2026-07-13
fix(producer): stabilize unacked-budget control

### LINKED ISSUE #2008: P0 perf: 1-broker producer throughput halved, acks-all now 0.69x Confluent — #1997 budget is feed-forward R̂×horizon on self-measured rate with no upward force
## Summary

Stress run [29254448922](https://github.com/thomhurst/Dekaf/actions/runs/29254448922) (commit e4b032c2): 1-broker producer throughput halved across every variant, and Dekaf now **loses to Confluent** on acks-all and idempotent:

| Scenario (1 broker) | Previous (~run 29237954294) | This run | Dekaf/Confluent ratio |
|---|---|---|---|
| Fire-and-forget | ~1.61M msg/s | 837k | 1.33 → 1.06 |
| Acks-all | ~1.64M msg/s | 836k | 1.29 → **0.69x (loss)** |
| Idempotent | ~1.65M msg/s | 740k | 1.39 → **0.76x (loss)** |
| Idempotent (3conn) | ~1.63M msg/s | 856k | 1.38 → **0.88x (loss)** |

The collapse is **scenario-independent** — FnF 837k / acks-all 836k / idempotent 740k all land at the same ~830k clamp. Signature: buffer utilization 2-4% the whole run, pressure `0/100`, Dekaf using only 0.90-0.95 cores (Confluent: 1.68-1.74). Delivery p95 is 0.74x of the 10ms target — the producer is self-throttled, not resource- or backpressure-bound. Round-trip throughput also fell 204k → 118k msg/s from the same mechanism (its 118 MB/s measured rate yields a ~1.2 MB budget ≈ one 1 MiB batch → near depth-1 pipelining).

Introduced by 5b9c1c75 (#1997, "make latency target a cap", fixing #1988).

## Root cause

The 10ms `DeliveryLatencyTargetMs` became a **feed-forward admission ceiling fed by a self-measured drain rate**, with every unconditional upward force removed:

1. **Budget = R̂ × horizon exactly.** `ComputeBudget` (`src/Dekaf/Producer/BrokerUnackedByteBudget.cs:900`) clamps the occupancy floor to `1.0 × rateBudgetBytes` (was `1.5×` via the deleted `OccupancyFloorHeadroom`). The 1.5× headroom was the escape valve that let occupancy demonstrably above the rate estimate push the budget toward broker capacity.
2. **Budget floor removed.** `CreateBrokerUnackedBudget` (`src/Dekaf/Producer/RecordAccumulator.cs:4556-4559`) now passes `floorBytes: 1` (was `max(2 × BatchSize, 1 MiB)`). Nothing bounds collapse.
3. **The latency controller is blind.** `UpdateDeliveryLatency` (`BrokerUnackedByteBudget.cs:459-505`) now measures **seal→send** only (was created→ack). Seal→send is ≈0 in healthy operation, so `_latencyBudgetScale` sits inert; latency "control" is purely open-loop via the horizon.

**The circularity:** R̂ (`_windowMaxRate` / `_retainedLoadedMaxRate`) is measured from acks of traffic the gate itself admitted (`OnAcked`, `BrokerUnackedByteBudget.cs:382-439`). Delivered rate can never exceed gate-admitted throughput, so `budget = R̂ × 10ms` is self-confirming at **any** throughput level — a neutral equilibrium with no tether to broker capacity. The only upward mechanism is the capacity probe, which is structurally broken (separate issue). Cold start uses `budget = cap` (32 MB) — that's why the first minute hits ~1.2M msg/s before the first estimator windows collapse the budget and the run never recovers.

The admission gate sits upstream of memory reservation (`TryAdmitAndReserve` → `IsBrokerAdmissionBlocked`, `RecordAccumulator.cs:4567-4577`), which is why BufferMemory backpressure never engages and both caller and send loop sleep at <1 core.

Acks-all ratio is worst (0.69x) simply because Confluent's acks-all baseline (~1.16M) is higher than its FnF baseline — no acks-specific Dekaf mechanism; a producer-side byte ceiling doesn't care about ack mode on RF=1.

## Fix plan

1. **Primary: make the target a feedback setpoint on measured end-to-end latency, not a feed-forward multiplier on self-measured rate.** MIMD on the latency error: while admission-block events are fresh **and** measured append/seal→ack latency EWMA is below target, grow the budget multiplicatively (e.g. ×1.05 per 100ms control interval) regardless of probe outcomes; shrink (e.g. ×0.75) only when measured latency exceeds target. Growth stops exactly when the latency budget is consumed — the behavior Confluent exhibits (1.2M msg/s and fine latency on the same box). Requires re-feeding a real end-to-end latency sample into the controller (revert/augment the seal→send redefinition at `BrokerUnackedByteBudget.cs:470`).
2. **Restore a meaningful floor**: `floorBytes` back to `max(2 × BatchSize, 1 MiB)`, or a BDP-derived floor `R̂ × 1.5 × servingRtt` that holds even when `_retainedLoadedMaxRate == 0`.
3. Stopgap if (1) is deferred: reinstate bounded occupancy headroom (1.25-1.5×) at `BrokerUnackedByteBudget.cs:900`.
4. The scale governor must distinguish "queue caused by over-admission" (shrinking helps) from "queue caused by a CPU-bound send loop" (shrinking only starves the pipe): only shrink when written-unacked occupancy is at budget.

## Acceptance gate — do not repeat the inversion

This is the third swing of the same pendulum (#1843 → #1988 → this). Any fix PR must be validated by a stress run asserting **both**:
- 1-broker FnF/acks-all/idempotent ≥ baseline band (≥1.4M msg/s FnF), and
- 3-broker delivery p95 ≤ 3× target and 1-broker p95 ≤ target band.

Fixing one side while regressing the other is not a fix.

Related: #1988 (the setpoint issue this inverted), #1843 (previous inversion), companion issues for the 3-broker latency loop, estimator decay/probe machinery, and coalesce fragmentation filed from the same run.

### LINKED ISSUE #2009: P0 perf: 3-broker delivery latency 26x target (acks-all p50 40ms / p95 265ms) — servingRttEwma feedback loop (gain 1.5) ratchets budget to 32MiB cap; seal→send sensor blind to queue
## Summary

Stress run [29254448922](https://github.com/thomhurst/Dekaf/actions/runs/29254448922) (commit e4b032c2): 3-broker delivery latency blew through the 10ms target while 1-broker latency is fine — the #1997 "cap" does not cap on multi-broker:

| Scenario (3 brokers) | Dekaf p50 / p95 / p99 | Confluent p50 / p95 / p99 | Gate |
|---|---|---|---|
| Acks-all | **40.4ms / 264.6ms / 338.4ms** | 8.3 / 23.3 / 46.1ms | p95 = 26.45× target (fail) |
| Idempotent | 17.9 / 150.9 / 239.2ms | 6.9 / 17.6 / 33.7ms | p95 = 15.09× target (fail) |
| Idempotent (3conn) | 17.3 / 128.5 / 195.4ms | — | p95 = 12.85× target (fail) |
| Fire-and-forget | 5.6 / 45.6 / 89.9ms | 5.0 / 11.1 / 19.3ms | p95 = 4.55× target (fail) |

1-broker for comparison: Dekaf p50 1.65ms / p95 7.45ms — under target. The elevated **p50** (4.89× Confluent on acks-all) is Little's law from a permanent standing queue, not stalls: throughput ratios stay ≥1.0 while every message transits ~32 MiB of queued bytes.

## Root cause: gain-1.5 positive feedback loop in the RTT floor, with the latency sensor blinded

Delta 98ffdc6a → e4b032c2 (#1995/#1997/#2001) redefined the budget machinery:

1. The budget horizon is `max(target × scale, 1.5 × max(minRtt, servingRttEwma))` (`src/Dekaf/Producer/BrokerUnackedByteBudget.cs:866-876`).
2. `_servingRttEwmaSeconds` is an EWMA of **loaded** (non-app-limited) request RTTs (`BrokerUnackedByteBudget.cs:401-407`) — i.e. it includes the queueing delay the budget itself created, and is deliberately never pulled toward the queue-drained probe RTT (:395-399).
3. Under saturation, send→ack RTT ≈ serviceRtt + unackedBytes/drainRate. Substituting into the horizon: `B′ = 1.5·rate·serviceRtt + 1.5·B` — **loop gain 1.5 > 1, no finite fixed point**. The budget ratchets geometrically until clamped at `_capBytes` = BatchSize × 32 × connections = **32 MiB/broker** (`ComputeCap`, :197-198; `RecordAccumulator.cs:4551-4552`).

Three amplifiers make the ratchet stick:
- Probe acceptance is `avgRate > baseline × 1.01` (:696-705) — delivered rate through a saturated broker never drops when queue is added, so probes essentially never fail under load; each success re-baselines. The probe multiplier is 1.5× **exactly when the RTT floor dominates** (`RttFloorProbeBudgetMultiplier`, :124, :878-884) — an accelerant in precisely the pathological regime.
- Decay is capped at 10%/s while admission is blocked (`BoundBudgetDecay`, :832-848), and under saturation admission is always blocked — up-moves instant, down-moves 10× slower.
- **The sensor that should catch this is blind**: `UpdateDeliveryLatency` (:459-475) measures seal→send only (fed pre-socket-write, `BrokerSender.cs:3524-3529`). Seal→send stays ≪10ms at any budget, so the scale governor actively grows back toward 1.0 (:491-503). Nothing measures the 264ms between socket write and ack.

**Ignition and the 1b/3b split:** the loop ignites when `1.5 × servingRtt > target` (6.7ms at 10ms target). 3-broker RF=3 acks=all replication service time is several ms unloaded → ignites immediately → rides the cap. Sanity check: 32 MiB ÷ per-broker drain (~130-200 MB/s) = ~160-250ms standing latency → matches observed p95 264ms (acks-all) / 151ms (idempotent). FnF (acks=1) has lower service time → equilibrium below cap → p95 ~45ms. 1-broker RF=1 has sub-ms service RTT and a CPU-bound send loop → no standing queue forms → floor never binds.

Secondary 3b contributor (FnF ratio 1.32→1.03): coalesce windows clamped to target/20 and target/10 (`BrokerSender.cs:2235-2253`) + `WaveCoalesceGate` spin suppression (`src/Dekaf/Producer/WaveCoalesceGate.cs:19-31`) produce smaller requests on the sparser per-broker arrival pattern → more broker CPU per byte → lower drain rate (covered in companion coalesce issue).

### Gen2 side effect

3-broker producer runs show 16-26 Gen2 collections vs 5-6 on 1-broker. At the ratcheted cap, ~96 sealed-unacked 1 MiB pinned `BatchArena` buffers stand in flight (3 × 32 MiB) — above the 128-slot static pool once pooled-batch retention is added. Budget oscillations (min-RTT probes drain the queue 50-250ms; capacity probes swing ±25-50%) thrash the pool: misses allocate fresh POH 1 MiB arrays (`RecordAccumulator.cs:742,763-764`), full-pool returns drop buffers (:781-787), and POH garbage is only reclaimed by Gen2. (Hardening tracked separately; fixing the ratchet removes the cause.)

## Fix plan

1. **Anchor the RTT floor to queue-drained service RTT, not loaded RTT.** The min-RTT probe already measures RTT at ~zero queue depth, and on acks=all it includes replication service time — exactly the BDP horizon needed to keep a replicated pipe full. Replace `max(minRtt, servingRttEwma)` (:867-869) with an EWMA of probe-window RTTs, or clamp `servingRttEwma ≤ k × probeRtt`. Deletes the gain-1.5 loop while preserving the "replicated service time keeps the pipe full" intent. If `1.5 × serviceRtt > target`, budget = BDP (a few MB) and the infeasible target is surfaced via a diagnostic gauge instead of silently riding the 32 MiB cap.
2. **Restore an end-to-end latency sensor** without recreating the #1988 setpoint: control on standing queueing delay itself — `unackedBytes / effectiveMaxRate` (both already maintained), or seal→ack minus probe-measured service RTT — fed into the `_latencyBudgetScale` governor (:477-504).
3. **Probe acceptance must require rate-up without RTT-up** (BBR probe-bw semantics): `avgRate > baseline × 1.01 && probeRtt ≤ preProbeRtt × (1+ε)`. Drop `RttFloorProbeBudgetMultiplier`.
4. **Condition-aware decay hysteresis**: bypass the 10%/s bound when `unackedBytes / rate > 2 × target` — persistent standing delay is proof the budget is too high.

## Acceptance gate

Same inversion guard as the companion 1-broker throughput issue: a fix must hold **both** 3-broker p95 ≤ 3× target **and** 1-broker throughput ≥ baseline band in the same stress run. History: #1843, #1988.

### LINKED ISSUE #2010: P1 perf: producer throughput decays -2.7%/min; 3conn stuck-low for 4 min — all-decay rate estimator + capacity probe evaluates wrong timescale against wrong statistic (#2001)
## Summary

Stress run [29254448922](https://github.com/thomhurst/Dekaf/actions/runs/29254448922) (commit e4b032c2), 1-broker fire-and-forget:

- Single producer: starts ~1.2M msg/s in minute 1, decays monotonically to ~600-650k by minute 15. Slope **-2.72%/min**, steady-state/peak **0.67** (gate needs ≥0.85). Compounded: ×0.66 over 15 min — matches the observed ratio exactly.
- 3conn variant shows the mirror image: **stuck at ~600k for the first ~4 minutes**, then jumps to ~1M at t≈239s and stays there.

Buffer stays 2-4% full throughout — this is the budget estimator drifting, not backpressure. Introduced/exposed by aa9cfdb7 (#2001, "Stabilize producer budget recovery") on top of the #1997 feed-forward budget (companion issue). This is #1856 (stuck-low no-probe equilibrium) reborn: probes now exist but are evaluated on the wrong timescale against the wrong statistic.

## Root cause

With `budget = R̂ × horizon` (see companion issue), throughput tracks the rate estimator. All estimator memory decays and the only upward force — the capacity probe — is structurally unable to succeed:

**Downward forces** (`src/Dekaf/Producer/BrokerUnackedByteBudget.cs`):
- `_windowMaxRate` forgets peaks after 2s (`WindowBucketCount=20 × 0.1s`, :62-63).
- `_retainedLoadedMaxRate` decays toward the current window max with a 60s half-life (`LoadedRateHalfLifeSeconds`, :64; applied in `AdvanceWindow` :624-631). Any dip (GC, scheduler stall on the 2-core client, broker hiccup) lowers the window max, throughput immediately follows (T = B/sojourn), and 60s later the dip is the new baseline.
- `BoundBudgetDecay` caps decline at 10%/s (:832-848) — that permits -2.7%/min a hundred times over; it is smoothing, not a floor.

**Broken upward force** — `UpdateCapacityProbe` (:656-732):
1. **Wrong timescale.** The probe evaluates 3 wire-RTTs after the first qualifying ack (`ProbeEvaluationRtts=3`, :122; deadline :689). But the system's response to a budget raise is a full pipeline sojourn: pending-append drain → up to 5ms linger → seal → send → ack ≈ 6-10ms, vs 3×RTT ≈ 1.5-6ms at 0.5-2ms wire RTT. Evaluation closes while probe-admitted bytes are still lingering in open batches — the averaged samples measure **pre-probe** throughput.
2. **Average-vs-max bias.** Success requires `averageRate > baseline × 1.01` where the baseline is `GetEffectiveMaxRate()` — a windowed **max** further inflated by the retained loaded rate (:696-697, :721). A 3-RTT average can't beat a 2s max: per-request BBR sample dispersion far exceeds the 1.25 probe gain ÷ 1.01. Probes fail even with 2× broker headroom. (Pre-#2001 compared probe *peak* vs window max — biased the other way, but roughly symmetric; #2001 changed peak→average without changing the baseline statistic.)

Failed probe → `DeactivateCapacityProbe` → decay continues → monotone drift.

**3conn stuck-low bistability:** cold start (cap = 96 MB for 3 connections) produces clustered-completion burst samples that seed `_retainedLoadedMaxRate` high; the budget then collapses onto a low first steady window. While `GetEffectiveMaxRate()` = stale-high retained rate, **every** probe fails (real traffic's average can't beat the stale baseline). The retained rate's 60s half-life is the only minutes-scale constant in the system: after ~4 half-lives (~4 min) it decays to the true window max, the first probe succeeds, and success chains (baseline := average, probe re-arms at ×1.25/×1.5, :697-705, :880-883) — producing exactly the observed step from 600k to 1M at t≈239s. The 1-conn run is the mirror: good first window → high start → monotone decay with no successful probe.

Discriminator (if diagnostics are extended): `MaxRateBytesPerSec` / `BudgetBytes` timeline (`RecordAccumulator.cs:4930-4936`) around t=239s.

## Fix plan

1. **Probe evaluation timescale**: deadline ≥ one full pipeline sojourn — `max(ProbeEvaluationRtts × rttTicks, horizonTicks + lingerTicks)` at :689; gate qualifying samples on *admission after probe start*, not send timestamp (sends after probe start still carry pre-probe admissions).
2. **Probe statistic**: compare like with like — probe-window max vs pre-probe window max, or probe average vs pre-probe windowed *average*. Never use `GetEffectiveMaxRate()` (includes the decaying retained value) as the success bar.
3. **Asymmetric memory**: a dip should not become the baseline in 60s while recovery requires a chain of fragile probes. Either lengthen the retained-rate half-life under stable conditions, or (better) adopt the MIMD latency-error growth from the companion feed-forward-budget issue, which removes the estimator's monopoly on upward movement entirely.
4. Add `BudgetBytes`/`MaxRateBytesPerSec`/probe-outcome counters to the stress connection timeline so estimator drift is visible in published run docs.

## Acceptance gate

15-min 1-broker FnF stress: slope ≥ -1.0%/min, steady-state/peak ≥ 0.85, and no stuck-low first-N-minutes phase on the 3conn variant — alongside the 3-broker p95 latency gate (inversion guard, see #1843/#1988 history).

### LINKED ISSUE #2011: P1 perf: producer CPU/msg 0.79→1.07-1.22 — coalesce windows capped at target/20 (#1997) and WaveCoalesceGate suppression (#1995) fragment requests
## Summary

Stress run [29254448922](https://github.com/thomhurst/Dekaf/actions/runs/29254448922) (commit e4b032c2): producer CPU μs/msg regressed **0.79 → 1.07-1.22** (baseline band 0.75-0.83) across 1-broker scenarios, and 3-broker FnF throughput ratio fell 1.32 → 1.03. Dekaf now does less work per core-second than before — this is not just the throughput denominator; requests got smaller.

Introduced by 1fdbde8c (#1995) + 5b9c1c75 (#1997), which were fixing #1990 (wave-coalesce spin CPU).

## Root cause: request fragmentation

Two changes halve or eliminate wave formation, so each ProduceRequest carries fewer bytes and per-request costs (framing, response handling, estimator sampling) are paid more often per message:

1. **Coalesce windows capped by latency target** (#1997): `SelectWaveCoalesceBounds` (`src/Dekaf/Producer/BrokerSender.cs:2235-2253`) now clamps quiet-window to `min(1000µs, target/20 = 500µs)` and max-window to `min(2000µs, target/10 = 1000µs)` at the default 10ms target — half the previous 1000/2000µs, regardless of the configured linger (5ms in stress). Wave formation time halves → request payload roughly halves.
2. **Spin suppression** (#1995): `WaveCoalesceGate` (`src/Dekaf/Producer/WaveCoalesceGate.cs:19-31`, wired at `BrokerSender.cs:1349-1360`) suppresses the coalesce spin after 8 fruitless spins and re-probes only every 64 sends. On sparse arrival patterns — admission-gated 1-broker traffic, or partitions fanned across 3 brokers — spins are frequently fruitless, so coalescing shuts off entirely and stays off for 64-send stretches.

Knock-on effects:
- More, smaller requests → more broker CPU per byte → lower drain rate → directly feeds the 3-broker FnF ratio drop (1.32 → 1.03) and compounds the budget-loop latency issue.
- Noisier per-request rate samples → worse behavior of the budget estimator and capacity-probe statistics (companion estimator issue).

Note: #1990's diagnosis (spin re-armed every send + per-ack budget recompute) was real; the cure overshot by also shrinking the coalesce *bounds* and making re-probe cadence fixed rather than adaptive.

## Fix plan

1. **Decouple coalesce bounds from `DeliveryLatencyTargetMs`** (`BrokerSender.cs:2243-2249`). Once the budget controller meters total delay end-to-end (companion issues), the coalesce window doesn't need to be a latency guard at all — it should scale with configured linger, not silently override it. A 500µs window to protect a 10ms budget is spending 5% of the budget to lose half the batching.
2. **Adaptive re-probe in `WaveCoalesceGate`**: re-probe on arrival-rate change signals (e.g. sends since last wake, queue depth crossing a threshold) rather than a fixed every-64-sends; or decay the suppression with time, not send count, so low-rate periods don't lock the gate shut.
3. Add avg bytes/request and requests/sec per broker to stress diagnostics so fragmentation is visible in run docs (currently only inferable from CPU/msg).

## Acceptance

CPU μs/msg back inside the 0.75-0.83 band on 1-broker producer scenarios at restored throughput (≥1.4M msg/s FnF), and 3-broker FnF ratio back above ~1.3, in the same run as the latency gates (inversion guard — #1843/#1988 history; this knob interacts with both companion budget issues).

### LINKED ISSUE #2034: P0 perf: delivery-rate sampler emits 10-15 GB/s tail-of-burst samples — budget 10-20x oversized and inert, probe baselines poisoned (96% probe failure)
## Summary

Stress run [29270211335](https://github.com/thomhurst/Dekaf/actions/runs/29270211335) (commit 072f640d, PR #2019 branch): the producer budget timelines show "max rate" estimates of **10–28 GB/s** — physically impossible on this link — and capacity probes failing 96% of the time (FnF 1b: 1,414 success / 33,845 fail; acks-all 1b: 1,284 / 33,745). The #2029 probe repair works as designed; the remaining defects are **upstream in the delivery-rate sampler** and in the **probe success baseline**.

Consequences:
- Budget = `effectiveMaxRate × horizon` (`BrokerUnackedByteBudget.cs:1093`) sized off a 10–15 GB/s garbage sample × 10–15 ms horizon = the observed 100–185 MiB acks-all budgets — 10–20x oversized and therefore **inert** (unacked ≈ 0, blocks ≈ 3/15min). An admission controller that never admits pressure is dead code at steady state, and its slow 600 s decay produces late-run block growth instead (see companion min-RTT window issue).
- Probe baselines become unbeatable → 96% failure noise masks any real capacity signal.

## Mechanism 1: tail-of-burst rate samples are inflated 10–20x

`OnAcked` (`BrokerUnackedByteBudget.cs:509-529`): `bytesPerSecond = deliveredDelta × F / intervalTicks` with `intervalTicks = max(ownRTT, now − DeliveredTimestampAtSend)`. Responses complete in clustered passes (`ProcessCompletedResponses` sweeps oldest-first, `BrokerSender.cs:2654-2751`). For the **last** request of a pipelined burst: its delivery snapshot predates the whole burst, so `deliveredDelta` = the entire pipeline (5–6 × ~640 KB ≈ 3–4 MB), while both its own RTT and its ack-axis interval collapse to one inter-pass gap (~250–400 µs). 3.8 MB / 300 µs ≈ **12 GB/s** — exactly the observed range. The sojourn guard documented at `BrokerUnackedByteBudget.cs:23-29` assumes sojourn ≥ pipeline drain time; for the tail request of a burst it is not.

The outlier then survives for minutes via `_retainedLoadedMaxRate` (600 s half-life, `BrokerUnackedByteBudget.cs:74,125-126,784-791`) and sizes the budget. It also corrupts RTT queue-correction, which divides by this rate (`CorrectRttForSelfQueueing`, `:659-669`).

## Mechanism 2: probe success poisons the next rung

On success the baseline becomes `Math.Max(averageRate, _capacityProbeMaxRate)` (`BrokerUnackedByteBudget.cs:887`) — the max of per-request burst samples, i.e. the 10–15 GB/s outliers. The next rung must beat a 10 GB/s outlier with a ~1 GB/s average → deterministic failure. Observed 1:24 success:failure ratio matches "occasional noise success, immediately poisoned next rung".

## Mechanism 3: probes arm while the gate isn't gating

Unacked bytes never reach the budget, so publishing budget × 1.25 admits zero additional bytes; delivered rate cannot rise 1% → the rate test (`:878-882`) fails every time. Probing an open gate is unwinnable by construction, yet each attempt increments the failure counter (`DeactivateCapacityProbe`, `:939-947`) — ~34k meaningless cycles/run that conflate "no demand" with "no headroom".

## Fixes (ranked)

1. **Fix the rate sampler** (`:509-529`): carry a delivery-epoch send-axis anchor in `DeliverySnapshot` (Linux tcp_rate style: `interval = max(ackElapsed, sendElapsed since the delivered-snapshot's last sent packet)`), or bound `deliveredDelta/interval` by line rate (`interval ≥ delta/lineRate`). Everything downstream becomes meaningful.
2. **Stop poisoning rungs**: at `:887` use the accepted windowed *average* as the next baseline (drop the `Math.Max(…, _capacityProbeMaxRate)`), comparing like with like across rungs.
3. **Don't arm probes while `UnackedBytes < budgetBytes/2`** — removes ~34k noise cycles/run and makes the success/failure diagnostic actually signal headroom.

## Acceptance

- Budget timeline max-rate ≤ observed NIC line rate (~1–2 GB/s on these runners).
- Probe success/failure ratio meaningful (failures should reflect genuine no-headroom verdicts, not baseline poisoning).
- Steady-state budget within ~2x of true BDP (rate × loaded RTT ≈ 5–15 MiB here) so admission control is live rather than inert.
- 3-broker latency gates stay green.

File:line references at commit 072f640d (PR #2019 branch). Refines #2010 (the evaluation-window part of #2010 was fixed by #2029; this is what remains). Related: #2008, #2019. Data: run 29270211335.

### LINKED ISSUE #2035: P1 perf: unacked-budget cold-start collapse (32 to 3.8 MiB in first second) halves round-trip/short-burst throughput; idle never restores cap; CPU/msg 1.60x Confluent
## Summary

Stress run [29270211335](https://github.com/thomhurst/Dekaf/actions/runs/29270211335) (commit 072f640d, PR #2019 branch): Producer → Consumer Round-Trip regressed to fail the trend gate:

- Dekaf 115,684 msg/s vs baseline 204,313 (band 152k–256k) — **repeated regression (fail)**.
- CPU 11.00 µs/msg vs baseline 6.86 — CPU/msg vs Confluent now **1.60x** (baseline 0.94; Dekaf used to be more efficient).
- Requests shrank to 415 KB avg (vs 594–638 KB in pure-producer scenarios): 628 requests over a ~1.75 s produce phase.
- Budget timeline: t=0 budget 32.0 MiB → **t=1s budget 3.8 MiB**, probes 0/24, blocks 0.

The queue-corrected RTT sensing from #2019 (f5e83c53) fixed the #2009 ratchet, but that ratchet was what had been *accidentally* keeping the budget near cap on short bursts. With sensing fixed, the budget collapses to `rate × ~10 ms` in the first second and a 2-second burst has no path back up.

## Mechanism (all in `src/Dekaf/Producer/BrokerUnackedByteBudget.cs`)

1. **The cold-start guarantee lasts exactly one ack.** While `effectiveMaxRate == 0`, `ComputeBudget` returns the cap (`:1069-1075`). The first acked pass calls `CompleteAckedPass` → `RecomputeBudget` (`:549-557`) → budget = `effectiveMaxRate × horizon` ≈ 430 MB/s × ~9 ms ≈ 3.8 MiB.
2. **Downward smoothing can't engage — ordering hole.** `BoundBudgetDecay` (`:1036-1060`) applies the 10%/s decay bound only when `admissionWasBlocked`. But a block cannot be recorded (`RecordAccumulator.cs:4746-4747`) until *after* the budget has already collapsed below standing unacked. On the first recompute `_lastBudgetUpdateTimestamp == 0` ⇒ `elapsedSeconds = 0` ⇒ raw snap (`:1043-1048`). 32 → 3.8 MiB is one instantaneous recompute, not decay.
3. **Idle never restores the cap.** The scenario warms up (1000 msgs), then idles several seconds (end-offset queries + 3× `GC.Collect()`, `ProducerRoundTripStressTest.cs:70-79`). `sustainedIdle` zeroes `_retainedLoadedMaxRate` (`:805-806`) and a >2 s gap clears the rate window (`:749-758`) — but `RecomputeBudget` only runs on ack passes (`:556`), so the published budget stays at the stale ~4 MiB from warmup. The burst starts pre-throttled; the cold-start branch (`:1069-1075`) is unreachable once any historical sample exists.
4. **No upward force can act in a 2 s burst.** Probes: success requires rate +1% with RTT ≤ 1.25x (`:878-882`) — unwinnable while supply-limited (blocks=0 → raising the budget admits nothing) and once gate-bound, filling the missing BDP necessarily inflates RTT past tolerance (that's the anti-bufferbloat rule, `:872-877`). All 24 startup probes fail. `_latencyBudgetScale` growth (#2020, `:621-628`) requires `blocked && latency < target` — mutually exclusive states here — and at 1.05×/100 ms (`:94,127`) would take far longer than the burst anyway.
5. **Downstream effects**: admission gates supply into the accumulator (`TryAdmitAndReserve`, `RecordAccumulator.cs:4698-4708`), holding it near-empty → waves seal small → 415 KB requests. CPU/msg 11.0 is mostly the stretched produce phase (≈1 extra wall-second × fixed background burn on 2 cores ≈ the whole 1.04 CPU-s delta) plus gated-append churn (block→drain cycles through `DrainPendingAppends` under lock, `RecordAccumulator.cs:1996-2099`, with a cross-thread ValueTask resume per gated message).

The 2.02 KB/msg allocation in this scenario is harness + typed-consumer cost (per-message payload encode for validation, `RoundTripMessageCodec.cs:170`; per-record `string`/`byte[]` materialization in consume), not a library leak — expected, and not the regression.

## Fixes (ranked)

1. **Smooth all downward budget moves**: apply the 10%/s decay bound in `BoundBudgetDecay` (`:1036-1060`) unconditionally (keep the existing `unacked ≥ 2×computed` fast-drain bypass at `:1054-1055`). From 32 MiB a 2 s burst would still see >26 MiB — no throttle, no probe dependence. Lowest risk, one function.
2. **Treat "no blocks and unacked ≪ budget" as app-limited**: skip downward recompute (and the `_latencyBudgetScale` shrink branch, `:603-620`) when the window recorded zero admission blocks and occupancy < ~half budget — the link is starved by supply, not capacity.
3. **Restore cap on sustained idle**: when the idle path clears the rate state (`:749-758,805-806`), republish cold-start semantics instead of leaving the stale small `_lastNormalBudgetBytes`.
4. **Probe warm-up (BBR STARTUP analogue)**: until a probe fails *while admission is blocked* (proof of a real ceiling), grow multiplicatively (e.g. 2×/probe) requiring only rate-not-down; the current rate-up-with-flat-RTT rule cannot recover a budget below true BDP.
5. **Don't count unwinnable probes as failures** (zero qualifying samples `:853-854`, or zero-block windows) — re-arm instead of incrementing `_capacityProbeFailureCount` (`:939-947`).

## Acceptance

- Roundtrip produce phase ≥ 200k msg/s (budget ≥ true BDP ≈ 5–6 MiB throughout the burst), blocks ≈ 0.
- CPU/msg ≤ 6.9 µs (back below Confluent).
- 3-broker acks-all latency gate stays green (the #2009 fix must survive — dual acceptance gate).

File:line references at commit 072f640d (PR #2019 branch). Related: #2008 (tracked the pre-#2019 form of this), #2009, #2019, #2020, #2029. Data: run 29270211335.

### LINKED ISSUE #2043: P0 perf: min-RTT refresh accepts queued outlier and jumps budget 6.9x
## Problem

Combined one-broker `producer-acks-all` acceptance on the latest #2034 + #2036 review heads still fails late-run stability. At 35-36s, the periodic minimum-RTT refresh replaces a 200 us floor with a queued 1,764 us probe sample in one step. With max delivery rate stable near 2.67 GB/s and latency scale at 0.1, the RTT safety horizon becomes dominant and the broker budget jumps from 2.6 MiB to 17.1 MiB (6.9x).

The larger standing queue is followed by throughput intervals down to 29,464 msg/s, p50/p95 delivery latency 25.75/114.95ms, and -21.2% first-third/last-third drift. Request shape remains healthy at 803.6 KiB/request, so this is not #2033 batching fragmentation or #2034 delivery-rate inflation.

## Root cause

`CompleteMinRttProbe` accepts `_minRttProbeMinimumSeconds` unbounded. A brief drain probe can remain partially queued across parallel broker connections, so one transient loaded window can replace a valid low floor with a much larger service estimate. The next budget recompute immediately multiplies the inflated floor by the RTT safety factors and current delivery rate.

## Expected

- A single refresh window cannot increase the minimum-RTT floor by an unbounded factor.
- Lower measurements remain immediately admissible.
- A sustained real path-latency increase can still converge over successive refresh windows.
- Unit coverage pins both bounded upward refresh and immediate downward refresh.
- Combined producer stress no longer shows the min-RTT-driven budget jump.

## Evidence

Detached acceptance head `7addeb05` (parent #2019 plus final review fixes from #2038/#2039), local 1-minute Dekaf-only one-broker `producer-acks-all` run:

- 10,446,389 delivered, 174,067 msg/s average, 198,284 median
- 803.6 KiB/request; max rate stable ~2.67 GB/s
- min RTT 200 -> 1,764 us; budget 2.6 -> 17.1 MiB
- p50 25.75ms, p95 114.95ms, max 1.2s
- drift -21.2%, slope -36.65%/min
- 100% broker-confirmed delivery, 0 errors

Blocks final acceptance of #2019 / #2009.

thomhurst:
Fixes #2009.
Fixes #2008.
Fixes #2010.
Fixes #2011.
Fixes #2035.
Fixes #2034.
Fixes #2043.

## Problem

Stress validation exposed coupled producer-control and scheduling failures:

- **#2009:** 3-broker delivery latency reached 26x target because `servingRttEwma` measured queueing admitted by the budget itself, creating a gain-above-one feedback loop.
- **#2008/#2010:** the feed-forward budget had no reliable upward recovery path; probe window mismatches and transient stalls caused stuck-low throughput.
- **#2011:** the one-shot active linger timer was rearmed for every send wave, dominating CPU and collapsing 1-broker acks-all throughput.
- **#2035:** the first acknowledgement snapped the 32 MiB cold-start cap to about 3.8 MiB because downward hysteresis applied only after admission blocks and treated the first elapsed interval as zero.

## Fix

1. Queue-correct RTT samples before updating minimum and serving estimates.
2. Bound loaded serving RTT to `2 x minRtt`, preventing residual feedback ratchets.
3. Compare capacity-probe rate and RTT in matching raw domains over wholly admitted traffic.
4. Retain demonstrated loaded capacity through transient stalls and report probe outcomes.
5. Let latency-governed scale grow above parity under proven demand; distinguish broker queueing from sender backlog when shrinking.
6. Preserve the standing-queue fast-drain bypass while rate-limiting every ordinary downward budget recompute to 10%/s, including the zero-elapsed first sample.
7. Arm linger with the existing idle-to-active signal, then use a reusable `PeriodicTimer` for active cadence; signal only on the first queue transition.
8. Isolate the process-wide consumer fetch diagnostics listener test from unrelated parallel tests.

## Tests

- Budget-controller state-machine suite: 73/73.
- Full net10 unit suite: 5,423/5,423.
- Release build: 0 warnings, 0 errors.
- Exact-head CI [29321669673](https://github.com/thomhurst/Dekaf/actions/runs/29321669673) and 15-minute stress [29327573274](https://github.com/thomhurst/Dekaf/actions/runs/29327573274) passed at `389f4667`; final Claude review must be CLEAR before merge.

## Validation gate

Merge remains gated on one exact-head 15-minute stress run proving:

- 3-broker p95 delivery latency <= 3x target.
- 1-broker acks-all throughput within the baseline band.
- Round-trip produce >= 200k msg/s and CPU <= 6.9 us/msg.

## Final exact-head validation

At `389f4667`:

- CI run 29321669673 passed.
- Stress run 29327573274 passed all 15 jobs.
- Worst 3-broker p95 delivery latency: 24.75 ms (< 3x 10 ms target).
- 1-broker acks-all: 1,285,376 median msg/s, zero errors.
- Roundtrip: 532,899 produce msg/s, 4.01 CPU us/msg, 2,000,000/2,000,000 consumed, zero missing/duplicate/corrupt/out-of-order/mispartitioned messages.

### DIFF
--- a/.github/scripts/stress_report.py
+++ b/.github/scripts/stress_report.py
@@ -613,6 +613,66 @@ def format_consumer_fetch_timeline(results, title):
     return lines
 
 
+def format_producer_budget_timeline(results, title):
+    """Correlate producer budget/rate/probe state with throughput intervals."""
+    rows = []
+    for result in results:
+        diagnostics = result.get('producerDeliveryDiagnostics') or {}
+        for sample in diagnostics.get('brokerBudgetSamples') or []:
+            rows.append((result, sample))
+    if not rows:
+        return []
+
+    rows.sort(key=lambda row: row[1].get('capturedAtUtc', ''))
+    displayed_rows = _sample_evenly(rows, MAX_TIMELINE_ROWS)
+    omitted_rows = len(rows) - len(displayed_rows)
+    lines = [
+        f"## Producer Budget Timeline - {title}",
+        "",
+        "| Client | Sample UTC | Broker | Budget / unacked | Max rate | Probe success/failure | Admission blocks | Nearest throughput sample |",
+        "|--------|------------|-------:|------------------|---------:|----------------------:|-----------------:|---------------------------|",
+    ]
+    for result, sample in displayed_rows:
+        sample_time = _parse_timestamp(sample.get('capturedAtUtc'))
+        intervals = result.get('throughput', {}).get('intervalSamples') or []
+        timestamped_intervals = [
+            (interval, _parse_timestamp(interval.get('capturedAtUtc')))
+            for interval in intervals
+        ]
+        timestamped_intervals = [
+            (interval, captured_at)
+            for interval, captured_at in timestamped_intervals
+            if captured_at is not None
+        ]
+        nearest = min(
+            timestamped_intervals,
+            key=lambda item: abs((item[1] - sample_time).total_seconds()),
+            default=(None, None),
+        )[0] if sample_time is not None else None
+        throughput = '-'
+        if nearest is not None:
+            throughput = (
+                f"{nearest.get('elapsedSeconds', 0):.1f}s / "
+                f"{nearest.get('messagesPerSecond', 0):,.0f} msg/s"
+            )
+        lines.append(
+            f"| {result.get('client', 'Unknown')} | {sample.get('capturedAtUtc', '-')} | "
+            f"{sample.get('brokerId', '-')} | "
+            f"{sample.get('budgetBytes', 0) / 1_048_576:.1f} MiB / "
+            f"{sample.get('unackedBytes', 0) / 1_048_576:.1f} MiB | "
+            f"{sample.get('maxRateBytesPerSec', 0) / 1_000_000:.1f} MB/s | "
+            f"{sample.get('capacityProbeSuccessCount', 0):,}/"
+            f"{sample.get('capacityProbeFailureCount', 0):,} | "
+            f"{sample.get('admissionBlockCount', 0):,} | {throughput} |"
+        )
+    if omitted_rows > 0:
+        lines.append(
+            f"*{omitted_rows:,} budget sample(s) omitted; rows sampled across the full timeline.*"
+        )
+    lines.append("")
+    return lines
+
+
 def format_latency_outlier_timeline(results, title):
     """Correlate sampled delivery stalls with scaling, throughput, and GC."""
     rows = [
@@ -902,6 +962,7 @@ def generate_scenario_tables(results, include_ratio=False, include_callout=False
             output.extend(format_connection_scale_timeline(scenario_results, title))
             output.extend(format_producer_request_diagnostics(scenario_results, title))
             output.extend(format_consumer_fetch_timeline(scenario_results, title))
+            output.extend(format_producer_budget_timeline(scenario_results, title))
             output.extend(format_latency_outlier_timeline(scenario_results, title))
             output.extend(format_transaction_verification(scenario_results))
             output.extend(format_roundtrip_validation_table(scenario_results))
--- a/.github/scripts/test_stress_report.py
+++ b/.github/scripts/test_stress_report.py
@@ -5,6 +5,7 @@ from stress_report import (
     format_roundtrip_validation_table,
     format_connection_scale_timeline,
     format_producer_request_diagnostics,
+    format_producer_budget_timeline,
     format_throughput_table,
     generate_scenario_tables,
     intra_run_throughput,
@@ -120,6 +121,33 @@ class StressReportTests(unittest.TestCase):
         self.assertIn("1→2", report)
         self.assertIn("20.0s / 2,000 msg/s", report)
         self.assertIn("120/150", report)
+
+    def test_producer_budget_timeline_surfaces_probe_outcomes(self):
+        value = stress_result("Dekaf", effective_rate=1400)
+        value["throughput"]["intervalSamples"] = [{
+            "capturedAtUtc": "2026-07-12T02:20:50+00:00",
+            "elapsedSeconds": 20.0,
+            "messagesPerSecond": 2000.0,
+        }]
+        value["producerDeliveryDiagnostics"] = {
+            "brokerBudgetSamples": [{
+                "capturedAtUtc": "2026-07-12T02:20:49+00:00",
+                "brokerId": 1,
+                "budgetBytes": 8 * 1024 * 1024,
+                "unackedBytes": 6 * 1024 * 1024,
+                "maxRateBytesPerSec": 750_000_000,
+                "capacityProbeSuccessCount": 3,
+                "capacityProbeFailureCount": 2,
+                "admissionBlockCount": 12,
+            }]
+        }
+
+        report = "\n".join(format_producer_budget_timeline([value], "Producer"))
+
+        self.assertIn("8.0 MiB", report)
+        self.assertIn("750.0 MB/s", report)
+        self.assertIn("3/2", report)
+        self.assertIn("20.0s / 2,000 msg/s", report)
 
     def test_scenario_tables_surface_latency_outlier_diagnostics(self):
         value = stress_result("Dekaf", effective_rate=1400)
--- a/src/Dekaf/Builders.cs
+++ b/src/Dekaf/Builders.cs
@@ -88,7 +88,8 @@ public sealed class ProducerBuilder<TKey, TValue>
     private RemoteCertificateValidationCallback? _remoteCertificateValidationCallback;
     private IRetryPolicy? _retryPolicy;
     private bool _enableAdaptiveConnections = true;
-    private int _maxConnectionsPerBroker = 10;
+    private int _maxConnectionsPerBroker = ProducerOptions.DefaultMaxConnectionsPerBroker;
+    private bool _isMaxConnectionsPerBrokerConfigured;
     private bool _enableDeliveryDiagnostics;
     private readonly Dictionary<string, ApplicationTelemetryMetric> _applicationMetrics = new(StringComparer.Ordinal);
 
@@ -103,6 +104,7 @@ public sealed class ProducerBuilder<TKey, TValue>
         _loggerFactory = clientInfrastructure.LoggerFactory;
         _connectionsPerBroker = clientInfrastructure.ConnectionsPerBroker;
         _maxConnectionsPerBroker = clientInfrastructure.MaxConnectionsPerBroker;
+        _isMaxConnectionsPerBrokerConfigured = true;
     }
 
     public ProducerBuilder<TKey, TValue> WithBootstrapServers(string servers)
@@ -279,13 +281,15 @@ public sealed class ProducerBuilder<TKey, TValue>
         ArgumentOutOfRangeException.ThrowIfLessThan(connectionsPerBroker, 1);
         ArgumentOutOfRangeException.ThrowIfGreaterThan(connectionsPerBroker, 32);
         _connectionsPerBroker = connectionsPerBroker;
+        if (!_isMaxConnectionsPerBrokerConfigured)
+            _maxConnectionsPerBroker = Math.Max(_maxConnectionsPerBroker, connectionsPerBroker);
         return this;
     }
 
     /// <summary>
     /// Configures the maximum connections for adaptive connection scaling.
     /// Adaptive scaling is enabled by default — this method only needs to be called
-    /// to change the maximum from the default of 10.
+    /// to change the maximum from the default of 3.
     /// <para>
     /// When sustained backpressure is detected, the producer will automatically add connections
     /// per broker (up to <paramref name="maxConnections"/>) to increase drain throughput.
@@ -297,13 +301,15 @@ public sealed class ProducerBuilder<TKey, TValue>
     /// are removed after sustained low utilization.
     /// </para>
     /// </summary>
-    /// <param name="maxConnections">Maximum connections per broker. Default: 10.</param>
-    public ProducerBuilder<TKey, TValue> WithAdaptiveConnections(int maxConnections = 10)
+    /// <param name="maxConnections">Maximum connections per broker. Default: 3.</param>
+    public ProducerBuilder<TKey, TValue> WithAdaptiveConnections(
+        int maxConnections = ProducerOptions.DefaultMaxConnectionsPerBroker)
     {
         ThrowIfClientOwnedConnectionSettings();
         ArgumentOutOfRangeException.ThrowIfLessThan(maxConnections, 1);
         _enableAdaptiveConnections = true;
         _maxConnectionsPerBroker = maxConnections;
+        _isMaxConnectionsPerBrokerConfigured = true;
         return this;
     }
 
--- a/src/Dekaf/Builders.cs
+++ b/src/Dekaf/Builders.cs
@@ -103,7 +103,7 @@ public sealed class ProducerBuilder<TKey, TValue>
         _bootstrapServers = clientInfrastructure.BootstrapServers;
         _loggerFactory = clientInfrastructure.LoggerFactory;
         _connectionsPerBroker = clientInfrastructure.ConnectionsPerBroker;
-        _maxConnectionsPerBroker = clientInfrastructure.MaxConnectionsPerBroker;
+        _maxConnectionsPerBroker = clientInfrastructure.ProducerMaxConnectionsPerBroker;
         _isMaxConnectionsPerBrokerConfigured = true;
     }
 
@@ -1424,7 +1424,7 @@ public sealed class ConsumerBuilder<TKey, TValue>
     private int _prefetchPipelineDepth = 3;
     private int _connectionsPerBroker = 2;
     private bool _enableAdaptiveConnections = true;
-    private int _maxConnectionsPerBroker = 4;
+    private int _maxConnectionsPerBroker = ConsumerOptions.DefaultMaxConnectionsPerBroker;
     private int _reconnectBackoffMs = 50;
     private int _reconnectBackoffMaxMs = 1000;
     private int _connectionsMaxIdleMs = ConnectionOptions.DefaultConnectionsMaxIdleMs;
--- a/src/Dekaf/Consumer/ConsumerOptions.cs
+++ b/src/Dekaf/Consumer/ConsumerOptions.cs
@@ -33,6 +33,8 @@ public enum OffsetCommitMode
 /// </summary>
 public sealed class ConsumerOptions
 {
+    internal const int DefaultMaxConnectionsPerBroker = 4;
+
     /// <summary>
     /// Bootstrap servers (host:port,host:port).
     /// </summary>
@@ -419,7 +421,7 @@ public sealed class ConsumerOptions
         }
     }
 
-    private readonly int _maxConnectionsPerBroker = 4;
+    private readonly int _maxConnectionsPerBroker = DefaultMaxConnectionsPerBroker;
 
     /// <summary>
     /// Whether to enable adaptive fetch sizing based on consumer throughput.
--- a/src/Dekaf/KafkaClient.cs
+++ b/src/Dekaf/KafkaClient.cs
@@ -96,6 +96,7 @@ public sealed class KafkaClientBuilder
     private int _connectionsPerBroker = 1;
     private int _maxInFlightRequestsPerConnection = 5;
     private int _maxConnectionsPerBroker = ProducerOptions.DefaultMaxConnectionsPerBroker;
+    private bool _isMaxConnectionsPerBrokerConfigured;
     private int _connectionsMaxIdleMs = ConnectionOptions.DefaultConnectionsMaxIdleMs;
     private TimeSpan _connectionTimeout = ConnectionOptions.DefaultConnectionTimeout;
     private bool _enableTcpKeepAlive = ConnectionOptions.DefaultEnableTcpKeepAlive;
@@ -369,6 +370,8 @@ public sealed class KafkaClientBuilder
         ArgumentOutOfRangeException.ThrowIfLessThan(connectionsPerBroker, 1);
         ArgumentOutOfRangeException.ThrowIfGreaterThan(connectionsPerBroker, 32);
         _connectionsPerBroker = connectionsPerBroker;
+        if (!_isMaxConnectionsPerBrokerConfigured)
+            _maxConnectionsPerBroker = Math.Max(_maxConnectionsPerBroker, connectionsPerBroker);
         return this;
     }
 
@@ -384,6 +387,7 @@ public sealed class KafkaClientBuilder
     {
         ArgumentOutOfRangeException.ThrowIfLessThan(maxConnectionsPerBroker, 1);
         _maxConnectionsPerBroker = maxConnectionsPerBroker;
+        _isMaxConnectionsPerBrokerConfigured = true;
         return this;
     }
 
--- a/src/Dekaf/KafkaClient.cs
+++ b/src/Dekaf/KafkaClient.cs
@@ -5,6 +5,7 @@ using Admin;
 using Dekaf.Internal;
 using Dekaf.Metadata;
 using Dekaf.Networking;
+using Dekaf.Producer;
 using Dekaf.Security;
 using Dekaf.Security.Sasl;
 using Microsoft.Extensions.Logging;
@@ -94,7 +95,7 @@ public sealed class KafkaClientBuilder
     private int _socketReceiveBufferBytes;
     private int _connectionsPerBroker = 1;
     private int _maxInFlightRequestsPerConnection = 5;
-    private int _maxConnectionsPerBroker = 10;
+    private int _maxConnectionsPerBroker = ProducerOptions.DefaultMaxConnectionsPerBroker;
     private int _connectionsMaxIdleMs = ConnectionOptions.DefaultConnectionsMaxIdleMs;
     private TimeSpan _connectionTimeout = ConnectionOptions.DefaultConnectionTimeout;
     private bool _enableTcpKeepAlive = ConnectionOptions.DefaultEnableTcpKeepAlive;
--- a/src/Dekaf/KafkaClient.cs
+++ b/src/Dekaf/KafkaClient.cs
@@ -2,6 +2,7 @@ namespace Dekaf;
 
 using System.Net.Security;
 using Admin;
+using Dekaf.Consumer;
 using Dekaf.Internal;
 using Dekaf.Metadata;
 using Dekaf.Networking;
@@ -95,7 +96,8 @@ public sealed class KafkaClientBuilder
     private int _socketReceiveBufferBytes;
     private int _connectionsPerBroker = 1;
     private int _maxInFlightRequestsPerConnection = 5;
-    private int _maxConnectionsPerBroker = ProducerOptions.DefaultMaxConnectionsPerBroker;
+    private int _maxConnectionsPerBroker = ConsumerOptions.DefaultMaxConnectionsPerBroker;
+    private int _producerMaxConnectionsPerBroker = ProducerOptions.DefaultMaxConnectionsPerBroker;
     private bool _isMaxConnectionsPerBrokerConfigured;
     private int _connectionsMaxIdleMs = ConnectionOptions.DefaultConnectionsMaxIdleMs;
     private TimeSpan _connectionTimeout = ConnectionOptions.DefaultConnectionTimeout;
@@ -371,7 +373,10 @@ public sealed class KafkaClientBuilder
         ArgumentOutOfRangeException.ThrowIfGreaterThan(connectionsPerBroker, 32);
         _connectionsPerBroker = connectionsPerBroker;
         if (!_isMaxConnectionsPerBrokerConfigured)
+        {
             _maxConnectionsPerBroker = Math.Max(_maxConnectionsPerBroker, connectionsPerBroker);
+            _producerMaxConnectionsPerBroker = Math.Max(_producerMaxConnectionsPerBroker, connectionsPerBroker);
+        }
         return this;
     }
 
@@ -387,6 +392,7 @@ public sealed class KafkaClientBuilder
     {
         ArgumentOutOfRangeException.ThrowIfLessThan(maxConnectionsPerBroker, 1);
         _maxConnectionsPerBroker = maxConnectionsPerBroker;
+        _producerMaxConnectionsPerBroker = maxConnectionsPerBroker;
         _isMaxConnectionsPerBrokerConfigured = true;
         return this;
     }
@@ -485,6 +491,7 @@ public sealed class KafkaClientBuilder
             ConnectionsPerBroker = _connectionsPerBroker,
             MaxInFlightRequestsPerConnection = _maxInFlightRequestsPerConnection,
             MaxConnectionsPerBroker = _maxConnectionsPerBroker,
+            ProducerMaxConnectionsPerBroker = _producerMaxConnectionsPerBroker,
             ConnectionsMaxIdleMs = _connectionsMaxIdleMs,
             MetadataMaxAge = _metadataMaxAge,
             MetadataRecoveryStrategy = _metadataRecoveryStrategy,
@@ -525,6 +532,7 @@ internal sealed class KafkaClientOptions
     public int ConnectionsPerBroker { get; init; }
     public int MaxInFlightRequestsPerConnection { get; init; }
     public int MaxConnectionsPerBroker { get; init; }
+    public int ProducerMaxConnectionsPerBroker { get; init; }
     public int ConnectionsMaxIdleMs { get; init; }
     public TimeSpan? MetadataMaxAge { get; init; }
     public MetadataRecoveryStrategy MetadataRecoveryStrategy { get; init; }
@@ -547,7 +555,8 @@ internal sealed class KafkaClientInfrastructure : IAsyncDisposable
         IDekafMemoryBudget memoryBudget,
         ILoggerFactory? loggerFactory,
         int connectionsPerBroker,
-        int maxConnectionsPerBroker)
+        int maxConnectionsPerBroker,
+        int producerMaxConnectionsPerBroker)
     {
         BootstrapServers = bootstrapServers;
         ConnectionPool = connectionPool;
@@ -556,6 +565,7 @@ internal sealed class KafkaClientInfrastructure : IAsyncDisposable
         LoggerFactory = loggerFactory;
         ConnectionsPerBroker = connectionsPerBroker;
         MaxConnectionsPerBroker = maxConnectionsPerBroker;
+        ProducerMaxConnectionsPerBroker = producerMaxConnectionsPerBroker;
     }
 
     public IReadOnlyList<string> BootstrapServers { get; }
@@ -565,6 +575,7 @@ internal sealed class KafkaClientInfrastructure : IAsyncDisposable
     public ILoggerFactory? LoggerFactory { get; }
     public int ConnectionsPerBroker { get; }
     public int MaxConnectionsPerBroker { get; }
+    public int ProducerMaxConnectionsPerBroker { get; }
 
     public static KafkaClientInfrastructure Create(KafkaClientOptions options)
     {
@@ -630,7 +641,8 @@ internal sealed class KafkaClientInfrastructure : IAsyncDisposable
             new ClientMemoryBudget(options.MemoryBudgetBytes),
             options.LoggerFactory,
             options.ConnectionsPerBroker,
-            options.MaxConnectionsPerBroker);
+            options.MaxConnectionsPerBroker,
+            options.ProducerMaxConnectionsPerBroker);
     }
 
     public async ValueTask DisposeAsync()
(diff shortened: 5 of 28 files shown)


thomhurst:
Rebased the final #2009/#2008/#2010 stack onto current main at eab99fd58c612ebb230ace9d20bc1422fa9a1576 after #2029 merged into this branch. Conflict resolution preserved producer-request, consumer-fetch/GC, and producer-budget timelines. Validation: Python reporter 19/19; focused 71+18+2; full net10 unit 5,417/5,417. The old b8bf759e stress run was cancelled as non-exact-head; a fresh 15-minute exact-head run was dispatched.

thomhurst:
Fixed the CI-only unit-test race at exact head \48e0a05d\: \Listener_CollectsFetchDurationAndTopicBytes\ now runs process-exclusively because enabling its \MeterListener\ makes unrelated concurrent consumer tests emit into the process-wide fetch instrument. Focused test: 1/1. Full net10 unit suite: 5,417/5,417. Obsolete stress run cancelled; replacement exact-head 15-minute run: https://github.com/thomhurst/Dekaf/actions/runs/29270084791

thomhurst:
Rebased cleanly again onto current main \12bb89d9\ (PR #2030 transactional send-loop change) so performance validation covers the final combined producer behavior. New exact head: \

thomhurst:
Stress validation of this branch: run [29270211335](https://github.com/thomhurst/Dekaf/actions/runs/29270211335) (072f640d).

**Fixed (keep):** the #2009 latency ratchet is gone — all delivery-latency gates green (3b acks-all p95/target 2.04, p50/Confluent 1.24, p99 0.69; worst p95/target 2.12 vs the previous 26x).

**Not fixed by this branch (pre-existing, now root-caused elsewhere):**
- 1-broker throughput still 0.61-0.83x Confluent — the budget is provably not the constraint on this branch (acks-all: 150+ MiB budget, ~0 unacked, 3 blocks/15min). Real ceiling = one-batch-per-request wire shape + single-threaded send path → #2033; garbage rate samples making the budget inert → #2034.
- FnF 1b steady/peak 0.78 gate fail (min-RTT windows + slow-path latch) → #2036.

**Made worse by this branch (side effect to address before/with merge):** round-trip halved (204k→115k, CPU/msg 1.60x Confluent). The queue-corrected sensing removed the accidental ratchet that kept short-burst budgets near cap; cold-start now collapses 32→3.8 MiB in the first second with no recovery path in a 2s burst → #2035 (fix 1 there — unconditional downward smoothing in `BoundBudgetDecay` — is small and would fit this PR).

Net: merging fixes the P0 latency inversion but trades the round-trip scenario; recommend picking up #2035 fix 1 in this branch, and treating #2033/#2034/#2036 as follow-ups.

thomhurst:
Exact-head 15-minute validation dispatched at d073b901: https://github.com/thomhurst/Dekaf/actions/runs/29279470918. Merge remains gated on 3-broker latency, 1-broker throughput, and round-trip >=200k / CPU <=6.9.

thomhurst:
Early round-trip artifact from exact-head run 29279470918: correctness PASS; produce 271,661 msg/s (>=200k); budget 29 MiB after 1s (cold-start collapse fixed). CPU is 7.20 us/msg, above <=6.9 gate, so not mergeable yet. This run also predates stacked #2038/#2039 and main-side #2042; final stress must be rerun after those land.

thomhurst:
Latest combined local acceptance (detached dcefa04a: merged #2038, latest #2039, and #2044) removes the late control collapse. 1-minute one-broker Dekaf acks-all delivered 12,836,611/12,836,611 (0 errors), 213,894 msg/s average / 215,252 median, +8.7% drift, +15.94% slope, 793 KiB/request, p50/p95 24.35/40.75ms. Min-RTT upward steps were all <=1.25x and budget stayed ~4-5.7 MiB after ramp; previous combined head jumped 2.6->17.1 MiB and collapsed to 29k msg/s. This is local evidence; final dedicated-runner acceptance still required after child PRs merge.

thomhurst:
Rebased exact stack onto main at a3d23206, incorporating merged #2042. Clean rebase; Release unit build 0 warnings/errors; net10 unit suite 5,430/5,430. Exact-head CI/review restarted.

thomhurst:
Exact-head 15-minute validation completed successfully at `a3d23206`: [run 29281633009](https://github.com/thomhurst/Dekaf/actions/runs/29281633009). Every job passed.

- 1b FnF: 1.496M vs 1.175M msg/s (1.27x), 0.887 us/msg; 3conn: 1.709M (1.45x), 0.82 us/msg.
- 1b acks-all: 1.334M vs 1.116M (1.20x), 0.99 us/msg.
- 1b idempotent: 1.613M vs 1.299M (1.24x), 0.847 us/msg; 3conn: 1.809M (1.39x), 0.80 us/msg.
- 3b FnF/acks-all/idempotent: 1.28x/1.27x/1.28x Confluent; p95 28.45/22.15/22.55 ms; zero errors.
- Round-trip: 250,000/250,000 correct; effective 261,952 msg/s; 5.392 us/msg; zero missing/duplicate/corrupt/out-of-order.
- Transactional EOS: PASS.

The prior exact-head review's missing-stress blocker is satisfied. The run also confirms residual #2034 telemetry on this head (1b max-rate 15.7-30.0 GB/s and 149-192 MiB budgets), now addressed in stacked PR #2046. Final merge remains gated on #2039 then #2046 and one exact combined-head run.

thomhurst:
Blocking shared-client regression addressed in stacked PR #2052 at `c397f9de`.

- implicit maximum now expands when `WithConnectionsPerBroker` selects a larger initial width
- explicitly configured maxima remain authoritative and retain the build-time invariant
- net10/net8 focused `KafkaClientBuilderTests`: 14/14 on each; both builds clean
- exact CI: https://github.com/thomhurst/Dekaf/actions/runs/29303732234

Will merge into this branch after exact CI/review are clear.

thomhurst:
Final exact-head merge-gate evidence for `389f466771e641a7c3f7436cba77db0278d39f8e`:

- CI: [29321669673](https://github.com/thomhurst/Dekaf/actions/runs/29321669673), success.
- 15-minute stress: [29327573274](https://github.com/thomhurst/Dekaf/actions/runs/29327573274), all 15 jobs success.
- Worst 3-broker Dekaf p95: 24.75 ms (< 30 ms gate); all producer scenarios had zero errors.
- 1-broker acks-all: 1,285,376 median msg/s.
- Roundtrip: 532,899 produce msg/s (>200k), 4.01 CPU us/msg (<6.9), 2,000,000/2,000,000 consumed, zero missing/duplicate/corrupt/out-of-order/mispartitioned/unexpected.
- #2034 diagnostics: 0.39–2.84 GB/s rate estimates across standard lanes, 4.8–36.5 MiB budgets, active admission/probe signals; no 10–28 GB/s ratchet.

The PR body now names this exact SHA and run, resolving the prior review's sole documentation/evidence blocker.

PR #2024, opened 2026-07-13
Rethrow asynchronous completion-source faults

### LINKED ISSUE #2014: P2 bug: PooledCompletionSource.TrySet* catch-all masks genuine completion faults as successful completions (#2002)
## Summary

e4b032c2 (#2002) wrapped every producer completion call site in `PooledCompletionSource.TrySetResult/TrySetException/TrySetCanceled` (`src/Dekaf/Producer/RecordAccumulator.cs:416-480`, used at ReadyBatch success/failure/orphan paths :7370/:7516/:7283, `PartitionBatch` :6484, slow-path worker :2298-2316). Each wrapper catches **all** exceptions and returns `true`.

## Problem

The catch-all exists to contain user continuation exceptions when continuations run inline (transactional mode). But it also swallows genuine completion-source faults — a double-complete race, state corruption, a bug in the source itself — and records them as "inline continuation exception" while **reporting the completion as successful**. In async mode (`RunContinuationsAsynchronously = true`, the default for all non-transactional produces) user code can never be on that stack, so any exception caught there is a library bug being silently masked. A producer could report a message delivered whose completion source actually faulted.

Secondary cost: the try/catch blocks JIT inlining of every completion call — one extra non-inlined call + null check per awaited message (small, but on the hot ack path).

## Fix

- Only swallow when the source is in inline-continuation mode (transactional); rethrow or `Debug.Fail` + fault the source when `RunContinuationsAsynchronously` is true.
- Consider recording distinct diagnostics for "user continuation threw" vs "completion source fault" so the two are never conflated in metrics.

Found while investigating the run [29254448922](https://github.com/thomhurst/Dekaf/actions/runs/29254448922) regressions; filed separately as a correctness/observability item — this masks failures, silent-failure class.

thomhurst:
## Summary
- classify inline user-continuation exceptions separately from genuine completion-source faults
- isolate async completion-source faults per message so sibling completions and cleanup always continue
- report genuine faults through `dekaf.producer.completion_source.faults` and return `false` instead of reporting success
- make producer ordering-test acknowledgements asynchronous to prevent sender-thread reentrant disposal

## Test plan
- net8.0 full unit suite: 5,290 passed
- net10.0 full unit suite: 5,395 passed
- ReadyBatch async source-fault success/failure regression: 2 passed

Closes #2014

### DIFF
--- a/src/Dekaf/Diagnostics/DekafMetrics.cs
+++ b/src/Dekaf/Diagnostics/DekafMetrics.cs
@@ -46,6 +46,12 @@ internal static class DekafMetrics
             unit: "{exception}",
             description: "Exceptions isolated from inline transactional produce continuations.");
 
+    internal static readonly Counter<long> CompletionSourceFaults =
+        DekafDiagnostics.Meter.CreateCounter<long>(
+            "dekaf.producer.completion_source.faults",
+            unit: "{fault}",
+            description: "Producer completion-source faults isolated while completing messages.");
+
     // Consumer metrics
     internal static readonly Counter<long> MessagesReceived =
         DekafDiagnostics.Meter.CreateCounter<long>(
--- a/src/Dekaf/Producer/PooledValueTaskSource.cs
+++ b/src/Dekaf/Producer/PooledValueTaskSource.cs
@@ -34,6 +34,8 @@ public sealed class PooledValueTaskSource<T> : IValueTaskSource<T>
     /// </summary>
     public short Version => _core.Version;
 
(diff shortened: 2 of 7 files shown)


thomhurst:
Addressed the blocking review in 159e795c: async completion faults are now isolated per source, recorded via a distinct metric, and return false so sibling completion loops continue. Added three-source ReadyBatch success/failure regressions. Also fixed the net8 ordering timeout by preventing final acknowledgement continuations from reentering sender disposal.

PR #2025, opened 2026-07-13
fix(producer): repair capacity-probe recovery (#2010)

### LINKED ISSUE #2010: P1 perf: producer throughput decays -2.7%/min; 3conn stuck-low for 4 min — all-decay rate estimator + capacity probe evaluates wrong timescale against wrong statistic (#2001)

thomhurst:
Stacked on #2020 (which is stacked on #2019). Retarget to `main` after both P0 governor fixes merge.

## Summary
- evaluate capacity probes only from batches admitted after probe start
- measure for at least delivery target + linger and compare probe/window averages
- retain demonstrated loaded capacity across transient one-minute stalls
- expose budget, max-rate, and probe success/failure timelines in both stress reporters

## Test plan
- [x] `BrokerUnackedByteBudgetTests` (71/71)
- [x] full unit suite (5,403/5,403)
- [x] producer admission-probe integration test
- [x] Python stress reporter tests (17/17)
- [x] BenchmarkDotNet short run: 14.35 ns/ack, 19.73 ns/pass, 0 B allocated
- [x] 15-minute issue-only FnF: steady/peak 0.954, no throughput trend-gate breach
- [x] stacked 1-minute comparison: 59.4k msg/s vs #2020 control 56.5k msg/s

Local absolute throughput remains far below the issue family's 1.4M target on both this stack and unchanged #2020, so the P0 dual-gate stress validation remains required before the stack merges.

Closes #2010

### DIFF
--- a/.github/scripts/stress_report.py
+++ b/.github/scripts/stress_report.py
@@ -471,6 +471,60 @@ def format_connection_scale_timeline(results, title):
     return lines
 
 
+def format_producer_budget_timeline(results, title):
+    """Correlate producer budget/rate/probe state with throughput intervals."""
+    rows = []
+    for result in results:
+        diagnostics = result.get('producerDeliveryDiagnostics') or {}
+        for sample in diagnostics.get('brokerBudgetSamples') or []:
+            rows.append((result, sample))
+    if not rows:
+        return []
+
+    rows.sort(key=lambda row: row[1].get('capturedAtUtc', ''))
+    lines = [
+        f"## Producer Budget Timeline - {title}",
+        "",
+        "| Client | Sample UTC | Broker | Budget / unacked | Max rate | Probe success/failure | Admission blocks | Nearest throughput sample |",
+        "|--------|------------|-------:|------------------|---------:|----------------------:|-----------------:|---------------------------|",
+    ]
+    for result, sample in rows:
+        sample_time = _parse_timestamp(sample.get('capturedAtUtc'))
+        intervals = result.get('throughput', {}).get('intervalSamples') or []
+        timestamped_intervals = [
+            (interval, _parse_timestamp(interval.get('capturedAtUtc')))
+            for interval in intervals
+        ]
+        timestamped_intervals = [
+            (interval, captured_at)
+            for interval, captured_at in timestamped_intervals
+            if captured_at is not None
+        ]
+        nearest = min(
+            timestamped_intervals,
+            key=lambda item: abs((item[1] - sample_time).total_seconds()),
+            default=(None, None),
+        )[0] if sample_time is not None else None
+        throughput = '-'
+        if nearest is not None:
+            throughput = (
+                f"{nearest.get('elapsedSeconds', 0):.1f}s / "
+                f"{nearest.get('messagesPerSecond', 0):,.0f} msg/s"
+            )
+        lines.append(
+            f"| {result.get('client', 'Unknown')} | {sample.get('capturedAtUtc', '-')} | "
+            f"{sample.get('brokerId', '-')} | "
+            f"{sample.get('budgetBytes', 0) / 1_048_576:.1f} MiB / "
+            f"{sample.get('unackedBytes', 0) / 1_048_576:.1f} MiB | "
+            f"{sample.get('maxRateBytesPerSec', 0) / 1_000_000:.1f} MB/s | "
+            f"{sample.get('capacityProbeSuccessCount', 0):,}/"
+            f"{sample.get('capacityProbeFailureCount', 0):,} | "
+            f"{sample.get('admissionBlockCount', 0):,} | {throughput} |"
+        )
+    lines.append("")
+    return lines
+
+
 def format_latency_outlier_timeline(results, title):
     """Correlate sampled delivery stalls with scaling, throughput, and GC."""
     rows = [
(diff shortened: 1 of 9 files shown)


thomhurst:
Superseded by #2029 after #2020 was squash-merged and its deleted base auto-closed this PR. The #2010 patch was range-diff identical and rebased onto the new #2019 stack head.

PR #2026, opened 2026-07-13
perf(stress): stabilize consumer GC topology

### LINKED ISSUE #2012: P1 perf: consumer decay recurs (-1.61%/min, steady/peak 0.80) on string-deserializing scenarios only; CPU/s flat, Gen2 count matches step count — #1992 closed without fix
## Summary

Stress run [29254448922](https://github.com/thomhurst/Dekaf/actions/runs/29254448922): the consumer decay #1992 described recurs — #1992 was closed by adding diagnostics (#1999), not by fixing the decay. This issue refiles it with sharper attribution from the new run.

- `consumer`: 2.79M → 2.52M msg/s (-10%), slope **-1.61%/min**, steady-state/peak **0.80** (gate fail); CPU μs/msg 0.51 → 0.58.
- `consumer-batch`: 2.96M → 2.62M, steady-state/peak 0.84 (fail).
- `consumer-raw` and `consumer-raw-batch`: **flat and healthy** (slope -0.04 / -0.12%/min).

## Analysis — what the new data pins down

1. **The CPU/msg regression is not new per-message work.** It decomposes almost entirely as constant CPU/s over falling msg/s: consumer -9.7% throughput → predicted CPU/msg 0.51 × 1.107 = 0.565 vs observed 0.58; consumer-batch predicted 0.565 vs observed 0.56. Signature = client-side stall/GC, not added work.
2. **#1999's diagnostics are exonerated.** All four consumer scenarios wire `ConsumerFetchDiagnosticsTracker` identically (verified in all four `Consumer*StressTest.cs`), yet raw scenarios are flat. Library-side additions are per-fetch or per-minute and bounded (reap events capped at 256, `ConnectionPool.cs:1298-1314`; snapshot 1/min). Worst-case metric-listener cost ≈ 0.003 μs/msg — 20× too small.
3. **The differentiator is the string-deserialization/GC layer.** Only the decaying scenarios run `CachingStringDeserializer` (XxHash128 over the 1000B payload + `ConcurrentDictionary` lookups for key and value, `src/Dekaf/Serialization/CachingStringDeserializer.cs:44-87`) and `ConsumeResult` materialization; raw is `<Ignore, ReadOnlyMemory<byte>>` zero-copy.
4. **Gen2 count correlates with step count; allocation volume does not.** GC stats: consumer 24/7/**3** (Gen0/1/2), batch 32/8/**4**, raw 46/2/**1**, raw-batch 177/2/**1**. Raw-batch has 7× consumer's Gen0 count yet is flat. The 3-4 Gen2s match the number of persistent step-downs #1992 observed (~minutes 6 and 9). Alloc/msg ≈ 0 B (cache absorbs replayed seed data) — mid-life promotion and Gen2 timing are the suspects, not churn.
5. **Testable hypothesis:** .NET 10 Server GC DATAS heap-count adaptation re-tunes at those Gen2 points and never re-expands, producing permanent throughput steps.

## Fix plan

1. **A/B the DATAS hypothesis**: one stress run with `DOTNET_GCDynamicAdaptationMode=0` on consumer/consumer-batch. Steps vanish → GC-adaptation interaction, tune GC config or cache behavior; steps persist → dig into `CachingStringDeserializer` promotion behavior (e.g. cache entry lifetime spanning Gen1).
2. **Add per-minute GC telemetry to the fetch timeline** (Gen0/1/2 deltas, pause time, heap size, heap count via `GC.GetGCMemoryInfo`) — `GcStats` is run-total only, so the minute-6/9 ↔ Gen2 correlation can't be confirmed from published data. This is the missing diagnostic, not more fetch metrics.
3. If DATAS is implicated, evaluate pinning heap count or documenting recommended GC settings for high-throughput consumers.

Related: #1992 (original report, closed without fix), #1999 (diagnostics; see companion CI issue — its timeline never reaches published docs).

thomhurst:
## Summary
- pin consumer stress jobs to stable Server GC topology after DATAS A/B tied Gen2 boundaries to throughput steps
- add per-minute GC collection deltas, pause time, heap size/count, and runtime mode to the consumer fetch timeline
- document the production DATAS opt-out and memory tradeoff

## Evidence
- 15-minute baseline, 2 logical processors: 3 Gen2 collections, 170,315 msg/s
- identical run with `DOTNET_GCDynamicAdaptationMode=0`: 0 Gen2 collections, 185,218 msg/s (+8.8%)

## Test plan
- [x] `dotnet build tests/Dekaf.Tests.Unit --configuration Release --no-restore`
- [x] focused `ConsumerFetchDiagnosticsTests` (9/9)
- [x] full net10 unit suite (5,392/5,392)
- [x] `npm run build --prefix docs`

Closes #2012

### DIFF
--- a/.github/workflows/stress-tests.yml
+++ b/.github/workflows/stress-tests.yml
@@ -128,6 +128,11 @@ jobs:
     env:
       BROKER_CPUS: 0-5
       CLIENT_CPUS: 6,7
+      # .NET 10 DATAS can shrink Server GC heap count at Gen2 boundaries on the
+      # two-core consumer runner, creating permanent throughput steps unrelated to
+      # either Kafka client. Pin consumer jobs to stable Server GC topology so paired
+      # comparisons measure client work; producer jobs retain the runtime default.
+      DOTNET_GCDynamicAdaptationMode: ${{ startsWith(matrix.scenario, 'consumer') && '0' || '1' }}
       # Broker log dirs live on tmpfs so disk I/O never caps ingestion. Sized above the
       # worst-case retention overshoot (see KafkaEnvironment.RetentionConfig for the
       # math); a fill halts the broker, so the disk monitor below samples usage during
--- a/docs/docs/performance.md
+++ b/docs/docs/performance.md
@@ -178,6 +178,19 @@ await foreach (var batch in consumer.ConsumeAsync(cts.Token).Batch(100))
 }
 ```
 
+For sustained CPU-bound consumers on .NET 10 Server GC, watch throughput and GC
+telemetry over time. If permanent throughput steps line up with Gen2 collections and
+a changing GC heap count, benchmark with dynamic GC adaptation disabled:
+
+```bash
+DOTNET_GCDynamicAdaptationMode=0 dotnet MyConsumer.dll
+```
+
+Set the variable before process startup. Disabling DATAS keeps the Server GC heap
+topology stable, which can improve sustained throughput on tightly CPU-pinned
+workloads, but it may retain more managed memory. Keep the runtime default unless an
+A/B test of your production-shaped workload shows the same Gen2-correlated decay.
+
 #### Low Latency
 
 ```csharp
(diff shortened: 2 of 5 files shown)


thomhurst:
Addressed the blocking review in 156278ab: restored the shared MeterListener isolation key on both diagnostics tests, avoiding unnecessary whole-suite serialization while retaining cross-test protection. Validation: net10.0 unit project builds clean; ConsumerFetchDiagnosticsTests 9/9 pass.

PR #2028, opened 2026-07-13
Ratchet BatchArena pool after sustained misses

### LINKED ISSUE #2017: P2 perf: BatchArena POH pool (128 slots) thrashes under multi-broker budget oscillation — 16-26 Gen2/15min on 3-broker producers
## Summary

Stress run [29254448922](https://github.com/thomhurst/Dekaf/actions/runs/29254448922): 3-broker producer scenarios show **16-26 Gen2 collections** in 15 minutes (acks-all 3b: 26; idempotent 3b: 16-19) vs 5-6 on the 1-broker equivalents, with total allocation fine (~2 B/msg). The churn is large pinned buffers, not per-message garbage.

## Root cause

Sealed-unacked batches each pin a 1 MiB `BatchArena` POH buffer until ack (retry needs the data; released at pipeline exit, `src/Dekaf/Producer/RecordAccumulator.cs:4453-4462`). The arena pool is static and capped at **128 slots** for the default 256 MB/1 MiB stress config (`ComputePoolSize`: 256/4 = 64 → clamped to `DefaultPoolSize` 128, `RecordAccumulator.cs:2125-2134`), and pooled `PartitionBatch` instances each retain an arena too (:640-641).

- 1 broker: in-flight arenas ≈ rate × 10ms ÷ 1 MiB ≈ a handful — well inside the pool.
- 3 brokers with the budget ratcheted to its 32 MiB/broker cap (see the companion 3-broker latency-loop issue): ~96 arenas standing in flight + pooled-batch retention + open batches — circulation exceeds 128. Budget oscillations (min-RTT probes drain the queue for 50-250 ms/broker; capacity probes swing ±25-50%) repeatedly swing in-flight count by tens of arenas: up-swings miss the pool → fresh `GC.AllocateUninitializedArray(pinned: true)` 1 MiB allocations (`RecordAccumulator.cs:742, 763-764`); down-swings hit a full pool → buffers dropped (:781-787). POH garbage is reclaimable only by Gen2 → the GC schedules Gen2s to keep up.

Fixing the budget ratchet shrinks the standing set back to BDP and removes the cause; this issue tracks independent hardening so the pool can't thrash whenever in-flight legitimately exceeds 128 (large clusters, high-BDP links).

## Fix

- Ratchet `BatchArena` pool size on sustained misses (same pattern as `MaybeRatchetPoolCapacity`, `RecordAccumulator.cs:4032-4049`), or
- Size the pool by broker count × (per-broker cap ÷ batch size) at producer construction, bounded by BufferMemory.

Either way, add a pool miss/drop counter to producer diagnostics so thrash is observable.

thomhurst:
## Summary
- grow the process-wide BatchArena pool after each sustained 128-miss interval
- cap dynamic growth at 512 arenas
- expose process-wide arena misses, drops, and capacity in producer delivery diagnostics

## Validation
- cross-target unit-test project build (net8.0, net10.0)
- focused producer tests: 105 passed
- full net10.0 unit suite: 5,396 passed; one unrelated BrokerSender ordering test timed out, then passed alone in 337ms

Closes #2017

### DIFF
--- a/src/Dekaf/Producer/IKafkaProducer.cs
+++ b/src/Dekaf/Producer/IKafkaProducer.cs
@@ -268,6 +268,9 @@ internal sealed class ProducerDeliveryDiagnosticsSnapshot
     public long ProduceRequestCount { get; init; }
     public double ProduceRequestElapsedSeconds { get; init; }
     public double ProduceRequestsPerSecond { get; init; }
+    public long BatchArenaPoolMisses { get; init; }
+    public long BatchArenaPoolDrops { get; init; }
+    public int BatchArenaPoolCapacity { get; init; }
     public List<ProducerBrokerRequestDiagnostic> BrokerProduceRequests { get; init; } = [];
     public List<ProducerCoalesceWidthDiagnostic> CoalesceWidthHistogram { get; init; } = [];
     public List<ProducerBatchDeliveryDiagnostic> Batches { get; init; } = [];
--- a/src/Dekaf/Producer/RecordAccumulator.cs
+++ b/src/Dekaf/Producer/RecordAccumulator.cs
@@ -650,8 +650,11 @@ internal sealed class BatchArena
     // With 256KB batches and 256MB buffer: 512 × ~280KB ≈ 140MB POH retention.
     // With 1MB batches: ComputePoolSize returns 128 (not 512), so 128 × ~1.1MB ≈ ~140MB.
     internal const int MaxPoolSizeCap = 512;
+    internal const long MissRatchetThreshold = 128;
     private static int s_maxPoolSize = DefaultPoolSize;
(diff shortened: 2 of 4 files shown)


thomhurst:
Addressed the blocking shared-state race in 111a85a3: the diagnostics test now brackets each monotonic process-wide value with before/live bounds instead of comparing a snapshot to a later exact read. Validation: clean net10.0 build; focused producer tests 105/105.

I did not add a test that resets/drives the live static pool: deterministic setup would require mutating global pool/counter state shared by the parallel suite, recreating the isolation defect this review identified. The threshold/cap arithmetic remains covered through the deterministic helper; the production miss-path wiring is the direct Interlocked.Increment → MaybeRatchetPoolSize call.

PR #2038, opened 2026-07-13
fix(producer): bound delivery epoch rates

thomhurst:
## Summary\n- measure cumulative delivery over the full send epoch\n- carry window-average rate between probe rungs\n- skip capacity probes while the admission gate has no demand\n\n## Validation\n- 5,424 net10.0 unit tests pass\n- BrokerUnackedByteBudget benchmarks: 15.21 ns / 18.68 ns, 0 B allocated\n\nCloses #2034

### DIFF
--- a/src/Dekaf/Producer/BrokerUnackedByteBudget.cs
+++ b/src/Dekaf/Producer/BrokerUnackedByteBudget.cs
@@ -53,6 +53,7 @@ internal sealed class BrokerUnackedByteBudget
     internal readonly record struct DeliverySnapshot(
         long DeliveredBytes,
         long DeliveredTimestamp,
+        long DeliveryEpochFirstSendTimestamp,
         long SendTimestamp,
         bool AppLimited,
         long OldestBatchTimestamp,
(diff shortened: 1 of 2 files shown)


thomhurst:
Addressed the blocking CAS race in c685850c: the winner now re-reads _totalDeliveredBytes after acquiring the epoch sentinel, so it cannot publish an older marker than the CAS observed. Added an 8-reader concurrent monotonic-snapshot regression test, updated the synchronization contract docs, and named the probe-demand fraction. Focused budget tests pass 77/77. Full suite exposed unrelated pool-identity test issue #2040; no rerun used to mask it.

thomhurst:
Fixed the blocking stale-publisher anchor race in 94f1df0b. Post-CAS publication now re-reads delivered bytes and preserves the existing first-send anchor when the publication is redundant; only a genuinely advanced epoch installs a new anchor. Added deterministic regression coverage. net10 build clean; BrokerUnackedByteBudgetTests 78/78.

PR #2090, opened 2026-07-14
Cover auto-commit retry and stability gates

### LINKED ISSUE #2066: AutoCommitLoopAsync resilience untested: failure retry path, Stable-state gate, 3-attempt close commit
## Context

The background auto-commit loop (`KafkaConsumer.AutoCommitLoopAsync`) has three behaviors with no direct coverage:

1. **Failure retry**: on commit exception it logs, delays 200ms, retries once, then swallows until the next interval. `ConsumerDirtyCommitTests.CommitAsync_WhenCommitFails_PreservesDirtyOffsetsForRetry` covers the manual path only; the loop's own retry/swallow behavior (and that a persistent failure never crashes the loop or the consumer) is untested.
2. **Stable gate**: the loop skips commits unless `_coordinator.State == CoordinatorState.Stable`, which is what prevents committing with a stale generation mid-rebalance. Untested.
3. **Close commit retries**: `CloseAsyncCore` attempts the final flush+commit up to 3 times. Only the single-attempt happy path is exercised (via `CloseAsync_UnknownCoordinator_RediscoversAndCommitsPendingOffset`).

Unit-testing 1 and 2 needs either the mock-connection consumer from `ConsumerDirtyCommitTests` plus a way to drive coordinator state, or a small internal seam. Worth doing when touching this area; the dirty-offset preservation semantics that make the retry *correct* are already covered.

thomhurst:
## Summary
- extract one private auto-commit cycle from the timer loop without changing retry behavior
- cover Stable coordinator gating and retry-once success
- verify two failed attempts are swallowed and the next cycle retries preserved dirty offsets
- verify close commit succeeds on its third allowed attempt

## Validation
- Release build net10.0 + net8.0: zero warnings
- ConsumerDirtyCommitTests: 17/17 per framework
- full unit suites: net10.0 5475/5475; net8.0 5368/5368

Closes #2066

### DIFF
--- a/src/Dekaf/Consumer/KafkaConsumer.cs
+++ b/src/Dekaf/Consumer/KafkaConsumer.cs
@@ -6995,6 +6995,10 @@ public sealed partial class KafkaConsumer<TKey, TValue> :
         {
             await CommitStoredOffsetsAsync(cancellationToken).ConfigureAwait(false);
         }
+        catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+        {
+            throw;
+        }
         catch (Exception ex)
         {
             LogAutoCommitFailed(ex);
--- a/src/Dekaf/Consumer/KafkaConsumer.cs
+++ b/src/Dekaf/Consumer/KafkaConsumer.cs
@@ -6976,27 +6976,34 @@ public sealed partial class KafkaConsumer<TKey, TValue> :
             try
             {
                 await Task.Delay(_options.AutoCommitIntervalMs, cancellationToken).ConfigureAwait(false);
-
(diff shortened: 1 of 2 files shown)


thomhurst:
Addressed the blocking cancellation finding in d9833dd6. TryAutoCommitAsync now rethrows token cancellation before the warning/retry handler, restoring silent shutdown behavior. Added a transport-level regression that cancels during OffsetCommit and asserts propagation plus exactly one request. ConsumerDirtyCommitTests pass 18/18 on net10.0 and net8.0; full unit suites pass 5476/5476 and 5369/5369.

PR #2095, opened 2026-07-14
Add callback for filtered consumer messages

### LINKED ISSUE #2068: Consumer LINQ filters silently drop skipped messages once commits advance past them — document and add defer-safe affordance
## Summary

The consumer LINQ extensions (`Where`, `SkipWhile`, `TakeWhile` in `src/Dekaf/Consumer/ConsumerExtensions.cs`) operate downstream of `ConsumeAsync`. By the time the predicate runs, the consumer has already yielded the message and advanced its position. Any commit — auto-commit (the default, `OffsetCommitMode.Auto`), `CommitAsync()`, or `StoreOffset(result)` + `CommitAsync()` for a later message — advances the committed watermark **past** every message the predicate skipped. Those messages are never redelivered to the group.

For a **deterministic** predicate (skip tombstones, skip other tenants) this is correct and expected: redelivery would just re-filter the same messages. That is what filtering means.

## The hazard

For a **stateful or temporal** predicate, users often intend *defer*, not *drop*:

```csharp
// Intent: "don't process while the downstream is unhealthy, pick these up later"
await foreach (var msg in consumer.ConsumeAsync(ct)
    .Where(_ => downstream.IsHealthy))   // ← messages skipped while unhealthy are GONE
{
    await ProcessAsync(msg);
    consumer.StoreOffset(msg);
    await consumer.CommitAsync();         // commits past the skipped offsets
}
```

Kafka commits are watermarks, not per-message acks — there is no way to commit offset N while "keeping" a skipped offset < N. Once the watermark passes a skipped message, at-least-once delivery no longer holds for it. `Where` is a drop, not a defer, and nothing in the API or docs currently says so.

Aggravating factors:

- **Auto-commit is the default.** Filtered offsets are committed past even if the app never commits manually — the consumer position advanced when the message was *consumed*, not when it was *processed*.
- Our own docs (`docs/docs/consumer/linq-extensions.md`) showcase stateful predicates (`SkipWhile(m => m.Value != "START")`, random 1% sampling) without any warning.
- `TakeWhile` has a subtle variant: the first message that fails the predicate has already been pulled from the source (position advanced) but is never yielded — auto-commit then commits past it too.

## Proposed fix

Three parts, first one is doc-only:

1. **Documentation (PR incoming, part of this issue):** add an explicit warning to `docs/docs/consumer/linq-extensions.md` and to the XML `<remarks>` of `Where`/`SkipWhile`/`TakeWhile`: filtering is for messages you want to *permanently ignore*; skipped messages are gone once any commit advances past them; never use stateful/temporal predicates as a deferral mechanism — use `Pause`/`Seek` or a retry/DLQ topic instead.

2. **Defer-safe affordance:** add a `Where` overload that surfaces skipped messages so callers can park them (e.g. produce to a retry topic) *before* the commit watermark passes them:

   ```csharp
   public static IAsyncEnumerable<ConsumeResult<TKey, TValue>> Where<TKey, TValue>(
       this IAsyncEnumerable<ConsumeResult<TKey, TValue>> source,
       Func<ConsumeResult<TKey, TValue>, bool> predicate,
       Func<ConsumeResult<TKey, TValue>, ValueTask> onFiltered,   // await before continuing
       CancellationToken cancellationToken = default)
   ```

   The callback is awaited before enumeration continues, so a caller that produces the skipped message to a parking topic (and awaits delivery) is guaranteed the message is safe before any later offset can be committed. A sync `Action<ConsumeResult<TKey, TValue>>` overload can accompany it for cheap cases (metrics, logging).

3. **Observability (optional):** a counter metric (e.g. `dekaf.consumer.filtered_messages`) incremented by the filtering operators, so silent drops show up in dashboards rather than as mysteriously missing data.

Part 2 is the substantive fix: it makes the safe pattern (park-then-commit) expressible in one place instead of forcing users to abandon the LINQ operators entirely when they need deferral.

## References

- `src/Dekaf/Consumer/ConsumerExtensions.cs` — `Where` (line ~19), `SkipWhile` (~142), `TakeWhile` (~172)
- `src/Dekaf/Consumer/KafkaConsumer.cs` — `StoreOffset(result)` stores `result.Offset + 1` (~3980); auto-commit publishes consumed position (`PublishActiveConsumedPosition`, ~3106)
- `docs/docs/consumer/linq-extensions.md` — examples using stateful predicates with no warning

thomhurst:
## Summary
- add a Where overload that awaits an asynchronous callback for rejected records
- document the manual-commit boundary required for crash-safe parking
- cover callback selection, ordering, failure propagation, and argument validation

## Validation
- ConsumerExtensionsTests: 24/24 on net10.0 and net8.0
- full unit suite: 5476/5476 on net10.0
- full unit suite: 5369/5369 on net8.0

Closes #2068

### DIFF
--- a/docs/docs/consumer/linq-extensions.md
+++ b/docs/docs/consumer/linq-extensions.md
@@ -67,6 +67,34 @@ await foreach (var msg in consumer.ConsumeAsync(ct)
 
 Filtered-out messages are skipped permanently (see [warning above](#filtering-skips-messages-permanently)). Keep predicates deterministic: a message should be filtered because of *what it is*, not because of *when it arrived* or the current state of your application.
 
+### Handling Filtered Messages Before Continuing
+
+`Where` has an overload that awaits an `onFiltered` callback before requesting the next message. Use it to record, inspect, or park rejected messages:
+
+```csharp
+// Build this consumer with OffsetCommitMode.Manual.
+await foreach (var msg in consumer.ConsumeAsync(ct).Where(
+    CanProcess,
+    async filtered =>
+    {
+        await ParkOnRetryTopicAsync(filtered, ct);
+    },
+    ct))
+{
+    await ProcessAsync(msg);
+    consumer.StoreOffset(msg);
+    await consumer.CommitAsync(ct);
+}
+```
+
+The callback must complete before enumeration continues, and a callback failure stops enumeration. This prevents a later manually committed offset from passing a rejected message before it has been parked.
+
+:::warning Manual commit is required for crash-safe parking
+
+The callback cannot suspend a consumer's background periodic auto-commit. If a rejected message must be parked durably before **any** commit advances past it, configure `OffsetCommitMode.Manual` and commit only after every earlier message has been processed or parked.
+
+:::
+
 ## Select - Transforming Messages
 
 Project messages to a different type:
--- a/src/Dekaf/Consumer/ConsumerExtensions.cs
+++ b/src/Dekaf/Consumer/ConsumerExtensions.cs
@@ -41,6 +41,45 @@ public static class ConsumerExtensions
     }
 
     /// <summary>
+    /// Filters consumed messages and invokes an asynchronous callback for each filtered message.
+    /// </summary>
+    /// <typeparam name="TKey">Key type.</typeparam>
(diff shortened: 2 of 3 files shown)


thomhurst:
@codex review

PR #2104, opened 2026-07-15
chore(deps): update Polyfill to 11.0.1

thomhurst:
## Summary

- update Polyfill from 11.0.0 to 11.0.1
- remove the `EncodingCompat` workaround
- use `Encoding.UTF8` span APIs directly and add required framework imports

## Why

Polyfill 11.0.1 removes the need for the compatibility wrapper that guarded empty spans on `netstandard2.0`. Call sites can now use the standard UTF-8 span APIs consistently across target frameworks.

## Impact

Encoding behavior remains unchanged while compatibility code and its extra branches are removed.

## Validation

- `git diff --cached --check`
- Build and tests not run, per request


thomhurst:
Addressed blocking review with current package evidence:

- NuGet now publishes signed `Polyfill` 11.0.1 (published 2026-07-15); restore succeeds.
- Compared cached 11.0.0 and 11.0.1 netstandard2.0 source. 11.0.1 adds explicit empty-input guards to `Encoding.GetBytes(ReadOnlySpan<char>, Span<byte>)`, `Encoding.GetByteCount(ReadOnlySpan<char>)`, and `Encoding.GetString(ReadOnlySpan<byte>)`; it also safely substitutes a one-byte stack buffer when the destination span is empty.
- `dotnet build Dekaf.sln --configuration Release` passed with 0 warnings/errors, including `src/Dekaf` netstandard2.0.
- Full unit suites passed: net10 5,544/5,544; net8 5,437/5,437.

The original `NU1102` was publication timing: CI ran before 11.0.1 reached NuGet. Commit 78c66588 triggers fresh checks now that package exists.

PR #2121, opened 2026-07-15
Fix consumer Gen2 collections under memory pressure

### LINKED ISSUE #2112: Consumer stress: 444 Gen2 / 1791 Gen1 collections in 15 min vs Confluent 0 / 336, despite lower alloc/msg
## Summary

Consumer stress in run [29379808486](https://github.com/thomhurst/Dekaf/actions/runs/29379808486) (1 broker, 15 min, 1.83M msg/s): Dekaf logged **444 Gen2 and 1,791 Gen1 collections vs Confluent's 0 Gen2 / 336 Gen1** — a full blocking GC roughly every 2 seconds for the whole run.

| | gen0 | gen1 | gen2 | alloc B/msg | throughput | CPU us/msg |
|---|---|---|---|---|---|---|
| Dekaf | 80,146 | 1,791 | **444** | 2,027 | 1.83M/s (1.51x) | 0.71 (0.65x) |
| Confluent | 64,665 | 336 | 0 | 2,440 | 1.21M/s | 1.09 |

Dekaf wins throughput, CPU/msg, and even allocates *less* per message than Confluent — yet its allocations are surviving to Gen2 while Confluent's die in Gen0. The problem is object **lifetime/size**, not volume: either LOH-sized buffers (anything ≥ 85 KB allocates straight to LOH and only collects at Gen2) or mid-life objects held across enough fetches to get promoted.

Prior art: #1945 tracked consumer LOH with Gen2 = 1,211 in the 07-12 era; whatever landed since roughly halved it (444), but the leak-shaped pattern remains. The raw-consumer scenarios (`consumer-raw`, `consumer-raw-batch`) show ~0 alloc/msg and 3/16 gen0 — so the churn lives in the deserializing consumer path or the stress harness's message handling, not the fetch/network layer.

## Why it matters

At 444 Gen2/15 min the consumer is one config tweak (larger fetch sizes, server GC heap growth) away from visible pause-time impact, and it burns budget the otherwise-winning CPU numbers hide. It is also the last standing item from the "only real Confluent gap is consumer CPU/msg" audit — CPU/msg now wins, GC shape doesn't.

## Ask

1. Capture a `dotnet-gcdump`/`dotnet-counters` (`loh-size`, `gen-2-gc-count`) session against a local consumer stress run and identify the surviving type(s) — expected suspects: fetch response buffers ≥ 85 KB escaping the pool, per-batch arrays promoted while awaiting processing, or string materialization in the harness's verification path.
2. If LOH: pool or slice the buffer below 85 KB; if promotion: shorten the hold window or pool the survivor.

Data: `stress-test-results-20260715-015956.json` (`consumer` 1b rows) in the run's `stress-test-results-merged` artifact.

thomhurst:
Fixes #2112

## Root cause

Live-heap and EventPipe GC profiling showed value strings dying in Gen0 and only ~9 MB retained in Gen2. Full collections were `InducedLowMemory`, not large managed allocations. The native response pool retained dormant unmanaged buffers independently in every adaptive capacity bucket and never trimmed them under process memory pressure.

## Change

- bound native response-buffer retention globally across capacity buckets, using the total consumer pipeline working set
- keep the native global allowance distinct from the managed ArrayPool per-bucket depth
- sample runtime memory load at most once per second on native returns (demand-driven: quiet pools stay bounded and trim on their next return, avoiding a timer per shared pool)
- trim dormant native buffers and suspend retention while runtime high-memory pressure is active
- preserve `netstandard2.0` compatibility where memory-load telemetry is unavailable
- cover cross-bucket bounds, trimming, thresholds, pressure transitions, and recovery

## Validation

- `dotnet build Dekaf.sln -c Release --no-restore`: 0 warnings, 0 errors
- net10 unit tests: 5551 passed
- net8 unit tests: 5444 passed
- focused response-pool tests: 29 passed
- one-minute local consumer stress under high host memory load: Gen2 reduced from 3 baseline to 2; allocation unchanged at 1.98 KB/msg; 71,911 msg/s

### DIFF
--- a/src/Dekaf/Consumer/KafkaConsumer.cs
+++ b/src/Dekaf/Consumer/KafkaConsumer.cs
@@ -1115,12 +1115,7 @@ public sealed partial class KafkaConsumer<TKey, TValue> :
             },
             loggerFactory,
             connectionsPerBroker: options.ConnectionsPerBroker,
-            ResponseBufferPool.Create(
-                CalculateMaximumFetchResponsePayloadBytes(options),
-                PoolSizing.ForConsumerResponseBuffers(
-                    options.BootstrapServers.Count,
-                    options.PrefetchPipelineDepth,
-                    options.MaxConnectionsPerBroker)),
+            CreateResponseBufferPool(options),
             telemetryMetricCollector: telemetryMetricCollector);
 
         var metadataManager = new MetadataManager(
@@ -1130,6 +1125,18 @@ public sealed partial class KafkaConsumer<TKey, TValue> :
             logger: loggerFactory?.CreateLogger<MetadataManager>());
 
         return (connectionPool, metadataManager, telemetryMetricCollector);
+    }
+
+    private static ResponseBufferPool CreateResponseBufferPool(ConsumerOptions options)
+    {
+        var responseBufferWorkingSet = PoolSizing.ForConsumerResponseBuffers(
+            options.BootstrapServers.Count,
+            options.PrefetchPipelineDepth,
+            options.MaxConnectionsPerBroker);
+        return ResponseBufferPool.Create(
+            CalculateMaximumFetchResponsePayloadBytes(options),
+            managedArraysPerBucket: responseBufferWorkingSet,
+            maxRetainedNativeBuffers: responseBufferWorkingSet);
     }
 
     internal static int CalculateMaximumFetchResponsePayloadBytes(ConsumerOptions options)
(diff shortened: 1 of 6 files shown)


thomhurst:
15-minute paired stress validation dispatched: https://github.com/thomhurst/Dekaf/actions/runs/29395225089

thomhurst:
Addressed all blocking review findings in e6752ef9: separated managed per-bucket and native global retention limits/cache keys; documented that ForConsumerResponseBuffers computes the total live pipeline working set used as the global native allowance; tested independent limits; documented demand-driven quiet-pool trimming. Full net10 5551/5551 and net8 5444/5444 pass; solution build is clean. The earlier net8 CI failure was an unrelated producer-test timeout and passes locally. @claude review

Issue #2109, opened 2026-07-15
3-broker producer max delivery latency spikes to 292–919 ms (Confluent ≤96 ms) while p99 stays ≤48 ms


thomhurst:
## Summary

Stress run [29379808486](https://github.com/thomhurst/Dekaf/actions/runs/29379808486) (main @ 2c97bc71): on 3-broker producer scenarios Dekaf's **max** delivery latency spikes to 292–919 ms while p99 stays ≤ 48 ms. Confluent's max on the same scenarios never exceeds 96 ms. These are rare multi-hundred-millisecond stall events, not a fat tail.

| Scenario (3 brokers) | Dekaf p99 | Dekaf max | Confluent max |
|---|---|---|---|
| producer | 39.0 ms | **292 ms** | 83 ms |
| producer (3conn) | 42.4 ms | **302 ms** | — |
| acks-all | 33.6 ms | **463 ms** | 96 ms |
| idempotent | 44.4 ms | **592 ms** | 94 ms |
| idempotent (3conn) | 48.2 ms | **919 ms** | — |

1-broker scenarios are much tamer (92–252 ms max), so the trigger correlates with multi-broker coordination.

## Suspects (unconfirmed — diagnostics gap)

- **Min-RTT probe** (`BrokerUnackedByteBudget.StartMinRttProbe`): the probe shrinks the horizon to `MinRttProbeSafetyMultiplier × minRtt` (1.0x), which stalls admission until the queue drains; messages already queued behind a deep budget at probe start wear the full drain time.
- Metadata refresh / leader-change pauses on one broker stalling batches routed to it.
- GC: producer runs show gen2 = 3 per 15 min — three gen2 pauses could line up with three spikes, but 592–919 ms would be an extreme pause for these heap sizes.
- Capacity probe interaction (`ProbeBudgetMultiplier` overshoot then decay).

## Ask

1. Add stall visibility to `producerDeliveryDiagnostics`: timestamps of min-RTT/capacity probe windows, plus an admission-block-duration log2 histogram (the request-RTT histogram already exists; blocked-admission time does not). `connectionScaleEvents`/`connectionReapEvents` are already captured — probe events are the missing sibling.
2. With one run of that data, correlate spike timestamps against probe windows and GC pause samples, then fix the actual trigger.

Worth noting: despite the spikes, Dekaf's max on 1b plain (92 ms) beats Confluent (148 ms) — this is specifically a 3-broker pattern.

[...]

PR #2124, opened 2026-07-15
Trace producer admission stalls and probe windows

thomhurst:
## Summary
- add bounded min-RTT and capacity-probe event windows to producer delivery diagnostics
- record contiguous broker admission-block durations in a log2 histogram
- correlate sampled latency outliers with probe windows, connection transitions, throughput, and GC in both stress reporters
- mark probe overlap as temporal correlation only

## Evidence
A local 1-minute 3-broker idempotent run emitted 176 probe events and 1,198 admission-block episodes. Most sampled late outliers overlapped frequent capacity probes, but large no-probe outliers also occurred; this PR therefore adds the missing visibility without claiming a causal fix.

## Validation
- `dotnet build Dekaf.sln -c Release --no-restore` (0 warnings/errors)
- net10 unit suite: 5,546 passed
- net8 unit suite: 5,439 passed
- Python stress-report suite: 26 passed
- local 3-broker producer diagnostic run: 3,752,942 delivered, 0 errors

Refs #2109

### DIFF
--- a/.github/scripts/stress_report.py
+++ b/.github/scripts/stress_report.py
@@ -804,6 +804,74 @@ def format_producer_budget_timeline(results, title):
     return lines
 
 
+def format_budget_probe_timeline(results, title):
+    """Show bounded per-broker min-RTT and capacity probe state transitions."""
+    rows = [
+        (result, event)
+        for result in results
+        for event in (result.get('producerDeliveryDiagnostics') or {}).get('budgetProbeEvents') or []
+    ]
+    if not rows:
+        return []
+
+    rows.sort(key=lambda row: row[1].get('occurredAtUtc', ''))
+    displayed_rows = _sample_evenly(rows, MAX_TIMELINE_ROWS)
+    lines = [
+        f"## Producer Budget Probe Events - {title}",
+        "",
+        "| Client | UTC | Broker | Probe | Outcome | Duration | Budget / unacked |",
+        "|--------|-----|-------:|-------|---------|---------:|------------------|",
+    ]
+    for result, event in displayed_rows:
+        lines.append(
+            f"| {result.get('client', 'Unknown')} | {event.get('occurredAtUtc', '-')} | "
+            f"{event.get('brokerId', '-')} | {event.get('probeType', '-')} | "
+            f"{event.get('outcome', '-')} | {event.get('durationMilliseconds', 0):,}ms | "
+            f"{event.get('budgetBytes', 0) / 1_048_576:.1f} MiB / "
+            f"{event.get('unackedBytes', 0) / 1_048_576:.1f} MiB |"
+        )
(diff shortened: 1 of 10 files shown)


thomhurst:
Addressed the blocking review and CI findings in 20feb488:

- empty `BudgetProbeEvents` now returns instead of recursively re-entering the reporter; regression verifies admission histograms still render without probe events
- disabled admission diagnostics now avoid both `Stopwatch.GetTimestamp()` and atomic tracking on the append fast path
- `RecordAdmissionAvailable` exits before `Interlocked.Exchange` when diagnostics are disabled
- min-RTT probe duration assertion accepts the legitimate 50–51ms truncation range across platform `Stopwatch.Frequency` values

Validation:
- full solution build: 0 warnings/errors
- net10.0 unit suite: 5,547 passed
- net8.0 unit suite: 5,440 passed
- changed diagnostic/reporting classes: 93 passed on net10.0, 93 passed on net8.0

Three unrelated timing/concurrency tests flaked in initial full runs; each passed immediately in isolated reruns, and both subsequent full framework runs were clean.

thomhurst:
Addressed both blocking follow-up findings in 97dad2e7:

- `IsBrokerAdmissionBlocked` now passes its already-computed monotonic `nowTicks` into `RecordAdmissionBlock`, avoiding a redundant timestamp read and preserving the decision instant
- Python admission histogram rendering derives the overflow bucket from each exported histogram length rather than hardcoding bucket 23
- added a short-histogram regression proving the top bucket renders as unbounded; updated the existing fixture to the production 24-bucket shape

Validation:
- solution build: 0 warnings/errors
- net10.0 unit suite: 5,547 passed
- net8.0 unit suite: 5,440 passed
- Python script suites: 72 passed

thomhurst:
Addressed all four blocking diagnostics findings in f34078fe:

- removed every appender-side `RecordAdmissionAvailable` call; appenders only open intervals, while the owning send-loop ack/cap pass privately closes them after budget recomputation
- `ResetDiagnostics` now CAS-reanchors an open episode at the reset boundary instead of erasing it; the lifetime admission counter remains intact because the latency controller consumes it
- periodic broker samples always include the O(1) `CurrentAdmissionBlockMicros` scalar while still omitting histogram arrays
- C# reporter now derives each overflow bucket from the received histogram length, matching Python/historical JSON behavior

Regression coverage adds reset-boundary episode preservation, rolling open-duration sampling, and short-array C# overflow rendering.

Validation:
- solution build: 0 warnings/errors
- affected C# classes: 113 passed
- net10.0 full suite: 5,549 passed
- net8.0 full suite: 5,442 passed
- Python suites: 72 passed

Two unrelated timing tests flaked during initial full runs; both passed isolated and the subsequent full net10.0 run was clean.

PR #2138, opened 2026-07-15
Lower bootstrap receive failures to Debug

### LINKED ISSUE #2130: Receive-loop connection reset before ApiVersions negotiation completes is logged at Error
## Observed

Run [29405601673](https://github.com/thomhurst/Dekaf/actions/runs/29405601673/job/87320184992): during broker startup, `KafkaConnection` logs

```
Error       Dekaf.Networking.KafkaConnection[1338499366] Error in receive loop
System.Net.Sockets.SocketException (104): Connection reset by peer
   at ... ResponseFrameReader.ReadSourceAsync ... KafkaConnection.ReceiveLoopAsync (KafkaConnection.cs:1799)
```

repeatedly (one per connection attempt) while the broker is still in `STARTING`. A reset on a connection that never completed ApiVersions negotiation is expected bootstrap churn, not a client error — Error severity misleads log readers and alerting.

## Proposed

Track whether the connection completed version negotiation (or served at least one successful response). Receive-loop termination via reset/EOF **before** that point → log at Warning or Debug; after it → keep Error. This composes with #2129 (metadata-init noise) but is a separate code path: `KafkaConnection.ReceiveLoopAsync` (src/Dekaf/Networking/KafkaConnection.cs:1799).

thomhurst:
Closes #2130.

## Summary
- track whether a connection has completed any matched response
- log receive-loop failures before that point at Debug because broker startup resets are expected bootstrap churn
- retain Error severity after the connection has served a response

## Validation
- TDD: pre-response case failed before implementation; post-response Error case passed
- KafkaConnectionTests: 46/46 on net8.0 and 46/46 on net10.0
- focused regression tests pass on net8.0 and net10.0 after simplification

### DIFF
--- a/src/Dekaf/Networking/KafkaConnection.cs
+++ b/src/Dekaf/Networking/KafkaConnection.cs
@@ -149,6 +149,7 @@ public sealed partial class KafkaConnection :
     private Task? _disposeTask;
     private int _disposed;
     private int _connected;
+    private bool _hasReceivedResponse;
     private long _lastUsedTimestampMs = Dekaf.MonotonicClock.GetMilliseconds();
     private readonly SemaphoreSlim _connectLock = new(1, 1);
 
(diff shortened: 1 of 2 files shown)


thomhurst:
Addressed the blocking late-response case in `ef3564e5`. A successfully framed response now marks the connection established when it matches either a pending request (including a completion race) or the cancelled-correlation tracker. Added a socket-level regression: cancel request, receive valid late response, then inject receive failure; failure logs Error. `KafkaConnectionTests`: 47/47 on net10 and net8.

Issue #2108, opened 2026-07-15
Producer p50 delivery latency pins at DeliveryLatencyTargetMs under saturation — budget horizon treats latency target as floor, not cap


thomhurst:
## Summary

In stress run [29379808486](https://github.com/thomhurst/Dekaf/actions/runs/29379808486) (main @ 2c97bc71, first run on the post-#2097 pinned-1-conn fairness config), Dekaf wins throughput and CPU in every paired producer scenario but **loses p50 delivery latency in all of them**. Where the broker budget fully saturates (plain / acks-all, 1 broker), p50 pins at **10.95–11.25 ms — i.e. at `DeliveryLatencyTargetMs` (10 ms) plus ~1 ms of serving RTT, overshooting the documented bound**.

| Scenario | Dekaf p50 (ms) | Confluent p50 (ms) | Dekaf p95/p99 | Confluent p95/p99 | Throughput ratio |
|---|---|---|---|---|---|
| producer 1b | 10.95–11.25 | 5.05–9.25 | 15.6 / 25.0 | 42.3 / 73.6 | 1.20x |
| acks-all 1b | 10.95–11.25 | 6.85–6.95 | 16.4 / 27.9 | 26.8 / 43.5 | 1.16x |
| idempotent 1b | 7.05 | 6.15–6.75 | 33.7 / 43.7 | 26.3 / 42.4 | 1.20x |
| producer 3b | 6.95 | 4.95 | 21.1 / 39.0 | 11.3 / 21.8 | 1.33x |
| acks-all 3b | 9.75 | 6.65 | 20.1 / 33.6 | 18.1 / 32.4 | 1.29x |
| idempotent 3b | 9.75 | 6.35 | 21.2 / 44.4 | 14.4 / 25.2 | 1.33x |

Note the 1-broker tails: Dekaf p95/p99 are already far better than Confluent. The loss is specifically the **median**, and it is architectural, not noise.

## Mechanism

`BrokerUnackedByteBudget.ComputeBudget` (src/Dekaf/Producer/BrokerUnackedByteBudget.cs:1252-1267):

```csharp
var latencyGovernedTargetSeconds = _targetSeconds * _latencyBudgetScale;
...
var horizonSeconds = _hasMinRttSample
    ? Math.Max(latencyGovernedTargetSeconds, rttFloorSeconds)   // target is a FLOOR
    : latencyGovernedTargetSeconds;
rateBudgetBytes = effectiveMaxRate * horizonSeconds;
[...]

Issue #1796, opened 2026-07-11
perf(producer): adaptive scale-down collapses acks=all throughput mid-run below Confluent


thomhurst:
## Summary

In the stress run [29159277976](https://github.com/thomhurst/Dekaf/actions/runs/29159277976), the **`producer-acks-all` / 1-broker / Dekaf** scenario is the only head-to-head case where Dekaf ends a run **slower than Confluent**. Dekaf starts ~27% ahead and finishes ~10% behind because throughput **collapses mid-run** and never recovers.

### Evidence (1-broker, acks=all, non-idempotent, fire-and-forget)

| Metric | Dekaf | Confluent |
|---|---|---|
| Peak msg/s | ~2.42M | ~1.46M |
| **First-third avg** | **1,754,379** | 1,375,973 |
| **Last-third avg** | **1,270,449** | **1,409,035** |
| Intra-run drift | **-27.6%** (decays) | +2.4% (stable) |
| Median-interval msg/s | 1,385,441 | 1,397,279 |
| alloc B/msg | 0 | 1289 |

Bucketed timeline (K msg/s, 30 × ~30s buckets, 900s run):
```
1639 1668 1556 1621 1547 1374 | 2082 2078 2001 2084 | 1406 1227 1293 1314 1330 1300 1219 1315 1283 1330 1358 1368 1261 1098 1126 1227 1285 1265 1244 1413
```
Shape: warmup ~1.6M → **scale-up spike ~2.0M (≈180–300s)** → **collapse to ~1.25M sustained for the rest of the run**. The collapse begins ~120s after the spike starts — exactly `ScaleDownSustainedMs`.

Confluent on the same broker holds a flat ~1.4M with no collapse, so the broker is **not** degrading. The fixed-connection Dekaf variants (`Dekaf (3conn)`) are stable across every scenario (producer 3conn 1b drift -3.2%, idempotent 3conn 1b -1.6%). **Only adaptive single-connection acks=all collapses** — this isolates the cause to the adaptive connection scaler, not the broker or the wire path.

## Root cause

`BrokerSender` adaptive scaling keys **both** scale-up and scale-down decisions on `RecordAccumulator.BufferUtilization` (`src/Dekaf/Producer/BrokerSender.cs:4166`), and scale-down fires when `BufferUtilization < 0.3` sustained for `ScaleDownSustainedMs` (2 min) — `BrokerSender.cs:4208`, `:511-512`.

[...]

PR #2154, opened 2026-07-15
feat: make stress profiling produce decision-ready bottleneck data

thomhurst:
## Summary

Follow-up to #2150: profiled stress runs shipped raw artifacts that still required manual interpretation, and the captured data missed the state behind most recent perf regressions. This PR makes profiled dispatches produce concrete, decision-ready data.

### Library: producer internal-state gauges (the big one)
Most recent root causes (#2108 budget horizon, #1856 stuck-low budget equilibrium, #1796 scale-down inversion) were adaptive-controller pathologies invisible to CPU traces and System.Runtime counters — they were inferred indirectly from throughput curves across paid runs. 15 new observable instruments on the `Dekaf` meter expose that state directly, per broker (tagged `messaging.kafka.broker.id` + `messaging.client.id`):

| Area | Instruments |
|---|---|
| Budget/BBR | `broker.budget_bytes`, `broker.unacked_bytes`, `broker.min_rtt`, `broker.max_delivery_rate`, `broker.queue_latency_ewma`, `broker.latency_budget_scale`, `broker.admission_blocks`, `broker.capacity_probe.successes/.failures` |
| Connections/in-flight | `broker.connections`, `broker.in_flight_bytes`, `broker.in_flight_requests` |
| BufferMemory | `buffer.used_bytes`, `buffer.limit_bytes`, `buffer.pressure_events` |

All pull-based (`ObservableGauge`/`ObservableCounter`): zero cost unless a listener polls, zero produce-hot-path work, reads only volatile-published accessors. Registry mirrors the existing consumer-lag gauge pattern. `dotnet-counters` in profiled runs now collects `System.Runtime,Dekaf`, so controller state arrives as a 1 Hz time series.

### Trace capture fixes
- **`full` profile level 4 → 5**: `GCAllocationTick` is verbose-only — level 4 silently dropped allocation-by-type data. Also adds `TplEventSource:0x1C3` for async causality (await/wait attribution in PerfView).
- **3 stack snapshots per window**, spread inside the traced window (concurrent EventPipe sessions) — distinguishes stuck stacks from transient ones.
- **Opt-in per-window gcdump** (`PROFILE_GCDUMP`, auto-on for `gc` dispatches), captured after each window's trace so the induced blocking GC never pollutes it.

### Post-run analysis (new)
- **`tools/Dekaf.TraceAnalyzer`** (TraceEvent-based): `dotnet-trace report` only analyzes CPU samples, so gc/contention/full traces shipped raw with zero automated analysis. The analyzer aggregates each window into allocations-by-type (with LOH ticks and rates), contention totals, GC pause/generation/reason stats, threadpool adjustment reasons, and exception counts. Built once up front (fail-fast before the paid run), invoked as a binary per window.
- **`stress_profile_summary.py`**: emits `profile-summary.md` with a per-window counter mean/max table and a window-to-window topN inclusive-CPU movement table ranked by delta — the "what changed between healthy and collapse" question answered directly. The topN parser targets the pinned dotnet-trace version's verified row format.

### Workflow fixes
- **EXTRA_ARGS argv bug**: #2150 changed `${EXTRA_ARGS}` (word-split) to `STRESS_ARGS+=("$EXTRA_ARGS")` (single element). Harmless today (single-token flag) but the first multi-token addition would reach the client as one unparseable argument. Now a proper bash array on both paired and 3conn paths.
- gcdump tool installs only for `gc` dispatches; analyzer build ownership moved into the profile script (single owner, fail-fast, no per-window MSBuild).

## Validation
- Analyzer verified against a real 6s gc-profile trace of an allocation-churn app (alloc-by-type table, GC pause stats confirmed sane; non-GC EE suspensions filtered).
- topN row format verified against dotnet-trace `PrintReportHelper` source; parser unit-tested against that shape.
- 5591 unit tests, 33 stress-tool tests, 91 CI script tests pass. `/simplify` review applied (4 agents; factory-collapsed gauge declarations, generic observe helpers, providers-content aggregation gate).

- Consolidate the three `FormatBytes` copies across `tools/` into a shared source file.
- Consider computing topN movement from speedscope JSON instead of scraping the report text (removes 70-char truncation collisions).

### DIFF
--- a/.github/scripts/stress_profile_summary.py
+++ b/.github/scripts/stress_profile_summary.py
@@ -1,0 +1,232 @@
+"""Turn raw stress-profile artifacts into a decision-ready markdown summary.
+
+Inputs (produced by tools/profile-stress-test.sh in --profile-dir):
+  profile-metadata.txt     key=value run metadata, including trace_windows and
+                           the measured_start/counters_start epochs.
+  runtime-counters.csv     1 Hz dotnet-counters output (System.Runtime + Dekaf).
+  <label>.topN-inclusive.txt  dotnet-trace topN report per trace window.
+
+Output: one markdown file with
+  1. per-window counter aggregates (mean/max per counter per window), and
+  2. window-to-window movement of topN inclusive CPU percentages,
+so bottleneck decisions come from tables instead of eyeballing raw artifacts.
+"""
+
+from __future__ import annotations
+
+import argparse
+import csv
+import re
+import sys
+from pathlib import Path
+
+# Matches dotnet-trace PrintReportHelper rows: "12. Name(args)   45.67%   12.34%"
+# The name field is fixed-width (70) so 2+ spaces always precede the metrics.
+TOPN_ROW = re.compile(r"^\s*(\d+)\.\s+(?P<name>.*?)\s{2,}(?P<inclusive>\d+(?:\.\d+)?)%\s+(?P<exclusive>\d+(?:\.\d+)?)%")
+
+
+def parse_metadata(path):
+    metadata = {}
+    if not path.exists():
+        return metadata
+    for line in path.read_text(encoding="utf-8").splitlines():
+        key, sep, value = line.partition("=")
+        if sep:
+            metadata[key.strip()] = value.strip()
+    return metadata
+
+
+def parse_windows(spec):
+    windows = []
+    for part in spec.split(","):
+        pieces = part.split(":")
+        if len(pieces) != 3:
+            continue
+        offset, duration, label = pieces
+        windows.append((int(offset), int(duration), label))
+    return windows
+
+
+def load_counter_rows(path):
+    """Return (offset_seconds, provider, counter, type, value) per CSV row.
+
+    dotnet-counters timestamps are local-time and culture-formatted, so rows
+    are anchored by tick index instead: each distinct timestamp string is one
+    1 Hz refresh interval from the counters start.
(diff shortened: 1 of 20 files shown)


thomhurst:
Addressed all three review items in 65ee32ed:

1. **Broker-gauge behavioral coverage** — added `ProducerStateGaugeIntegrationTests`: produces 50 awaited messages against a real broker, then asserts every one of the 12 per-broker instruments reports at least one measurement carrying the `messaging.kafka.broker.id` tag for the test's client id, with value sanity beyond presence (`broker.connections >= 1`, `broker.budget_bytes > 0` — cold start publishes the cap, `latency_budget_scale` in (0, 1], everything non-negative). Also asserts a disposed producer stops reporting entirely. Verified green locally against a real container (6.7s).

2. **Per-source exception isolation** — `ObserveProducers` (both overloads) now guards each source the way `ObserveAllConsumerLag` does, with one improvement: the enumerable overload materializes the selector's lazy iterator *inside* the try, so enumeration-time failures are also isolated (the consumer-lag guard only covers callback creation). A producer mid-disposal can no longer drop other producers' measurements from a scrape.

3. **Package ordering** — `Microsoft.Diagnostics.Tracing.TraceEvent` moved to its alphabetical slot after `Microsoft.Bcl.AsyncInterfaces`.

Full unit suite re-verified at 5591/5591 after the isolation change.

PR #2158, opened 2026-07-16
Amortize broker admission and seek window knee

thomhurst:
## Summary

- amortize broker-wide admission with bounded 64 KiB batch-local byte leases instead of one CAS per record
- refund unused lease credit at seal and carry exact logical batch bytes through routing, acknowledgement, retry, failure, and disposal
- keep the controller request quantum fixed to configured batch capacity; observed partial request sizes never redefine its step or floor
- start at 16 request quanta, retain an 8-quantum safety floor, and search the knee with whole-quantum treatments
- evaluate capacity changes only after a 30-second cooldown and three settled, generation-qualified 500 ms epochs
- retain a lower window only at >=99% baseline goodput; retain a higher window only with >=1% gain and bounded controlled delay
- remove blind multiplicative recovery and active minimum-RTT drain probes that could manufacture starvation
- remove the obsolete reflection-based admission-probe integration test

## Why

The first version reserved every record globally and learned its floor from acknowledged request size. Once the gate caused partial seals, smaller requests reduced the next floor, which caused more partial seals: a positive-feedback collapse. A full-batch lease avoided per-record contention but stranded roughly one batch per active partition and produced the same starvation in a different form.

The final design uses small fixed lease chunks. Most records consume batch-local credit, broker CAS frequency falls by roughly 64x for 1000 B records, and unused credit is bounded to less than one chunk per open batch. The controller works in immutable request quanta and treats every window change as an experiment, so latency outliers cannot lower throughput without goodput evidence.

## Concurrency and ownership

- `BrokerUnackedByteBudget`: multi-writer atomic lease/accounting facade
- `PartitionBatch`: owns lease credit; refunds slack and transfers exact logical bytes at seal
- `BrokerWindowController`: allocation-free, single-writer control state on the broker response loop
- admission generations prevent feedback from an old window deciding a new window
- leader changes transfer already-admitted ownership; terminal paths release exactly once
- one oversized lease may enter an empty window, avoiding valid-record deadlock

## Validation

- .NET 10 unit suite: 5,516 passed, 0 failed
- producer integration suite: 26 passed, 0 failed
- full Release solution build: 0 warnings, 0 errors
- changed-file whitespace/style verification: passed
- local Docker diagnostic (directional): 148,608 msg/s and 1.81 us/msg, versus 87,290 msg/s and 2.45 us/msg for full-batch leases; one settled down probe retained 14 MiB and the window did not collapse
- full GitHub Actions CI: passed

## GitHub Actions stress result

[...]

PR #2161, opened 2026-07-16
ci: add exact-SHA stress A-B-A gate

thomhurst:
## What changed

- adds optional `baseline_sha` dispatch input for supported duration-based producer lanes
- checks out and builds the exact baseline beside the candidate
- runs baseline → candidate → baseline on one paid runner with identical workload settings
- compares throughput, latency, CPU, allocations, and stability with 3% adverse tolerance
- marks controls above 10% drift inconclusive; both regression and inconclusive verdicts block acceptance
- uploads baseline controls separately and keeps the candidate in the normal stress summary
- records the Pareto-safe workflow in `CLAUDE.md`

## Why

Repeated identical-SHA runs varied enough across Ubicloud VMs to make absolute throughput and CPU/message comparisons ambiguous. Bracketing the candidate with the immediate previous product SHA makes runner drift observable and prevents selecting a latency-only or throughput-only regression.

## Validation

- `python -m unittest discover -s .github/scripts -p "test_stress_*.py"` — 105 passed
- workflow YAML parse — passed
- actionlint — no new findings; existing custom-runner and legacy shell warnings remain
- replayed historical #2158 artifacts through the comparator; it reproduced the known +3.07% CPU regression while preserving the throughput/latency wins


thomhurst:
Addressed both non-blocking review suggestions in `99e03d03`:

- added `test_decisive_regression_overrides_noisy_controls`, pinning regression precedence when control drift is also high;
- removed the unused `$GITHUB_OUTPUT` surface and step id.

Validation: full stress-script suite now passes 105/105; workflow YAML parse passes.

PR #2192, opened 2026-07-17
fix(tests): root-cause and fix BrokerSenderMuteOrderingTests 30s hangs (ArrayPool double-return poisoning)

### LINKED ISSUE #2187: Flaky 30s hangs in BrokerSenderMuteOrderingTests: mock connection turns unscripted sends into an infinite retriable-exception loop
## Symptom

`Unit Tests (net8.0)` CI job intermittently fails with a 30s test timeout, e.g. on PR #2185 (run 29580062024, unrelated hosting/DLQ changes):

- `NonIdempotentProducer_MultipleInFlight_PipelinesSamePartitionBatches` - failed (canceled) after 30s
- Also observed hung locally: `InFlightWait_MutedPartitionFilter_MovesCoalescedNormalBatchToCarryOver`

Local repro on Windows, net8.0, Release: ~4/25 class-filtered runs fail when the machine is under load; 0/40 when the single test runs alone. Not thread-pool starvation - in a captured hang, all pool workers were idle.

## Root cause (from a dump of a hung run)

Captured with `dotnet-dump` mid-hang. The heap contains **6,611 faulted-task holders, 3,307 `InvalidOperationException`, 1,653 `NullReferenceException`** accumulated in ~20s, and one live thread looping inside `BrokerSender.SendCoalescedAsync`:

```
Dekaf.Producer.ReadyBatch.IsCurrentIncarnation(Int32)
Dekaf.Producer.BrokerSender+<SendCoalescedAsync>d__190.MoveNext()   <- exception-handling path
System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
...
Dekaf.Producer.BrokerSender+<SendLoopAsync>d__151.MoveNext()
```

The `InvalidOperationException` is:

```
Message: Queue empty.
  Queue`1.Dequeue()
  BrokerSenderMuteOrderingTests+<>c__DisplayClass4_0.<CreateMockConnection>g__DequeueResponse|0()
  TestKafkaConnection+<SendPipelinedAfterWriteAsync>d__63.MoveNext()
  -> BrokerSender+<SendCoalescedAsync>d__190.MoveNext()
```

The `NullReferenceException` is `MetadataManager.NegotiateApiVersionsAsync` on the reconnect path (the mock does not script version negotiation).

**Chain:** the test harness scripts exactly N responses and throws `Queue empty.` on any additional send. When a timing race produces one extra send, `SendCoalescedAsync` treats the harness exception as a retriable connection failure and retries against the same empty queue, forever (~165 exceptions/sec, so effectively no backoff on this path). The test never sees its expected `LogPipelinedSend`/ack events and dies at the TUnit 30s timeout instead of failing with a diagnosis.

## Suggested fixes

1. **Test harness (primary):** `CreateMockConnection` should fail fast and loud on an unscripted send - e.g. record the unexpected send and return a response task that faults the test immediately (or a non-retriable error), instead of throwing `InvalidOperationException` that the sender retries forever. That converts a 30s opaque hang into an instant assertion naming the extra send.
2. **Product (secondary, needs its own validation):** audit backoff on `SendCoalescedAsync`'s synchronous-exception retry path - thousands of retry cycles in seconds suggests the retry storm is not paced by `RetryBackoffMs` there. Also `NegotiateApiVersionsAsync` NREs on the reconnect path deserve a null-guard look.
3. Root-cause why the extra (unscripted) send happens at all under load - candidates: request-timeout redelivery, batch coalescing changing request grouping, or incarnation-based redelivery racing the scripted responses.

Per CLAUDE.md, flaky tests indicate real bugs - this one is a genuine harness-design defect amplified by product retry behavior.

thomhurst:
Fixes #2187.

## What was actually wrong

#2187 diagnosed the symptom (mock throws `Queue empty.` → `SendCoalescedAsync` retries forever → opaque 30s hang) but not the mechanism that produced the extra/missing sends. Stress-looping the class under load (3 concurrent 25-run lanes, ~15/75 failing) with new instrumentation pinned it down:

**`MetadataRerank_RequestTimeoutClearsMigrationFence` invoked the send-loop-owned `HandleTimedOutRequests` reflectively from the test thread while the sender loop was live.** Both threads walk the same pending-response list and call `PendingResponse.ReturnBatchesArray()` on the same entry, so the rented `ReadyBatch[]`/`int[]` arrays are returned to the process-global `ArrayPool<T>.Shared` **twice**. Duplicate pool references persist for the rest of the test process. A later test that pipelines two requests (`NonIdempotentProducer_MultipleInFlight_...`, `InFlightWait_...` — exactly the tests seen hanging) then rents **the same array for both in-flight `PendingResponse`s**:

- processing response #1 reads the aliased array and completes the *other* request's batch (captured in a failing run: send #1 = firstBatch, send #2 = secondBatch, ack for offset 100 delivered to secondBatch, `second.pooled=True`, `first` orphaned with trace `CSGW`);
- `ReturnBatchesArray(clearArray: true)` then nulls the shared slots, so response #2 logs `expected []` and its ack is swallowed;
- the orphaned batch either leaves the test waiting for an ack that never comes (30s hang) or gets retried against the exhausted mock script (`Queue empty.` storm from the issue's dump — the other victim mode).

## Fixes

**Test (root cause):** the racy test now advances the injected timestamp and wakes the send loop with an unrelated-topic batch so the loop's *own* iteration runs the request-timeout sweep, then polls `GetMigratingPartitionCount`. Loop-owned methods are never invoked reflectively on a live sender.

**Product hardening (same hazard class):** `DisposeCoreAsync` waits up to 5s for the send loop; previously, on timeout it proceeded to fail pending responses and return their arrays while the still-running loop's exit cleanup would return them again. The cleanup is now gated on the loop actually having completed (its own exit path fails the batches and returns the arrays when it exits); a warning is logged instead.

**Fail-fast harness (issue's primary ask):** a new shared `ScriptedProduceResponses` (adopted by both `BrokerSenderMuteOrderingTests` and `BrokerSenderSendLoopTests`, which carried an identical copy of the hang-prone raw-`Dequeue` mock) replaces the raw scripted queue. An unscripted (extra) send:
- records a diagnostic naming the consumed script count plus the sender-side stack,
- cancels a guard token linked into every in-test wait (`ct = GuardUnscriptedSends(ct)`), aborting the test instantly,
- parks the sender on a never-completing response (no retriable-exception storm, no reconnect NREs),
- and an `[After(Test)]` hook rethrows the diagnostic so the failure names the extra send.

Validated by sabotage: scripting 1 response where 2 sends occur now fails in ~130-240ms with `Unscripted send: all 1 scripted produce responses were already consumed...` instead of a 30s opaque timeout.

**Deterministic coalescing scripts:** `MutedPartitionHeldBack_...` and `RetriableError_OnOnePartition_...` enqueued batch C *after* releasing the response, racing the loop: the retry could dispatch alone before C's channel event drained, splitting send 2 and producing a genuinely unscripted third send under load (caught by the new harness in 243ms). C is now enqueued while send 1 is still in flight.

**Diagnostics kept for the next investigation:** bounded event capture in `PipelinedSendLogger` (`DumpEvents`), opt-in produce-request snapshots in `TestKafkaConnection` (`CaptureProduceRequests` / `CapturedProduceRequests` — captured at write time via the request's internal scratch arrays, because the reusable scratch request is cleared after each send), and bounded waits with a full state dump in `NonIdempotentProducer_MultipleInFlight_...`.

## Validation

- Stress: 3 concurrent lanes x 30 class runs (net8.0, Release): **90/90 pass** (baseline before fixes: ~15/75 failing with 30s hangs). Post-cleanup re-run: mute class 50/50 + send-loop class 12/12 under mutual load.
- Full unit suite: net10.0 5565/5565, net8.0 5439/5439. (One unrelated pre-existing full-suite-only consumer flake observed once — `Revoke_WhenCommitExceedsRebalanceTimeout_Continues`, 0/6 solo repro — tracked separately.)
- Harness fail-fast exercised via deliberate sabotage (see above), then reverted.

Two more double-return windows exist on exception paths and deserve their own hardening pass: a mid-pass exception in `ProcessCompletedResponses` between `ReturnBatchesArray()` and the compaction `RemoveRange` leaves returned entries in the list for the loop-exit cleanup to re-return; the loop-exit cleanup itself has the same partial-failure shape. Both are unreachable without an exception escaping paths that are already per-batch guarded, so they are noted rather than changed here.

### DIFF
--- a/src/Dekaf/Producer/BrokerSender.cs
+++ b/src/Dekaf/Producer/BrokerSender.cs
@@ -4851,6 +4851,10 @@ internal sealed partial class BrokerSender : IAsyncDisposable
             LogBatchCleanupStepFailed(ex, _brokerId);
         }
 
+        // True when the loop finished (normally or faulted — its exit cleanup ran either way);
+        // false when the WaitAsync above timed out and the loop is still running.
+        var sendLoopExited = _sendLoopTask.IsCompleted;
+
         // A final send-loop iteration may have already detached a retirement drain after
         // clearing the raw connection fields. Wait for every such drain before inspecting
         // the remaining scale-down state.
@@ -4900,7 +4904,16 @@ internal sealed partial class BrokerSender : IAsyncDisposable
         ClearMigratingPartitions();
 
         var totalPending = _totalPendingResponseCount;
-        if (totalPending > 0)
+        if (totalPending > 0 && !sendLoopExited)
+        {
+            // The send loop is still running and owns _pendingResponsesByConnection. Touching
+            // the lists here would race the loop's own exit cleanup and return the same rented
+            // arrays to ArrayPool.Shared twice — duplicate pool entries later hand one array to
+            // two concurrent requests, pairing responses with the wrong batches (#2187). The
+            // loop's exit path fails these batches and returns the arrays when it completes.
+            LogSkippingPendingResponseCleanup(_brokerId, totalPending);
+        }
+        else if (totalPending > 0)
         {
             LogFailingPendingResponses(_brokerId, totalPending);
 
@@ -5645,6 +5658,9 @@ internal sealed partial class BrokerSender : IAsyncDisposable
 
     [LoggerMessage(Level = LogLevel.Warning, Message = "BrokerSender[{BrokerId}] failing {RemainingCount} pending responses during disposal")]
     private partial void LogFailingPendingResponses(int brokerId, int remainingCount);
+
+    [LoggerMessage(Level = LogLevel.Warning, Message = "BrokerSender[{BrokerId}] send loop still running after disposal wait; leaving {RemainingCount} pending responses to its exit cleanup")]
+    private partial void LogSkippingPendingResponseCleanup(int brokerId, int remainingCount);
 
     [LoggerMessage(Level = LogLevel.Warning, Message = "BrokerSender[{BrokerId}] non-fatal exception during batch cleanup step (suppressed)")]
     private partial void LogBatchCleanupStepFailed(Exception exception, int brokerId);
--- a/src/Dekaf/Protocol/Messages/ProduceRequest.cs
+++ b/src/Dekaf/Protocol/Messages/ProduceRequest.cs
@@ -89,6 +89,17 @@ public sealed class ProduceRequest : IKafkaRequest<ProduceResponse>, IKafkaReque
         return true;
     }
 
+    /// <summary>
+    /// Number of topic entries regardless of whether the request was populated through
+    /// <see cref="TopicData"/> or the internal scratch arrays. Pairs with
+    /// <see cref="GetTopicEntry"/>; exists so tests can observe scratch-built requests
+    /// without reflecting into private fields.
+    /// </summary>
+    internal int TopicEntryCount => _topicDataScratch is not null ? _topicDataScratchCount : TopicData.Count;
+
+    internal ProduceRequestTopicData GetTopicEntry(int index) =>
+        _topicDataScratch is { } topicDataScratch ? topicDataScratch[index] : TopicData[index];
+
     public void Write(ref KafkaProtocolWriter writer, short version)
     {
         writer.WriteCompactNullableString(TransactionalId);
@@ -146,6 +157,19 @@ public sealed class ProduceRequestTopicData
         _partitionDataScratchStart = 0;
         _partitionDataScratchCount = 0;
     }
+
+    /// <summary>
+    /// Number of partition entries regardless of whether the topic was populated through
+    /// <see cref="PartitionData"/> or the internal scratch arrays. Pairs with
+    /// <see cref="GetPartitionEntry"/>; exists so tests can observe scratch-built requests
+    /// without reflecting into private fields.
+    /// </summary>
+    internal int PartitionEntryCount => _partitionDataScratch is not null ? _partitionDataScratchCount : PartitionData.Count;
+
+    internal ProduceRequestPartitionData GetPartitionEntry(int index) =>
+        _partitionDataScratch is { } partitionDataScratch
+            ? partitionDataScratch[_partitionDataScratchStart + index]
+            : PartitionData[index];
 
     internal bool TryGetSinglePartition(out ProduceRequestPartitionData partition)
     {
--- a/tests/Dekaf.Tests.Unit/Producer/BrokerSenderMuteOrderingTests.cs
+++ b/tests/Dekaf.Tests.Unit/Producer/BrokerSenderMuteOrderingTests.cs
@@ -24,6 +24,10 @@ namespace Dekaf.Tests.Unit.Producer;
 /// </summary>
 public sealed class BrokerSenderMuteOrderingTests
 {
+    // Well below the 30s TUnit timeout so a stalled wait fails with a state dump instead of
+    // an opaque test-level timeout, while still generous for loaded CI runners.
+    private static readonly TimeSpan DiagnosticWaitTimeout = TimeSpan.FromSeconds(15);
+
     private static readonly MethodInfo CoalesceBatchMethod = typeof(BrokerSender).GetMethod(
         "CoalesceBatch",
         BindingFlags.Instance | BindingFlags.NonPublic)!;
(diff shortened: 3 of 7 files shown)


thomhurst:
Addressed the review in 6f201be7.

**Guard wiring gap in `BrokerSenderSendLoopTests` (the blocking item):** confirmed — the harness was adopted there but `Guard(...)` was never linked into the tests'' cancellation tokens, so an unscripted send would only surface via the `[After(Test)]` hook after the full 30-120s timeout. Every test in that file that exercises the mock send path now reassigns its token via a new `GuardUnscriptedSends(cancellationToken)` helper (it chains `Guard` across all `ScriptedProduceResponses` the test registered, covering the one test that builds its mock inline). The only unguarded `CreateMockConnection` call site left is `Constructor_AppliesCapWithoutInflatingCongestionWindow`, which takes no token and never sends.

Validated the wiring the same way as the mute-ordering class: sabotaging `SendLoop_ResponseCompletion_WakesSendLoopAndProcessesBatch` with an empty script fails in ~150ms with `After(Test) hook failed: Unscripted send: all 0 scripted produce responses were already consumed...` instead of running out its 120s timeout. Sabotage reverted afterwards.

**Disposal trade-off:** agreed and intentional, exactly as you framed it — when the send loop is genuinely wedged past the 5s disposal wait, pending batches are now left to the loop''s own exit cleanup rather than force-failed, because the force-fail path is what double-returned the arrays and corrupted the shared `ArrayPool`. A stuck send loop is already an anomaly; corrupting unrelated components was the strictly worse outcome. The `LogSkippingPendingResponseCleanup` warning records when this path is taken.

Re-validation after the change: both producer test classes 60/60 under 3-lane concurrent load; full unit suite green on net8.0 and net10.0 (the one intermittent full-suite failure seen during runs is the pre-existing consumer flake tracked in #2193, unrelated to this diff).

thomhurst:
Addressed review round 2 in 12310d45.

**1. `CancelAsync` catch that can never fire (CONFIRMED)** - fixed. You are right that `CancelAsync()` on a disposed source faults the *returned task* rather than throwing synchronously, so the `catch (ObjectDisposedException)` was dead and the discarded task was an unobserved-task-exception hazard - ironic in a diagnostics harness, agreed. The call now observes the fault via `ContinueWith(static t => _ = t.Exception, ..., OnlyOnFaulted | ExecuteSynchronously, TaskScheduler.Default)`, matching the repo''s documented task-exception pattern. Kept the cancel outside the lock (a synchronous `Cancel` under the lock could resume guarded waits inline on the sender thread).

**4. Duplicated guard/teardown glue** - consolidated. Both test classes now inherit a shared `ScriptedProduceResponseFixture` (in `ScriptedProduceResponses.cs`) providing `RegisterScript`, `GuardUnscriptedSends` (chains every registered script), and the `[After(Test)]` rethrow hook; the two divergent per-class copies are deleted. Re-validated the base-class hook fires under TUnit via the same sabotage as before: empty script -> ~130ms failure with the named "Unscripted send" diagnostic (sabotage reverted).

**5. `WaitUntilAsync` vs `TestWait.UntilAsync`** - deduplicated. `TestWait.UntilAsync` gained a `(condition, timeout, timeoutMessage, cancellationToken, pollInterval)` overload; the local fork is deleted and both the diag-marker wait and the sweep poll use it (1ms poll to keep the original latency).

**6. Reflection into scratch fields** - replaced. `ProduceRequest`/`ProduceRequestTopicData` now expose internal scratch-aware accessors (`TopicEntryCount`/`GetTopicEntry`, `PartitionEntryCount`/`GetPartitionEntry`) alongside the existing `TryGetSingleBatch`/`TryGetSinglePartition`; `TestKafkaConnection` uses them and the string-keyed reflection block is deleted, so a scratch-field rename is now a compile error.

**2 (wedged-loop array leak) and 3 (dual cleanup ownership)** - agreed on both framings; not changed here. The leak is the deliberate lesser evil vs. pool corruption, and the deeper fix you describe (send loop as sole unconditional owner of `_pendingResponsesByConnection` cleanup, which would also subsume the exception-path windows already listed in the PR description) is now tracked in #2195 together with adding a metric to distinguish "recovered late" from "leaked forever".

Re-validation after the changes: full unit suite green on net8.0 (5439) and net10.0 (5565); both producer classes 60/60 under 3-lane concurrent stress.

Issue #2195, opened 2026-07-17
BrokerSender: consolidate pending-response cleanup ownership; remaining double-return windows and wedged-loop array leak


thomhurst:
Follow-ups from #2192 (which fixed the #2187 ArrayPool double-return corruption). The review on that PR confirmed the shipped gate is correct but identified the remaining work:

1. **Consolidate cleanup ownership (altitude).** `DisposeCoreAsync` still carries its own copy of the pending-response cleanup, gated on `sendLoopExited`, while the send loop''s exit path has the authoritative copy. Two code paths that can walk the same lists and call `ReturnBatchesArray()` is exactly the shape that produced #2187. Deeper fix: make the send loop the sole unconditional owner of `_pendingResponsesByConnection` cleanup (disposal only cancels + awaits), deleting the duplicated block instead of gating it.

2. **Wedged-loop array leak (accepted trade-off, needs observability).** When the send loop is genuinely deadlocked past the 5s disposal wait, pending responses'' rented `ReadyBatch[]`/`int[]` arrays are now intentionally left to the loop''s exit cleanup - which never runs if the loop never exits. Callers do not hang (`KafkaProducer.DisposeAsync` force-fails in-flight batches independently), but the arrays leak silently with only a Warning log (`LogSkippingPendingResponseCleanup`). Consider a counter/metric so "recovered late" is distinguishable from "leaked forever".

3. **Exception-path double-return windows (from #2192''s follow-ups section).**
   - `ProcessCompletedResponses`: an exception escaping between `ReturnBatchesArray()` and the compaction `RemoveRange` leaves already-returned entries in the list for the loop-exit cleanup to re-return.
   - The loop-exit cleanup itself has the same partial-failure shape (a throw mid-walk leaves early-returned entries listed).
   Both are unreachable today without an exception escaping paths that are already per-batch guarded; fixing (1) properly likely subsumes them (single idempotent owner).

PR #2200, opened 2026-07-17
fix(producer): make PendingResponse array return exactly-once; expose orphaned-reservation diagnostics

### LINKED ISSUE #2199: Producer refund leak: delivered wave's BufferMemory reservations orphaned → 120s max.block.ms starvation (release-gate run 29594249742, AOT/4.3.1)
## Failure

Release-gate CI run [29594249742](https://github.com/thomhurst/Dekaf/actions/runs/29594249742), job `Release Matrix AOT (4.3.1) / produce` (attempt 1): `BackpressureIntegrationTests.BufferFull_ProduceBlocks_UntilBatchSent` failed with

```
KafkaTimeoutException: Failed to allocate buffer within max.block.ms (120000ms).
Requested 1111 bytes, current usage: 65508/65536 bytes.
```

First-ever occurrence of this test failing (added 2026-02-07, never modified). Attempt 2 passed in 114ms; all 7 sibling produce lanes were green — a narrow race, plausibly widened by the AOT lane's lack of JIT warmup (faster first-wave ack).

## Evidence (attempt-1 artifact 8412583758, embedded per-test OTel spans)

- Producer config: BufferMemory=65536, MaxBlock=2min, linger 5ms, acks=Leader, non-idempotent, 1 partition, single connection (2 socket connects total in spans).
- ~59 publish spans (messages 0–58, ≈65,508 reserved bytes) all completed in ~4.5ms — appended, sent, acked, `ProduceAsync` resolved **successfully** (the test awaits sequentially and only surfaced a later message's timeout, so all earlier tasks succeeded).
- Every later publish span blocked the full 120,006ms. The timeout message reads the **live** counter (`PendingAppend.OnTimeout`) — usage was still 65508/65536 at t=120s, i.e. **zero admissions and zero refunds for 120 seconds after a fully delivered wave**.
- Dispose then logged `Producer disposed in 0ms (graceful=True)` — consistent: no unsealed and no in-flight batches remained (`FlushAsync` fast-path), because the delivered wave had exited the pipeline and the remaining messages were never appended.

So a fully delivered wave's DataSize reservations were permanently orphaned. Not a dead sender loop (no error logs, wait is timeout-bounded), not a lost wakeup (`ReleaseMemory` itself runs `DrainPendingAppends`).

## Analysis

Exhaustive static analysis of the single-owner paths shows every code path that resolves user tasks with success also releases memory:

- post-write release loop (`BrokerSender.SendCoalescedAsync`, gated by `IsCurrentIncarnation`);
- success completion `ProcessResponseBatches` → `CompleteSend` → `CleanupBatch` → `CleanupBatchResources` → `ReleaseBatchMemory` (per-batch try/catch, exceptions route to `FailAndCleanupBatch`);
- the `_memoryReleased` tri-state is reset only in pool `Initialize`; `TrySetMemoryReleased` is test-only.

The surviving mechanism is **cross-request array aliasing via `ArrayPool.Shared` double-return poisoning** — the #2187 family. `PendingResponse.ReturnBatchesArray` was documented "must be called exactly once — not idempotent" yet had five call sites (timeout sweep, faulted-task, normal processing, abandon, dispose). One concrete window: in `ProcessCompletedResponses` the arrays are returned **inside** the processing loop but entry compaction happens **after** it — an exception escaping mid-loop leaves already-returned entries queued for re-processing next iteration (second return + wrong-batch pairing). A double return anywhere in the process (the produce job runs Producer / MultiInFlightProducer categories first, in the same exe) poisons the shared pool; a later request's send-side `batches[]`/`generations[]` then aliases another request's arrays, letting the release-side incarnation gate see stale/nulled slots (skipping releases — a null slot even NRE'd past the rest of the request's releases) while the response side sees rewritten current slots and completes users.

## Fix (PR to follow)

- One-time-return claim on `PendingResponse`'s rented arrays: only the first cleanup path returns them; later calls no-op + log Warning (kills the poisoning family at the source, regardless of which path misfires).
- Null-guard in the post-write release loop so one corrupted slot cannot skip the remaining batches' memory releases.
- Buffer-timeout message now includes in-flight/unsealed batch counts (this failure would have read `in-flight batches: 0, unsealed batches: 0` — instantly identifying orphaned reservations instead of looking like slow network).
- Tests: guard claim semantics, cross-copy no-op contract, and a 20-wave leak invariant (`BufferedBytes == 0` after every acked wave).

## Follow-ups (from #2195, made observable but not structurally removed by the PR)

- Single-ownership consolidation: one place removes-from-list-and-cleans a `PendingResponse` exactly once by construction (five lifecycle paths today). The new Warning log (`PendingResponse arrays already returned`) is the telemetry to find any path that actually double-fires.
- Return-then-compact ordering in `ProcessCompletedResponses` (exception between return and compaction re-processes stale entries).
- Wedged-loop array-leak observability (deferred from #2192).

Related: #2187, #2192, #2195.

thomhurst:
Fixes #2199.

## Root cause

Release-gate run [29594249742](https://github.com/thomhurst/Dekaf/actions/runs/29594249742) (AOT/4.3.1 produce lane): `BufferFull_ProduceBlocks_UntilBatchSent` timed out after the full 120s `max.block.ms` with live usage 65508/65536. The attempt-1 artifact's trace spans show a fully delivered first wave (~59 messages, sent + acked + `ProduceAsync` resolved successfully in 4.5ms) whose BufferMemory reservations were **never refunded** — zero admissions for 120s, then an instant graceful dispose (nothing left in flight). Full evidence trail and static-analysis elimination in #2199.

The surviving mechanism is cross-request array aliasing after an `ArrayPool.Shared` double return of a `PendingResponse` batches/generations array (#2187 family; production windows tracked in #2195). `ReturnBatchesArray` was documented "must be called exactly once — not idempotent" yet had five call sites; e.g. an exception escaping `ProcessCompletedResponses` between the return and list compaction leaves the returned entry queued for re-processing and a second return.

## Changes

- **Exactly-once array return**: `PendingResponse` carries a per-request `ArrayReturnGuard` (reference type — the struct is copied by value, every copy must share the claim). All five cleanup paths funnel through `ReturnPendingResponseArrays`; the first call returns the arrays, later calls no-op and log a Warning. This kills the double-return → pool-poisoning family at the source and provides the telemetry to find any path that actually double-fires.
- **Release-loop robustness**: the post-write memory-release loop skips null slots (matching the sibling catch handlers' `batch is null ||` idiom) instead of NRE-ing past the remaining batches' releases — previously one corrupted slot orphaned every subsequent batch's reservation in the request.
- **Diagnosable timeouts**: buffer-allocation timeout messages now include in-flight and unsealed batch counts via a shared `BuildBufferTimeoutMessage` (both throw sites stay in sync). The observed failure would have read `in-flight batches: 0, unsealed batches: 0` — an immediate refund-leak signature instead of "network too slow".

## Tests

- `ArrayReturnGuard` claim semantics: first-wins, 16-way concurrent exactly-one-winner.
- `PendingResponse.TryReturnBatchesArray` cross-copy no-op contract (standalone instance — no reflective calls into live senders, per the #2187 lesson).
- Leak invariant: 20 acked waves through the scripted send loop, asserting `BufferedBytes` returns to 0 after every wave.
- Shared test helpers (`CreateTestBatch`/`CreateSuccessResponse`/`CreateSender`) hoisted into `ScriptedProduceResponseFixture`; 7 duplicate copies deleted across the sibling fixtures.

## Verification

- Unit suite: 5591/5591 (new tests looped 5x for stability).
- Backpressure integration category: 9/9 against a real broker via Docker — includes the originally failing test.
- Perf: the guard adds one small allocation + one `Interlocked.Exchange` per **request** (per batch group, not per message) — within the project's per-batch allocation budget; no stress lane needed.

## Explicitly out of scope (tracked in #2199)

Single-ownership consolidation of the five cleanup lifecycle paths and the return-then-compact ordering in `ProcessCompletedResponses` — the Warning log added here is the instrument for finding any real double-cleanup path before that refactor.

Also (named per review): a stale re-processed entry would re-fire its entry-level side effects that run before the array return — `_unackedBudget.OnAcked`/`NotePartialDeliverySuccess` and the user-facing `_onAcknowledgement` interceptor callback — which sit outside the per-batch incarnation guard. Tracked as an explicit item in #2199.

### DIFF
--- a/src/Dekaf/Producer/BrokerSender.cs
+++ b/src/Dekaf/Producer/BrokerSender.cs
@@ -170,7 +170,7 @@ internal sealed partial class BrokerSender : IAsyncDisposable
     /// at send start (before its first await) — the generations detect batch object
     /// recycling between send and response processing. Ownership of the rented
     /// <paramref name="BatchGenerations"/> and <paramref name="Batches"/> arrays transfers
-    /// to this PendingResponse; they are returned by <see cref="ReturnBatchesArray"/>.
+    /// to this PendingResponse; they are returned by <see cref="TryReturnBatchesArray"/>.
     /// </summary>
     private readonly record struct PendingResponse(
         PipelinedResponse<ProduceResponse> ResponseTask,
@@ -183,6 +183,15 @@ internal sealed partial class BrokerSender : IAsyncDisposable
         BrokerUnackedByteBudget.DeliverySnapshot DeliverySnapshotAtSend)
     {
         /// <summary>
+        /// One-time-return claim for the rented arrays. Lives in a reference type because
+        /// PendingResponse is a struct copied by value (lists, locals, sweeps) — every copy
+        /// must observe the same claim, or two copies could each return the arrays and
+        /// poison <see cref="ArrayPool{T}.Shared"/> with a double-returned array (#2187:
+        /// a poisoned pool hands the same array to two renters, cross-contaminating requests).
+        /// </summary>
+        private ArrayReturnGuard ReturnGuard { get; } = new();
+
+        /// <summary>
         /// Returns true if the batch at index <paramref name="i"/> is the same incarnation
         /// that was present when this PendingResponse was created.
         /// </summary>
@@ -192,13 +201,27 @@ internal sealed partial class BrokerSender : IAsyncDisposable
 
         /// <summary>
         /// Clears batch references and returns pooled arrays to <see cref="ArrayPool{T}.Shared"/>.
-        /// Must be called exactly once per PendingResponse entry — not idempotent due to struct copy semantics.
+        /// Only the first call (across all copies of this struct) returns the arrays; later
+        /// calls are no-ops that return false so the caller can log the breached
+        /// exactly-once cleanup contract instead of double-returning to the pool.
         /// </summary>
-        public void ReturnBatchesArray()
+        public bool TryReturnBatchesArray()
         {
+            if (!ReturnGuard.TryClaim())
+                return false;
+
             ArrayPool<int>.Shared.Return(BatchGenerations);
             ArrayPool<ReadyBatch>.Shared.Return(Batches, clearArray: true);
+            return true;
         }
+    }
+
+    /// <summary>Single-claim flag backing <see cref="PendingResponse.TryReturnBatchesArray"/>.</summary>
+    internal sealed class ArrayReturnGuard
+    {
+        private int _claimed;
+
+        public bool TryClaim() => Interlocked.Exchange(ref _claimed, 1) == 0;
     }
 
     private enum SendLoopEventType : byte
@@ -1992,7 +2015,7 @@ internal sealed partial class BrokerSender : IAsyncDisposable
                                     recoveredBatches, recoveredBatchSet);
                         }
                         pr.ResponseTask.Abandon();
-                        pr.ReturnBatchesArray();
+                        ReturnPendingResponseArrays(in pr);
                     }
                     Interlocked.Add(ref _totalPendingResponseCount, -pendingList.Count);
                     var removedBytes = SumEncodedBytes(pendingList);
@@ -2758,7 +2781,7 @@ internal sealed partial class BrokerSender : IAsyncDisposable
                         }
                     }
 
-                    pending.ReturnBatchesArray();
+                    ReturnPendingResponseArrays(in pending);
                     continue;
                 }
 
@@ -2847,7 +2870,7 @@ internal sealed partial class BrokerSender : IAsyncDisposable
                 if (response is ProduceResponse poolableResponse)
                     poolableResponse.Return();
 
-                pending.ReturnBatchesArray();
+                ReturnPendingResponseArrays(in pending);
             }
 
             // Compact: remove processed entries from the end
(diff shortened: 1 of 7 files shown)


thomhurst:
Addressed the review in 7d9dd5e8:

- **Finding 1 (fixed):** the `Acks.None` completion loop now carries the same `batch is null ||` guard as its three siblings, with the pool-poisoning rationale comment.
- **Finding 2 (scoped):** valid, distinct gap — the entry-level side effects that run before the array return for a stale re-processed entry (`_unackedBudget.OnAcked`/`NotePartialDeliverySuccess`, and the user-facing `_onAcknowledgement` interceptor callback) have no idempotency claim. Named explicitly in #2199's follow-ups now so it can't be lost inside the "single-ownership consolidation" framing. Not fixed here: the new `PendingResponse arrays already returned` Warning is the instrument for confirming whether the return-then-compact gap fires at all before adding more per-entry state.
- **Finding 3 (kept as-is, deliberately):** piggybacking the claim `int` on `PooledPipelinedResponse` would couple the arrays' return contract to that object's pool-reset lifecycle — a reset/re-rent racing the claim reintroduces exactly the class of cross-owner ambiguity this PR removes, and `ReadyBatch._returnedToPool` guards an object that *is* the pooled resource, which isn't the shape here (the claim protects two arrays owned jointly by copies of a struct). One self-documenting, directly-testable allocation per broker request buys an audit-proof invariant; taking the trade.
- **Finding 4 (fixed):** `PendingResponse` is now `internal`; the test constructs it directly and the reflection boilerplate is gone. The test also gained the stronger cross-copy assertion (claim shared through the reference-typed guard between two struct copies).

Unit suite 5591/5591 after the changes.

PR #2203, opened 2026-07-17
fix(producer): consolidate pending-response cleanup

### LINKED ISSUE #2195: BrokerSender: consolidate pending-response cleanup ownership; remaining double-return windows and wedged-loop array leak

thomhurst:
## Summary

- make the send loop the sole owner of pending-response cleanup
- claim entries before processing and compact tombstones afterward, preventing exception paths from exposing returned arrays to exit cleanup
- add deferred and recovered cleanup counters so wedged-loop leaks remain observable

## Root cause

Disposal and the send-loop exit path could both walk `_pendingResponsesByConnection`. Completed and timed-out responses also returned pooled arrays before list compaction, so an escaping exception could leave stale entries visible to exit cleanup.

## Validation

- `dotnet build src/Dekaf/Dekaf.csproj --no-restore`
- full net10.0 unit suite: 5,592 passed

Closes #2195

### DIFF
--- a/src/Dekaf/Diagnostics/DekafMetrics.cs
+++ b/src/Dekaf/Diagnostics/DekafMetrics.cs
@@ -53,6 +53,18 @@ internal static class DekafMetrics
             unit: "{fault}",
             description: "Producer completion-source faults isolated while completing messages.");
 
+    internal static readonly Counter<long> PendingResponseCleanupDeferred =
+        DekafDiagnostics.Meter.CreateCounter<long>(
+            "dekaf.producer.pending_response.cleanup.deferred",
+            unit: "{response}",
+            description: "Pending responses whose pooled arrays remained owned by a send loop after disposal timed out.");
+
+    internal static readonly Counter<long> PendingResponseCleanupRecovered =
+        DekafDiagnostics.Meter.CreateCounter<long>(
+            "dekaf.producer.pending_response.cleanup.recovered",
+            unit: "{response}",
+            description: "Deferred pending-response array cleanups later completed by the send loop; deferred minus recovered is still stranded.");
+
     // Consumer metrics
     internal static readonly Counter<long> MessagesReceived =
         DekafDiagnostics.Meter.CreateCounter<long>(
--- a/src/Dekaf/Producer/BrokerSender.cs
+++ b/src/Dekaf/Producer/BrokerSender.cs
@@ -192,6 +192,13 @@ internal sealed partial class BrokerSender : IAsyncDisposable
         private ArrayReturnGuard ReturnGuard { get; } = new();
 
         /// <summary>
+        /// True for the default-valued tombstone left in a pending list by
+        /// <c>ClaimPendingResponseAt</c>. Live entries always have Count &gt; 0
+        /// (SendCoalescedAsync never constructs an empty PendingResponse).
+        /// </summary>
+        public bool IsClaimed => Count == 0;
+
+        /// <summary>
         /// Returns true if the batch at index <paramref name="i"/> is the same incarnation
         /// that was present when this PendingResponse was created.
         /// </summary>
@@ -3081,8 +3088,9 @@ internal sealed partial class BrokerSender : IAsyncDisposable
     /// the connection and calls cancelInFlightRequests() for the node.
     /// <para/>
     /// Processing happens synchronously in the single-threaded send loop. Entries are
-    /// removed from the connection's pending list BEFORE processing, so ProcessCompletedResponses
-    /// cannot also process the same batches (eliminating the race condition).
+    /// claimed (tombstoned) in the connection's pending list BEFORE processing, so
+    /// ProcessCompletedResponses cannot also process the same batches (eliminating the
+    /// race condition), and the list is cleared once after the scan.
     /// <para/>
     /// The affected connection is also invalidated so the next send
     /// establishes a fresh connection. Response tasks from the old connection are orphaned —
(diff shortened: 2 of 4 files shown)


thomhurst:
Addressed the review in 3fd8bf54:

**O(n²) drains (blocking finding):** `HandleTimedOutRequests` and `CleanupPendingResponsesOnLoopExit` now use the same forward-scan `ClaimPendingResponseAt` + single `Clear()` pattern as `ProcessCompletedResponses` instead of repeated `RemoveAt(0)` — O(n) total, and the forward scan preserves per-partition FIFO redelivery order. The tombstone convention is now a named predicate (`PendingResponse.IsClaimed`) used at all four check sites, and the now-unused `RemovePendingResponseAt` helper is deleted.

**Test coverage gap (non-blocking finding):** added `ResponseProcessingFailure_WithOtherEntriesPending_CleansUpAllEntries` — drives the real send loop to two in-flight pending entries on one connection, completes the first response, and injects a throw from inside its processing (via the throttle-observation callback). Asserts the surviving entry is redelivered, both deliveries observe the failure, and `_totalPendingResponseCount` / deferred-cleanup count return to zero — the exact multi-entry claim/tombstone interleaving from #2195.

Validation: full unit suite 5593/5593 locally (Release). One unrelated observation while validating: `LockFreeStackTests.TryPush_TryPop_DoNotAllocateInSteadyState` failed once under the full parallel suite (7816 bytes allocated on its dedicated probe thread) and passed on immediate re-runs and in isolation — a pre-existing nondeterministic allocation-probe flake, untouched by this PR; may deserve its own issue. Possible follow-up from the same sweep: `FailPendingTransactionEnrollmentBatches` still front-drains a carry-over queue with `RemoveAt(0)` (sole-owner queue, no claim race, but same quadratic shape) — left out of scope here.

thomhurst:
Follow-up from the same sweep, added in c0578220: `FailPendingTransactionEnrollmentBatches` now drains each partition's carry-over queue with a forward scan + single `ClearQueue` (new `PartitionCarryOver` helper) instead of repeated `RemoveAt(0)`. The backlog there is bounded by BufferMemory rather than max-in-flight, so a deep retry backlog could stall the send loop quadratically during a transactional enrollment failure. The reorder also closes a latent lost-batch hazard: the old remove-then-finalize order could drop a batch (removed but not yet failed) if finalization threw, leaving its delivery task hanging with no exit-cleanup path — finalize and fail are now independently guarded and unprocessed entries stay queued for the loop-exit sweep. Covered by the five existing `SendLoop_TransactionEnrollment*` tests; full suite 5593/5593.

The LockFreeStack allocation-probe flake is now tracked in #2205.

thomhurst:
Round 4 feedback addressed in 7a573ac1:

**Blocking — finalize guard dropped by the helper extraction:** correct catch, thank you. `FailTransactionEnrollmentBatch` now guards `FinalizeFailedTransactionEnrollmentRetry` and `FailAndCleanupBatch` independently, exactly as suggested — restoring `c0578220`'s safety for the pending-queue path and raising the coalesced-array path to the same standard. The blast-radius argument (one unmute hiccup escaping to the send loop's outer catch and tearing down the whole sender) is what settles it; the loop-exit backstop I was leaning on only covers the still-queued entries, not the sender's lifetime.

**Round 4 minor:** restored the guard around the reusable response-lookup `Clear()` in `ProcessCompletedResponses`' finally so a throw can never skip the pooled-array returns below it.

**Round 2 minor:** `pending.ResponseTask.Abandon()` in `HandleTimedOutRequests` is now wrapped like its sibling in `CleanupPendingResponsesOnLoopExit`.

**Round 3 suggestion:** added `SendLoop_TransactionEnrollmentFailure_FailsEntireCarriedOverBacklog` — one partition accumulates a three-batch carry-over backlog behind a pending enrollment (later batches diverting via the 'O' diag path, no extra enrollment attempt), then a late enrollment failure must fail every queued batch with the enrollment error, leave `carryOver` drained, and keep the send loop alive.

Full unit suite 5594/5594 locally (Release).

PR #2217, opened 2026-07-18
perf(producer): ValueTask-native ProduceAllAsync — no per-message Task allocation or thread-pool hop

### LINKED ISSUE #2212: perf(producer): awaited-completion costs — thread-pool hop per acked completion; ProduceAllAsync allocates a Task per message
## Context

`ProduceBatch` 1000x1000 is the only producer benchmark not beating Confluent (ratio 1.04, huge variance). A 2026-07-18 audit traced the awaited-completion path and found two library-side costs; the produce hot path itself is allocation-free in steady state (pooled `PooledValueTaskSource`, arena-encoded records, thread-local serialization buffers).

## Cost 1: `RunContinuationsAsynchronously = true` forces a thread-pool hop per awaited completion

`RentCompletion` (`KafkaProducer.cs:1067-1073`) leaves `PooledValueTaskSource` at its default `RunContinuationsAsynchronously = true` (`PooledValueTaskSource.cs:21`). When `BrokerSender` acks a batch, `SetResult` schedules **one thread-pool work item per awaited message** — 1000 hops per benchmark iteration, plus `Task.WhenAll` fan-in. This scheduling burst + the associated garbage drives the observed variance and eats the CPU headroom that wins the smaller batch combos.

`SetRunContinuationsAsynchronously` already exists (`PooledValueTaskSource.cs:47`) but nothing calls it for the batched-await case.

Proposal: opt-in (builder flag or heuristic) synchronous continuations for produce completions, following the existing `WithInlineTransactionCompletions` precedent (`IKafkaProducer.cs:521-524`). Needs care: inline continuations run on the send loop thread — must not let user continuations block the sender (that is why the default is async). Possibly restrict to `Task.WhenAll`-style continuations or document the hazard.

## Cost 2: `ProduceAllAsync` calls `.AsTask()` per message

`KafkaProducer.ProduceAllAsync` (`KafkaProducer.cs:1642-1663`) converts every pending `ValueTask<RecordMetadata>` to a `Task` (~120-150 B each, 1000 allocations per 1000-message call). Since collecting raw ValueTasks for deferred await is forbidden by the pooled-source contract (`IKafkaProducer.cs:40-43`), `ProduceAllAsync` is the API users are told to use for fan-out — it should be the allocation-free path.

Proposal: make `ProduceAllAsync` ValueTask-native: e.g. a counting completion (single pooled `IValueTaskSource` decremented by each message's continuation) instead of N `Task` conversions + `WhenAll`. Kills ~150 B/msg and the WhenAll machinery for the batch-produce use case.

## Acceptance

- ProduceBatch 1000x1000 ratio clearly < 1.0 with the improved `ProduceAllAsync` benchmarked.
- No latency-percentile or throughput regression in the producer stress lane (Pareto rule, CLAUDE.md #8).

From the 2026-07-18 four-agent benchmark audit.

thomhurst:
Fixes #2212.

## Problem

Two library-side costs on the awaited-completion path (from the 2026-07-18 benchmark audit):

1. **One thread-pool hop per acked message**: `RentCompletion` left `PooledValueTaskSource` at `RunContinuationsAsynchronously = true`, so a 1000-message `ProduceAllAsync` scheduled 1000 thread-pool work items per iteration plus the `Task.WhenAll` fan-in.
2. **One `Task` allocation per message**: `ProduceAllAsync` converted every `ValueTask<RecordMetadata>` via `.AsTask()` (~120–150 B each).

## Change

**`ProduceAllCompletion`** (new, internal): a pooled counting `IValueTaskSource<RecordMetadata[]>`. Each registered produce operation attaches a pooled `PendingOp` continuation that stores its result into the results array and decrements a shared counter; the last completion fires the single aggregate continuation. Failure keeps the smallest-index exception, matching `Task.WhenAll` input-order semantics; every registered operation is always observed, so pooled per-message sources return to their pool even when the aggregate faults.

**Inline per-message continuations**: `ProduceAllAsync` now produces with `runContinuationsAsynchronously: false` — safe because the only continuation on each per-message completion is the bounded harvest (array store + decrement), following the `WithInlineTransactionCompletions` precedent. The aggregate source keeps `RunContinuationsAsynchronously = true`, so **user code never runs on the sender loop** — the 1000 hops collapse to one hop per batch.

**Dispatch unification**: the retry fork (`_retryPolicy`) and the topic/key/value feature predicate now exist once — a private `ProduceAsync(message, bool runContinuationsAsynchronously, ct)` plus the flag threaded through the topic/key/value dispatcher. Public produce passes `true`, transactions pass `!InlineTransactionCompletions`, `ProduceAllAsync` passes `false`.

**Steady-state allocations per `ProduceAllAsync` call**: results array (returned to caller) + async wrapper state machine — both per-batch. Per-message: zero (pooled `PendingOp`s, no interlocked ops in the registration loop; counter pre-set to `count + 1` with a registration guard).

**Benchmark**: `Dekaf_ProduceBatch` now uses `ProduceAllAsync` (per the issue's acceptance criteria), with the message array built in `GlobalSetup` so the measured body isolates the produce path.

## Known trade-offs

- Beyond 4096 concurrent pending registrations process-wide, excess `PendingOp`s allocate (~104 B each) — degradation back to roughly the old `AsTask()` cost, not worse. Capacity references `ValueTaskSourcePool.FallbackMaxPoolSize`.
- A synchronous produce throw mid-loop now surfaces after in-flight registrations complete (previously it propagated immediately, abandoning them); the same exception type is thrown.
- `tools/.../ProducerBenchmarks.cs` will conflict trivially with #2216 (`Dekaf_ProduceBatch` body); resolution: keep `ProduceAllAsync` + #2216's pre-created keys.

## Verification

- Unit: 5609/5609 pass (10 new `ProduceAllCompletionTests`: ordering, out-of-order async completion, mixed sync/async, smallest-index failure, sync-throw abort, pool reuse, 1000-way concurrent harvest, inline-continuation completion).
- Integration (local Docker, Kafka 4.3.1): ProduceAllAsyncTests 4/4, TopicProducerTests 15/15, ConvenienceApiTests 16/16, TransactionTests 5/5.
- Acceptance gates still to run in CI: ProduceBatch 1000x1000 ratio < 1.0 on the benchmark workflow, and a producer stress lane per the Pareto rule before merge if maintainer wants the paid validation.

### DIFF
--- a/src/Dekaf/Producer/IKafkaProducer.cs
+++ b/src/Dekaf/Producer/IKafkaProducer.cs
@@ -25,7 +25,10 @@ public interface IKafkaProducer<TKey, TValue> : IInitializableKafkaClient, IAsyn
     /// // Single message - await immediately (recommended)
     /// var metadata = await producer.ProduceAsync(message);
     ///
-    /// // Multiple messages in parallel - convert to Task (recommended)
+    /// // Multiple messages in parallel - ProduceAllAsync (recommended, no per-message Task)
+    /// var results = await producer.ProduceAllAsync(messages);
+    ///
+    /// // Alternatively, convert each ValueTask to Task
     /// var tasks = new List&lt;Task&lt;RecordMetadata&gt;&gt;();
     /// for (int i = 0; i &lt; count; i++)
     ///     tasks.Add(producer.ProduceAsync(message).AsTask());
@@ -131,8 +134,10 @@ public interface IKafkaProducer<TKey, TValue> : IInitializableKafkaClient, IAsyn
     /// </summary>
     /// <remarks>
     /// <para>This method is the recommended way to produce multiple messages in parallel.
-    /// It handles the <see cref="ValueTask{TResult}"/> to <see cref="Task{TResult}"/> conversion
-    /// internally, avoiding the error-prone pattern of storing ValueTask instances in collections.</para>
+    /// It awaits each message's <see cref="ValueTask{TResult}"/> internally through a single
+    /// counting completion, avoiding both the error-prone pattern of storing ValueTask instances
+    /// in collections and the per-message <see cref="Task{TResult}"/> allocations of
+    /// <c>AsTask()</c> + <see cref="Task.WhenAll(Task[])"/>.</para>
     /// </remarks>
     /// <param name="messages">The messages to produce.</param>
     /// <param name="cancellationToken">Cancellation token.</param>
--- a/src/Dekaf/Producer/KafkaProducer.cs
+++ b/src/Dekaf/Producer/KafkaProducer.cs
@@ -495,18 +495,18 @@ public sealed partial class KafkaProducer<TKey, TValue> : IKafkaProducer<TKey, T
     public ValueTask<RecordMetadata> ProduceAsync(
         ProducerMessage<TKey, TValue> message,
         CancellationToken cancellationToken = default)
-    {
-        if (_retryPolicy is not null)
-            return ProduceAsyncWithRetry(message, runContinuationsAsynchronously: true, cancellationToken);
-
-        return ProduceAsyncCore(message, runContinuationsAsynchronously: true, cancellationToken);
-    }
+        => ProduceAsync(message, runContinuationsAsynchronously: true, cancellationToken);
 
     internal ValueTask<RecordMetadata> ProduceTransactionAsync(
         ProducerMessage<TKey, TValue> message,
         CancellationToken cancellationToken)
+        => ProduceAsync(message, !_options.InlineTransactionCompletions, cancellationToken);
+
+    private ValueTask<RecordMetadata> ProduceAsync(
+        ProducerMessage<TKey, TValue> message,
+        bool runContinuationsAsynchronously,
+        CancellationToken cancellationToken)
     {
-        var runContinuationsAsynchronously = !_options.InlineTransactionCompletions;
         if (_retryPolicy is not null)
             return ProduceAsyncWithRetry(message, runContinuationsAsynchronously, cancellationToken);
 
@@ -683,6 +683,7 @@ public sealed partial class KafkaProducer<TKey, TValue> : IKafkaProducer<TKey, T
         Headers? headers,
         int? partition,
         DateTimeOffset? timestamp,
+        bool runContinuationsAsynchronously,
         CancellationToken cancellationToken)
     {
         if (_retryPolicy is not null || _interceptors is not null || Diagnostics.DekafDiagnostics.Source.HasListeners())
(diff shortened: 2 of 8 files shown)


thomhurst:
Rebased onto main (resolved benchmark conflict with #2216's pre-generated keys harness fix — batch messages now pre-built from `_keys` in `[GlobalSetup]`) and addressed review feedback in 36747ed2:

**Codex P2 (fault vs cancellation):** `ProduceAllCompletion.RecordFailure` now prefers a non-cancellation fault over `OperationCanceledException`, matching `Task.WhenAll` (any faulted child faults the aggregate regardless of cancelled siblings). Smallest index still breaks ties within the same class. Covered by three new unit tests (`Failure_FaultBeatsLowerIndexCancellation`, `Failure_LowerIndexCancellationDoesNotReplaceRecordedFault`, `Failure_AllCancellations_SurfacesCancellation`).

**Claude review (sync-throw coverage):** added producer-level tests exercising `KafkaProducer.ProduceAllAsync`'s actual catch block:
- Unit: both overloads on an uninitialized producer — synchronous throw at index 0 surfaces from the aggregate await without hanging (`KafkaProducerProduceAllTests`).
- Integration: `ProduceAllAsync_SyncThrowMidLoop_InFlightStillDeliveredAndExceptionSurfaces` — an `OnSend` interceptor cancels the token mid-registration; asserts messages appended before the throw still deliver, `OperationCanceledException` surfaces, nothing after the aborted loop is produced, and no hang. (Note: a *throwing* OnSend interceptor can't trigger this path — interceptor exceptions are swallowed by design — so the test cancels the token instead, which trips the entry `ThrowIfCancellationRequested` synchronously mid-loop.)

Validation: producer unit suite 1027/1027, `ProduceAllAsyncTests` integration 5/5 (Docker), full solution builds clean.

thomhurst:
Addressed Codex's second P2 (instrumented batch continuations) in 8165f0f8.

The `runContinuationsAsynchronously` bool is now a three-state `ProduceContinuationMode` threaded through the produce dispatch chain:

- **`Async`** — all public produce paths, unchanged.
- **`InlineWhenDirect`** — what `ProduceAllAsync` requests: inline is honored only while the bounded harvest is the pooled source's *direct* continuation. Whenever an instrumented or retry wrapper interposes its own state machine (`AwaitWithActivity`, `AwaitWithMetrics`, `ProduceAsyncWithRetry`), the mode is upgraded to `Async`, so metric listeners, activity export/dispose, and retry-policy code never run on the broker ack thread. The metrics check is hoisted per call so the mode decision and the wrapper selection can't disagree if a listener starts mid-call; `ProduceAsyncSlow` re-checks for the same reason. `AwaitWithCancellation` is not a wrapper (it registers the token and returns `completion.Task` directly), so the cancellation-only path keeps the inline win.
- **`Inline`** — the documented `InlineTransactionCompletions` opt-in keeps its contract (continuations on the sender thread, including through instrumented awaits). The existing `KafkaProducerFastPathTests` (`TransactionProduceAsync_CompletesContinuationInlineOnSenderThread`, `..._SequentialAwaitCanReenterProducerOnSenderThread`) pin this and caught my first, uniform version of the gate — scoping the upgrade to `InlineWhenDirect` keeps that opt-in intact.

New coverage: `ProduceAllAsync_WithTracingAndMetricsEnabled_AllDelivered` (integration) runs `ProduceAllAsync` under active `ActivityListener` + `MeterListener`, exercising the upgraded wrapper paths end-to-end.

Validation: full unit suite 5614/5614, `ProduceAllAsyncTests` integration 6/6, builds clean.

PR #2268, opened 2026-07-18
fix(protocol): reject unsupported API ranges

### LINKED ISSUE #2227: bug(protocol): reject absent or disjoint API version ranges
## Problem

`MetadataManager.GetNegotiatedApiVersion` does not compute a true intersection between client and broker version ranges.

Current logic:

```csharp
negotiated = Math.Min(ourMaxVersion, brokerVersions.MaxVersion);
negotiated = Math.Max(negotiated, ourMinVersion);
negotiated = Math.Max(negotiated, brokerVersions.MinVersion);
```

This can return a version unsupported by one side:

- broker `[0,3]`, client `[5,6]` returns `5` (broker does not support it)
- broker `[5,6]`, client `[0,3]` returns `5` (client does not support it)
- absent API key falls back to the client's minimum and callers may send a request the broker does not implement

Several call sites use the returned version directly without first calling `HasApiKey` or `SupportsApiVersion`.

Kafka's client model treats an absent API or empty intersection as `UnsupportedVersion`, not as a usable fallback:
https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/NodeApiVersions.java

## Required change

Compute the actual inclusive intersection:

```text
lowestUsable  = max(clientMin, brokerMin)
highestUsable = min(clientMax, brokerMax)

if lowestUsable > highestUsable:
    unsupported
else:
    use highestUsable
```

Missing broker information after negotiation must also be unsupported. Do not silently select a local minimum.

Use a typed Dekaf exception (`BrokerVersionException` or a dedicated unsupported-API exception) containing:

- API key/name
- client range
- broker range, or "API absent"
- target broker/connection when available

Validate invalid caller ranges such as `ourMinVersion > ourMaxVersion`.

## Acceptance criteria

- Disjoint ranges in both directions throw before serialization/write.
- Missing API key throws before serialization/write.
- Overlapping ranges select the highest common version.
- Exact single-version intersections work.
- Invalid client ranges are rejected.
- Optional feature probes have a non-throwing `Try...`/support API rather than depending on sentinel versions.
- All direct call sites are audited so unsupported APIs fail consistently.
- Focused unit tests cover every boundary above.

## Relevant code

- `src/Dekaf/Metadata/MetadataManager.cs`
- `src/Dekaf/Admin/AdminClient.cs`
- `src/Dekaf/Consumer/ConsumerCoordinator.cs`
- `src/Dekaf/Producer/KafkaProducer.cs`
- `src/Dekaf/ShareConsumer/ShareConsumerCoordinator.cs`
- `src/Dekaf/Telemetry/ClientTelemetryManager.cs`

## Dependency note

This fix can land before per-connection negotiation. The same intersection primitive should then be reused by the per-connection capability model.

thomhurst:
## Summary
- negotiate the true inclusive client/broker API-version intersection
- throw detailed `BrokerVersionException` failures for absent or disjoint APIs before request serialization
- add allocation-free `TryGetNegotiatedApiVersion` probes and use them for optional admin/telemetry capabilities
- cover overlap, exact-version, disjoint, absent, and invalid-range boundaries

## Test plan
- [x] `dotnet build tests/Dekaf.Tests.Unit --configuration Release`
- [x] focused `MetadataManagerTests` (31/31)
- [x] full net10 unit suite (5623/5623)
- [x] `dotnet build tests/Dekaf.Tests.Integration --configuration Release`
- [x] `ProtocolVersionTests` integration suite (11/11)
- [x] `/simplify`

## Performance

`ApiVersionNegotiationBenchmarks.GetNegotiatedApiVersion`, .NET 10.0.10, same machine/configuration, `[MemoryDiagnoser]`:

| Revision | Mean | Allocated |
|---|---:|---:|
| baseline `d9de72e6` | 1.446 ns | 0 B |
| candidate `da7597bf` | 1.476 ns | 0 B |

Delta: +2.1%; 99.9% confidence intervals overlap (`1.422-1.470 ns` vs `1.414-1.538 ns`), so no material throughput regression. Hot path remains zero-allocation.

Closes #2227

### DIFF
--- a/src/Dekaf/Admin/AdminClient.cs
+++ b/src/Dekaf/Admin/AdminClient.cs
@@ -4357,19 +4357,11 @@ public sealed class AdminClient : IAdminClient
         => RetryHelper.WithRetryAsync(operation, _metadataManager, cancellationToken);
 
     private bool SupportsApiRange(Protocol.ApiKey apiKey, short lowestSupportedVersion, short highestSupportedVersion)
-    {
-        if (!_metadataManager.HasApiKey(apiKey))
-            return false;
-
-        var apiVersion = _metadataManager.GetNegotiatedApiVersion(
+        => _metadataManager.TryGetNegotiatedApiVersion(
             apiKey,
             lowestSupportedVersion,
-            highestSupportedVersion);
-
-        return apiVersion >= lowestSupportedVersion &&
-               apiVersion <= highestSupportedVersion &&
-               _metadataManager.SupportsApiVersion(apiKey, apiVersion);
-    }
+            highestSupportedVersion,
+            out _);
 
     private static WriteTxnMarkersResponsePartition FindAbortTransactionPartition(
         WriteTxnMarkersResponse response,
--- a/src/Dekaf/Admin/AdminClient.cs
+++ b/src/Dekaf/Admin/AdminClient.cs
@@ -1532,12 +1532,10 @@ public sealed class AdminClient : IAdminClient
                 Topics = topicOffsets
             };
 
-            // Cap at v7 for admin offset commits — v8+ requires valid MemberId/GenerationId
-            // for group membership validation, which admin operations don't have.
             var apiVersion = _metadataManager.GetNegotiatedApiVersion(
                 Protocol.ApiKey.OffsetCommit,
                 OffsetCommitRequest.LowestSupportedVersion,
-                Math.Min(OffsetCommitRequest.HighestSupportedVersion, (short)7));
+                OffsetCommitRequest.HighestSupportedVersion);
 
             var response = await connection.SendAsync<OffsetCommitRequest, OffsetCommitResponse>(
                 request,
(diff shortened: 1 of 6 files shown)


thomhurst:
CI failure root cause fixed in 98afbce6. AlterConsumerGroupOffsetsAsync passed an impossible client range 8..7; it now negotiates Dekaf OffsetCommit support 8..9. Kafka defines v8 as the first flexible version and v9 as the same request for KIP-848; Java Admin topic-name requests likewise cap at v9: https://github.com/apache/kafka/blob/trunk/clients/src/main/resources/common/message/OffsetCommitRequest.json and https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/requests/OffsetCommitRequest.java. Validation: focused unit 1/1; full unit 5,624/5,624; Kafka 4.3.1 AdminWorkflowTests 20/20.

PR #2318, opened 2026-07-19
feat(serialization): introduce IAsyncSerde for per-message async serdes (#2309)

### LINKED ISSUE #2309: Feature request: Introduce IAsyncSerde
Use case: migration from Confluent.Kafka nuget package to Dekaf. Our current implémentation use a custom encryption / compression stack implemented in Serdes that require IO. This stack use short time living keys.

thomhurst:
Closes #2309.

## What

Introduces first-class asynchronous serde support for serializers/deserializers that must perform per-message I/O — the issue's driving use case is a Confluent.Kafka migration whose custom encryption/compression serde stack fetches short-lived keys:

- **New interfaces** (`Dekaf.Serialization`): `IAsyncSerializer<T>` (writes into an `IBufferWriter<byte>`), `IAsyncDeserializer<T>`, and combined `IAsyncSerde<T>`. Both carry a `CancellationToken` and document the pooled-memory validity contract (inputs/outputs only valid until the returned task completes).
- **Builder API**: overloads of the existing methods — `WithKeySerializer(IAsyncSerializer<TKey>)`, `WithValueSerializer(IAsyncSerializer<TValue>)`, `WithKeyDeserializer(IAsyncDeserializer<TKey>)`, `WithValueDeserializer(IAsyncDeserializer<TValue>)` — mirroring Confluent's same-name sync/async overload pattern for easy migration. Configuring both sync and async for the same component fails at `Build()`.
- **Producer**: when an async serializer is configured, every produce entry point (`ProduceAsync` ×3 incl. transactions/retry, `ProduceAllAsync` ×2, `FireAsync` ×3) diverts to a dedicated async path *before* any synchronous serialize: serialize into a pooled heap-safe buffer writer (`AsyncSerializationBufferWriter`, thread-cached, backed by `ProducerDataPool.BytePool`), then enqueue via the existing `EnqueueAppend` pooled-memory worker path (same machinery as the metadata-miss slow path). Fire-and-forget observes delivery in the background: errors reach the delivery handler or the FnF log, and the pooled completion source is always harvested. Mixed configurations (sync key + async value) use the existing thread-local sync pooled path for the sync component.
- **Consumer**: when an async deserializer is configured, `ConsumeAsync` and `ConsumeOneAsync` await deserialization per record before constructing the `ConsumeResult` (new internal pre-deserialized ctor; record fields are hoisted before the await because they reference pooled batch storage). `ConsumeBatchAsync` throws `NotSupportedException` (batch records are deserialized during synchronous iteration) with guidance to use `ConsumeAsync`/`ConsumeOneAsync`. Null semantics mirror the eager path exactly (null key skips the deserializer; null value invokes it with empty data + `IsNull = true`).
- Defense in depth: with an async serde configured, the sync serializer/deserializer slot holds a throwing placeholder, so any path that bypassed the async routing fails loudly instead of producing garbage.
- Docs: new "Async Serializers (IAsyncSerde)" section in `docs/serialization/custom.md`.

## Hot-path impact (Rule 1 / Rule 8)

The synchronous fast path is untouched except for **one readonly-bool branch per produce/consume entry point** (`_hasAsyncSerializers` / `_hasAsyncDeserializers`), the same shape and placement as the existing `_keyPreparer is not null` check. In `ConsumeAsync` the flag is hoisted once per stream like `hasInterceptors`. No allocations, no virtual dispatch, no async machinery added to the sync path.

Before/after `[MemoryDiagnoser]` numbers (broker-free, `--inProcess`, same machine):

`ProducerFireHotPathBenchmarks.FireBatch` (broker-free sync fire fast path — covers the exact entry point that gained the routing branch; sync serializers configured, so the branch is the only delta). Both runs `--inProcess` on the same machine (out-of-process BDN boilerplate build fails locally — pre-existing toolchain issue):

| Config | main (baseline) | this PR | Allocated |
|---|---|---|---|
| MessageSize=1000, DeliveryLatencyTargetMs=0 | 377.2 ns (median 370.1) | 334.2 ns (median 332.5) | 0 B → 0 B |
| MessageSize=1000, DeliveryLatencyTargetMs=10 | 888.3 ns (median 934.4) | 900.3 ns (median 930.7) | 0 B → 0 B |

Medians agree within ~1% (target-10) and the target-0 delta favors the candidate — parity within run-to-run noise, and the Allocated column stays `0 B`. The consumer-side `ConsumeOneBufferedFastPathBenchmarks` could not produce usable numbers locally (the in-process toolchain forces InvocationCount=1, yielding μs-scale noise); the consumer sync-path delta is one readonly-bool added to the existing `ConsumeOneAsync` fast-path condition plus a hoisted local in the `ConsumeAsync` loop — the same shape as the existing `hasInterceptors` hoist.

## Semantics notes

- With async serializers, `FireAsync`'s returned `ValueTask` still completes at append/enqueue (not delivery). BufferMemory backpressure surfaces via the background delivery observer (handler/log) rather than synchronously — same model the `ProduceAsync` metadata-miss slow path already uses.
- `IAsyncSerializerPreparer<T>` is unchanged and remains the right tool for one-time async setup (Schema Registry); the new interfaces are for genuinely per-message async work.

## Not in scope (follow-ups)

- `ConsumeBatchAsync` async-deserialization support (would need per-batch eager async materialization).
- ShareConsumer async serdes (builder intentionally has no async overloads yet).
- A Confluent-style `AsSyncOverAsync()` bridge wrapper.
- SchemaRegistry serializers implementing `IAsyncSerializer` natively.

## /simplify pass (applied)

Four parallel review agents (reuse / simplification / efficiency / altitude) ran on the diff; applied: the async path now reuses `ProduceInternalAsync` directly (per-component async branch at its serialize lines — the parallel ~90-line method was deleted), a shared `AwaitProduceCompletionAsync` tail replaces three copies of the instrumented-await ladder (adopted by `ProduceAsyncSlow` too), fire-and-forget appends via the accumulator's existing span/callback machinery instead of a per-message observer task + completion source, `ProduceInternalAsync` now checks the thread-local topic cache first (matters when every async-serde message routes through it), the new `ConsumeResult` ctor chains to the master ctor instead of hand-assigning 14 fields, the buffer writer got the `Array.MaxLength` growth clamp (drift from `DetachableBufferWriter`) and warm-buffer retention on reset, and the placeholder invariant is now enforced in the client constructors (not just the builder). Skipped deliberately: the `ConsumeOne` async sibling stays a documented mirror of the sync method (no zero-hot-path-cost sharing found), the `ProducerMessage` wrapper allocation in the arg-based async diverts (removing it would re-duplicate the produce path to save ~72 B on an already-allocating opt-in path), and a builder slot-union redesign (behavior-changing; the Build-time conflict throw covers it).

## Testing

- Unit (12 tests, all green): builder conflict validation ×4, async-serde build, `ConsumeBatchAsync` guard, placeholder throws ×2, `AsyncSerializationBufferWriter` round-trip/empty/growth/advance-past-end.
- Integration (6 tests, all green vs real Kafka via Testcontainers): XOR "encryption" serde with a forced async hop — `ProduceAsync`+`ConsumeAsync` round-trip (key+value, call-count-verified), `ConsumeOneAsync` (mixed sync/async), `ProduceAllAsync` ×50, `FireAsync`, `FireAsync` with delivery handler, null-value tombstone round-trip.
- Full solution builds clean (net10.0 + netstandard2.0), 0 warnings.
- Full unit suite run 6× total post-change: one run had a single failure whose identity was lost before the report was captured (overwritten by the subsequent run); 5 consecutive full-suite runs after it were 5873/5873 green, including 4 back-to-back capture runs that produced no failure to attribute. Flagging per the flaky-test rule rather than hiding it — if CI reproduces it, the name will pin it down.

### DIFF
--- a/docs/docs/serialization/custom.md
+++ b/docs/docs/serialization/custom.md
@@ -186,3 +186,60 @@ var consumer = await Kafka.CreateConsumer<string, Order>()
     .WithValueDeserializer(codec)
     .BuildAsync();
 ```
+
+## Async Serializers (IAsyncSerde)
+
+When serialization itself requires I/O — for example an encryption stack that fetches short-lived
+keys per message — implement `IAsyncSerializer<T>` / `IAsyncDeserializer<T>` (or the combined
+`IAsyncSerde<T>`):
+
+```csharp
+using Dekaf;
+using Dekaf.Serialization;
+
+public class EncryptingSerde : IAsyncSerde<Order>
+{
+    public async ValueTask SerializeAsync(
+        Order value, IBufferWriter<byte> destination,
+        SerializationContext context, CancellationToken cancellationToken = default)
+    {
+        var key = await FetchEncryptionKeyAsync(cancellationToken);
+        var payload = Encrypt(JsonSerializer.SerializeToUtf8Bytes(value), key);
+        destination.Write(payload);
+    }
+
+    public async ValueTask<Order> DeserializeAsync(
+        ReadOnlyMemory<byte> data,
+        SerializationContext context, CancellationToken cancellationToken = default)
+    {
+        var key = await FetchEncryptionKeyAsync(cancellationToken);
+        // `data` is pooled memory — only valid until this method's task completes.
+        return JsonSerializer.Deserialize<Order>(Decrypt(data.Span, key))!;
+    }
+}
+
+// Same builder methods — the async overload is picked by type
+var producer = await Kafka.CreateProducer<string, Order>()
+    .WithValueSerializer(new EncryptingSerde())
+    .BuildAsync();
+
+var consumer = await Kafka.CreateConsumer<string, Order>()
+    .WithValueDeserializer(new EncryptingSerde())
+    .BuildAsync();
+```
+
+Behavior and trade-offs:
+
+- **Prefer synchronous serializers** unless you genuinely need per-message async work. Async
+  serdes route every message through a slower pooled-buffer path with async state-machine
+  overhead; the synchronous zero-allocation fast path is unaffected when they are not configured.
+- For a **one-time async setup** (like a Schema Registry fetch) keep a synchronous serializer and
+  implement `IAsyncSerializerPreparer<T>` instead.
+- Producer: `ProduceAsync`, `ProduceAllAsync`, and `FireAsync` all work. Fire-and-forget delivery
+  errors go to the delivery handler (or the log) as usual.
+- Consumer: records are delivered via `ConsumeAsync` and `ConsumeOneAsync`. `ConsumeBatchAsync`
+  iterates records synchronously and throws `NotSupportedException` when an async deserializer is
+  configured.
+- Mixed configurations are fine — e.g. a synchronous key serializer with an async value serializer.
+- Configuring both a synchronous and an asynchronous serializer for the same component fails at
+  `Build()` with `InvalidOperationException`.
--- a/src/Dekaf/Builders.cs
+++ b/src/Dekaf/Builders.cs
@@ -65,6 +65,8 @@ public sealed class ProducerBuilder<TKey, TValue>
     private AwsMskIamConfig? _awsMskIamConfig;
     private ISerializer<TKey>? _keySerializer;
     private ISerializer<TValue>? _valueSerializer;
+    private IAsyncSerializer<TKey>? _asyncKeySerializer;
+    private IAsyncSerializer<TValue>? _asyncValueSerializer;
     private Microsoft.Extensions.Logging.ILoggerFactory? _loggerFactory;
     private ulong? _bufferMemory;
     private int? _maxBlockMs;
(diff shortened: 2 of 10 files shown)


thomhurst:
Fixed in `d2176fc5`.

Async serialization now acquires a producer-local slot before renting payload buffers. Capacity is derived from `BufferMemory / MaxRequestSize`; slot ownership transfers with `AppendWorkItem` and is released only after accumulator admission/append, enqueue failure, cancellation, or disposal. The append channels remain unbounded structurally, but can retain only the BufferMemory-derived number of serialized payloads. Slot waits honor caller cancellation and `MaxBlockMs`.

Regression coverage verifies capacity derivation and exactly-once worker release.

Validation:
- focused async serde/append/BufferMemory suites: 45/45 net10, 45/45 net8
- full units: 5874/5874 net10, 5747/5747 net8
- integration project build: 0 warnings/errors
- MemoryDiagnoser `AcquireReleaseUncontended`: 30.28 ns, 0 B allocated

@claude review

Issue #2231, opened 2026-07-18
feat(producer): implement Produce v13 topic-ID requests


thomhurst:
## Context

Dekaf implements Produce v9-v12 and therefore caps negotiation at v12. Kafka 4.3.1 supports Produce v13.

Produce v13 implements KIP-516 topic IDs:

- request topic names are replaced by topic IDs
- response topic names are replaced by topic IDs
- broker may return `UNKNOWN_TOPIC_ID`

Schema:
https://github.com/apache/kafka/blob/4.3.1/clients/src/main/resources/common/message/ProduceRequest.json

Dekaf already receives topic IDs in modern Metadata responses and uses them in Fetch. Produce remains name-based, so a Kafka 4.3 broker can advertise v13 but Dekaf will never select it.

## Required change

- Extend Produce request/response models and serializers through v13.
- Resolve application topic names to topic IDs from the current metadata snapshot before building v13 requests.
- Preserve application-facing topic names and delivery-report semantics; callers should not need to use IDs directly.
- Correlate v13 responses back to original topic/partition batches without dictionary allocation per message.
- Handle `UNKNOWN_TOPIC_ID` as a metadata-invalidating/retriable condition using existing retry budgets.
- Use v13 only when the actual target connection supports it; fall back to v12 for older supported brokers.
- Keep transactional Produce behavior correct for both TV1 and TV2.

## Performance constraints

Produce is a critical hot path.

- no per-message allocation
- no repeated topic-name-to-ID lookup per record when batching can resolve once
- no extra metadata RPC on cache hit
- response correlation must remain batch/partition scoped
- benchmark throughput, allocations, latency, and request size against v12

## Acceptance criteria

- v13 request/response snapshots match Kafka 4.3.1 schema.
- Multi-topic/multi-partition responses map back to correct delivery results.
- Metadata topic-ID change invalidates old mapping.
- `UNKNOWN_TOPIC_ID` refreshes metadata and retries without duplicate completion.
- v12 fallback works against Kafka 4.0-era advertised ranges.
- Transactional and non-transactional Produce paths covered.
- Fuzz corpus includes v13 response.
- Performance benchmark shows no material regression.

## Relevant code

- `src/Dekaf/Protocol/Messages/ProduceRequest.cs`
- `src/Dekaf/Protocol/Messages/ProduceResponse.cs`
- `src/Dekaf/Producer/BrokerSender.cs`
- `src/Dekaf/Metadata/ClusterMetadata.cs`

[...]

PR #2321, opened 2026-07-19
fix(producer): preserve pin on local ID miss

thomhurst:
## Summary
- tag the pre-send Produce v13 topic-ID cache miss with an internal exception type
- preserve the healthy pinned connection and avoid transport-failure logging for that local miss
- prove retry uses the same connection, sends only after metadata supplies the topic ID, and acknowledges once

Follow-up to the blocking current-head review on merged #2313 / #2231.

## Performance
The successful coalesced-send path gains no assignment, branch, allocation, or virtual dispatch. The only changed runtime behavior is the already-exceptional local topic-ID-miss path; it allocates the same single `KafkaException` object via a sealed internal subtype.

Attempted exact base `Dekaf_ProduceBatch` `[MemoryDiagnoser]` run on `11810b81`, but all four baseline cases produced no measurements: the Docker Kafka broker advertised the host-mapped port back inside the container, so topic creation and initial metadata timed out. The failed baseline was not repeated unchanged, and no candidate comparison is claimed.

## Validation
- Release unit build: 0 warnings/errors
- `BrokerSenderSendLoopTests` + `ReadyBatchIncarnationTests`: 81/81 net10.0; 81/81 net8.0
- full units: 5,940/5,940 net10.0
- full net8.0: 5,812 passed, 1 unrelated existing timing failure in `ConsumerProtocol_SteadyHeartbeat_CrossesMaxPollWhileAcquiringConnection_DoesNotSend`; investigation found its final `DidNotReceive` includes steady-heartbeat calls recorded before the measured phase
- Release integration build: 0 warnings/errors

### DIFF
--- a/src/Dekaf/Producer/BrokerSender.cs
+++ b/src/Dekaf/Producer/BrokerSender.cs
@@ -4087,8 +4087,11 @@ internal sealed partial class BrokerSender : IAsyncDisposable
         {
             // Send failed (connection error, timeout, etc.) — retry batches instead of permanently failing.
             // Aligned with Java Kafka's Sender: transient failures cause reenqueue for retry.
-            _pinnedConnections[connectionIndex] = null; // Invalidate only the broken connection
-            LogResponseFailed(ex, _brokerId);
+            if (ex is not LocalTopicIdMissException)
+            {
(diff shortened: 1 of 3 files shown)


thomhurst:
@claude review

PR #2323, opened 2026-07-19
test(consumer): isolate heartbeat phase

### LINKED ISSUE #2322: test(consumer): isolate pre-measurement heartbeat calls
## Failure

The full net8.0 unit suite failed `ConsumerProtocol_SteadyHeartbeat_CrossesMaxPollWhileAcquiringConnection_DoesNotSend` because its final `DidNotReceive` observed 81 steady-heartbeat calls made before the test's measured connection-acquisition phase.

## Root cause

`EnsureActiveGroupAsync` starts the coordinator heartbeat loop. `StopHeartbeatAsync` stops it, but the substitute retains calls already made. The test then installs its deterministic connection gate and asserts against the substitute's entire call history, so scheduling determines whether pre-phase calls cause failure.

## Expected fix

Clear received calls after `StopHeartbeatAsync` and before installing/releasing the connection gate, or assert only calls occurring after a deterministic phase boundary. Do not add delays or rerun as a flake workaround.

## Evidence

- net10.0 full suite in the same worktree: 5,940/5,940
- net8.0 full suite: 5,812 passed, 1 failed
- failure reported 81 matching `MemberEpoch > 0` calls

thomhurst:
## Summary
- clear substitute call history after the background heartbeat loop is stopped
- make the max-poll acquisition test assert only calls from its explicit measured heartbeat
- remove scheduling dependence without delays or retries

Closes #2322.

## Validation
- Release unit build: 0 warnings/errors
- exact test: 20/20 consecutive net8.0; 1/1 net10.0
- full units: 5,942/5,942 net10.0; 5,815/5,815 net8.0

Test-only change; no runtime or allocation impact.

### DIFF
--- a/tests/Dekaf.Tests.Unit/Consumer/ConsumerCoordinatorKip848Tests.cs
+++ b/tests/Dekaf.Tests.Unit/Consumer/ConsumerCoordinatorKip848Tests.cs
@@ -496,6 +496,8 @@ public sealed class ConsumerCoordinatorKip848Tests : IAsyncDisposable
         await using var coordinator = new ConsumerCoordinator(options, _connectionPool, _metadataManager);
         await coordinator.EnsureActiveGroupAsync(new HashSet<string> { "test-topic" }, CancellationToken.None);
         await coordinator.StopHeartbeatAsync();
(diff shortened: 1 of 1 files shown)


thomhurst:
@claude review

PR #2329, opened 2026-07-19
test(retry): make cancellation deterministic

### LINKED ISSUE #2326: Make RetryHelper backoff cancellation test deterministic
## Problem
`RetryHelperTests.CancellationDuringBackoff_StopsRetrying` uses `CancellationTokenSource(TimeSpan.FromMilliseconds(20))`. Under suite load, cancellation can race the best-effort metadata refresh failure. The catch filter then stops swallowing `InvalidOperationException`, so the test observes that exception instead of reaching the retry backoff.

Observed locally during PR #2320 validation:
- expected `OperationCanceledException`
- actual `InvalidOperationException: Failed to refresh metadata from any broker`
- same unchanged code passed an immediately preceding full run

## Required fix
Replace wall-clock cancellation with deterministic synchronization at the backoff boundary (for example cancel from `onRetry` before `Task.Delay`). Do not rerun and accept a pass.

thomhurst:
## Summary
- replace wall-clock cancellation in the RetryHelper backoff test
- cancel deterministically from the onRetry boundary after metadata refresh and before Task.Delay
- preserve the assertion that cancellation prevents a second operation attempt

## Validation
- unit build: 0 warnings/errors
- target test: 1/1 net10.0, 1/1 net8.0
- full unit: 5981/5981 net10.0, 5854/5854 net8.0
- /simplify complete

Closes #2326

### DIFF
--- a/tests/Dekaf.Tests.Unit/Retry/RetryHelperTests.cs
+++ b/tests/Dekaf.Tests.Unit/Retry/RetryHelperTests.cs
@@ -76,7 +76,7 @@ public sealed class RetryHelperTests
     public async Task CancellationDuringBackoff_StopsRetrying()
     {
         await using var metadataManager = CreateUnavailableMetadataManager();
-        using var cancellation = new CancellationTokenSource(TimeSpan.FromMilliseconds(20));
+        using var cancellation = new CancellationTokenSource();
         var attempts = 0;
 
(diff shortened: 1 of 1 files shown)


thomhurst:
@claude review\n@codex review

PR #2333, opened 2026-07-19
test(networking): tolerate timeout close

### LINKED ISSUE #2325: test(networking): tolerate expected timeout close
## Failure

Full net8.0 units intermittently fail `ConnectAsync_WhenNegotiationTimesOut_ThrowsKafkaException` because the mock server's synchronous `ReadExactly` throws `IOException: connection forcibly closed by remote host`.

## Root cause

The client is intentionally configured with a 50 ms negotiation timeout and closes the socket when that timeout fires. Even with the server on a `LongRunning` task, scheduling can let the expected client close happen before the server finishes reading the request frame. The test then awaits `serverTask` and surfaces this expected cleanup race as failure.

## Expected fix

Make the mock server tolerate the expected socket-close `IOException`/`EndOfStreamException`, while retaining assertions that the client throws `KafkaException(RequestTimedOut)` and remains disconnected. Do not add delay or rerun as a flake workaround.

## Evidence

- reproduced in full net8.0 suite after rebasing #2299 onto current main
- 5,885 other net8.0 tests passed
- full net10.0 suite passed 6,013/6,013
- #2299 has no diff in `KafkaConnectionCapabilityHandshakeTests.cs`

thomhurst:
Closes #2325.

The negotiation-timeout test again uses its intended 50ms deadline. Its mock server now tolerates the expected `IOException`/`EndOfStreamException` read failure when the timed-out client closes the socket, while the test still requires `KafkaException(RequestTimedOut)` and a disconnected client.

Validation:
- unit build: clean, 0 warnings/errors
- exact test: 5/5 net10.0 and 5/5 net8.0
- full handshake class: 14/14 net10.0 and 14/14 net8.0
- `/simplify`: complete

Test-only change; benchmark not applicable.

### DIFF
--- a/tests/Dekaf.Tests.Unit/Networking/KafkaConnectionCapabilityHandshakeTests.cs
+++ b/tests/Dekaf.Tests.Unit/Networking/KafkaConnectionCapabilityHandshakeTests.cs
@@ -472,11 +472,19 @@ public class KafkaConnectionCapabilityHandshakeTests
                 using var socket = listener.AcceptSocketAsync(cancellationToken)
                     .AsTask().GetAwaiter().GetResult();
                 using var stream = new NetworkStream(socket, ownsSocket: false);
-                var length = new byte[sizeof(int)];
-                stream.ReadExactly(length);
-                var frame = new byte[BinaryPrimitives.ReadInt32BigEndian(length)];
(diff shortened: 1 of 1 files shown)


thomhurst:
@claude review
@codex review

PR #2334, opened 2026-07-19
fix(producer): probe wave before deadline

### LINKED ISSUE #2327: Probe queued wave sibling before zero-linger deadline
## Problem
`BrokerSenderSendLoopTests.SendLoop_WaveCoalesce_RearmsAfterIdleWaitAtDefaultLinger` intermittently records only two width-2 requests instead of three. Reproduced 1 failure in 30 isolated net10 runs, and once under the full suite.

Root cause: at zero linger the quiet window is 75 microseconds. `onWaveCoalesceStarted` synchronously queues the sibling, then the loop timestamps its deadline. If the sender thread is preempted between timestamping and the first `while` condition, the deadline expires before the already-queued sibling is read. It is sent in a later request.

## Required fix
Guarantee at least one non-blocking channel probe after arming the wave, while retaining the hard bounds and zero-allocation hot path. Add deterministic regression coverage and `[MemoryDiagnoser]` / before-after performance evidence because this changes the send hot path.

thomhurst:
Closes #2327.

A wave now performs one non-blocking channel probe before checking its quiet deadline. This prevents a queued sibling from being split solely because the sender was preempted after arming the zero-linger deadline. The existing quiet/hard deadline condition is unchanged for every subsequent iteration.

## Validation
- unit build: clean, 0 warnings/errors
- exact deterministic regression: 1/1 net10.0, 1/1 net8.0
- `BrokerSenderSendLoopTests`: 71/71 net10.0, 71/71 net8.0
- benchmark build: clean, 0 warnings/errors
- `/simplify`: complete; final implementation uses `do/while`, adding no boolean/helper to later probes

## MemoryDiagnoser / performance

Same-run exact comparator, .NET 10.0.10, i7-12700K:

| Path | Mean | Allocated |
|---|---:|---:|
| Deadline-first expired skip (old) | 13.24 ns | 0 B |
| Probe-first queued read/write (new) | 22.52 ns | 0 B |

Cost: +9.28 ns once per armed request wave. Subsequent iterations retain the prior deadline condition exactly. No per-message allocation, no changed quiet/hard bounds, and no blocking operation.

### DIFF
--- a/src/Dekaf/Producer/BrokerSender.cs
+++ b/src/Dekaf/Producer/BrokerSender.cs
@@ -1508,14 +1508,14 @@ internal sealed partial class BrokerSender : IAsyncDisposable
                         _options.TransactionalId is not null))
                 {
                     waveCoalesceArmed = false;
-                    _onWaveCoalesceStarted?.Invoke();
                     var coalescedCountBeforeSpin = coalescedCount;
                     var started = Stopwatch.GetTimestamp();
                     var hardDeadline = started + waveCoalesceMaxTicks;
                     var quietDeadline = Math.Min(started + waveCoalesceQuietTicks, hardDeadline);
-                    while (coalescedCount < maxCoalesce
-                           && coalescedPartitions.Count < _knownPartitions.Count
-                           && Stopwatch.GetTimestamp() < quietDeadline)
+                    _onWaveCoalesceStarted?.Invoke();
+                    // Probe once before consulting the deadline. A queued sibling must not be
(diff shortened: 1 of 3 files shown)


thomhurst:
@claude review
@codex review

thomhurst:
Rebased onto current main to include #2333's pooled pending-request timeout-close stabilization, which addresses the net10 CI process crash (ManualResetValueTaskSourceCore.ValidateToken). Validation after rebase: Release net10 build clean (0 warnings), full unit suite 6,057/6,057 passed.

PR #2344, opened 2026-07-19
fix(consumer): survive live partition reassignment

### LINKED ISSUE #2247: Integration tests: partition reassignment mid-run with active producer and consumer
## Gap

Admin-side \`AlterPartitionReassignments\` is exercised in `AdminClientTests` / `MultiConnectionProducerTests`, but there is **no test with a producer/consumer actively running while partitions move between brokers** (replica set changes, leader moves onto a newly added replica). This is a different failure mode from the existing leader-handoff tests (`ProducerLeaderFailoverIntegrationTests`, `ConsumerLeaderFailoverIntegrationTests`), where the replica set is stable and only leadership changes.

## Scenario

On a 3-broker cluster:
1. Create topic (e.g. 4 partitions, RF=2), start a continuous producer and a consumer group consuming with commits.
2. Issue `AlterPartitionReassignments` moving several partitions' replica sets to different brokers (including moving the leader onto a replica that was not previously in the ISR).
3. Wait for reassignment completion while traffic continues.

## Proposed tests

- `Producer_PartitionReassignment_ContinuesWithoutLossOrDuplicates` — acked records all readable exactly once after reassignment completes; idempotent and non-idempotent variants.
- `Consumer_PartitionReassignment_ContinuesWithoutSkippedOrDuplicateRecords` — sequence oracle over consumed records across the reassignment window; committed offsets remain valid.
- `Producer_ReassignmentCancellation_RecoversCleanly` — cancel an in-flight reassignment, verify clients re-resolve leadership and continue.

Follow the existing failover-test pattern (sequence oracles, `NotLeaderOrFollower` retry paths). Reuse the 3-broker container infra from `ConsumerLeaderFailoverIntegrationTests`.

## Acceptance criteria

- Tests pass reliably (no timing-dependent flakiness — deterministic synchronization per CLAUDE.md).
- **Any client bugs surfaced by these tests (stale replica routing, retry storms, lost/duplicated records, stuck metadata) must be root-caused and fixed as part of this work — not deferred to follow-up issues.**

thomhurst:
Closes #2247.

## What changed

- adds deterministic 3-broker reassignment coverage for idempotent and non-idempotent producers
- verifies an active consumer group preserves per-partition sequences and committed offsets while four RF=2 partitions move from `[1,2]` to `[3,2]`
- verifies cancellation of a provably in-progress reassignment recovers cleanly
- fixes the client bug exposed by the consumer scenario: KIP-320 inline leader endpoints are now registered with connection pools before the new leader snapshot is published

Broker 3 is held offline until `ListPartitionReassignments` proves it is being added. Traffic then runs while the reassignment is active; broker 3 starts, becomes leader, and a final batch proves the same clients continue after leadership moves. Cancellation keeps broker 3 offline so the operation cannot complete before cancellation is observed.

Native GitHub `blockedBy` was checked before claim and again before publication; #2247 has no blockers.

## Validation

- Release integration build: 0 warnings/errors
- focused inline-leader unit regression: 1/1
- full unit net10: 6121/6121
- full unit net8: 5994/5994
- 3-broker partition reassignment class: 4/4 in 1m 19.98s

No benchmark was run: the source change executes only while handling a Fetch leader error and adds no work to successful fetch or per-message paths.

### DIFF
--- a/src/Dekaf/Metadata/MetadataManager.cs
+++ b/src/Dekaf/Metadata/MetadataManager.cs
@@ -832,9 +832,15 @@ public sealed partial class MetadataManager : IAsyncDisposable
         int leaderEpoch,
         NodeEndpoint? leaderEndpoint = null)
     {
+        if (leaderId < 0 || leaderEpoch < 0)
+            return false;
+
         BrokerNode? broker = null;
         if (leaderEndpoint is not null)
         {
+            if (leaderEndpoint.NodeId != leaderId)
+                return false;
+
             broker = new BrokerNode
             {
                 NodeId = leaderEndpoint.NodeId,
@@ -842,6 +848,11 @@ public sealed partial class MetadataManager : IAsyncDisposable
                 Port = leaderEndpoint.Port,
                 Rack = leaderEndpoint.Rack
             };
+
+            // Publish the route before the metadata snapshot can expose this broker as
+            // leader. Otherwise a concurrent fetch can observe the new leader ID while
+            // the connection pool still considers it unknown.
+            RegisterBroker(leaderEndpoint.NodeId, leaderEndpoint.Host, leaderEndpoint.Port);
         }
 
         return _metadata.TryUpdatePartitionLeader(topicName, partition, leaderId, leaderEpoch, broker);
--- a/tests/Dekaf.Tests.Integration/PartitionReassignmentIntegrationTests.cs
+++ b/tests/Dekaf.Tests.Integration/PartitionReassignmentIntegrationTests.cs
@@ -1,0 +1,487 @@
+using Dekaf.Admin;
+using Dekaf.Consumer;
+using Dekaf.Producer;
+
+namespace Dekaf.Tests.Integration;
+
+[Category("Admin")]
+[Category("Resilience")]
+[NotInParallel("RackAwareKafkaContainer")]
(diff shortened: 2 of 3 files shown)


thomhurst:
CI root cause identified; not rerunning unchanged. `KafkaConnection.DisposeCoreAsync` calls `FailAllPendingRequests`, queues the pending-request continuation, then immediately calls `ReturnAllPendingRequestsToPool`, whose reset increments the `ManualResetValueTaskSourceCore` token before the continuation consumes it. Linux CI hit that race and crashed in `PooledPendingRequest.OnCompleted` token validation.

PR #2345 addresses this exact pooled-reset race. This PR should update from `main` after #2345 lands, then run fresh CI.

thomhurst:
The failed net10 job was caused by the disposal race fixed on `main` by #2345, not this PR: `ManualResetValueTaskSourceCore.ValidateToken` → `PooledPendingRequest.OnCompleted` at `KafkaConnection.cs:5220` ([failed job](https://github.com/thomhurst/Dekaf/actions/runs/29690052490/job/88201070067)).

Rebased onto `f63930f2` so the exact causal fix is present. Post-rebase validation:
- unit + integration Release builds: 0 warnings/errors
- `ConsumerLeaderDiscoveryTests`: 24/24
- full unit net10: 6150/6150
- full unit net8: 6023/6023

Native `blocked_by` for #2247 was rechecked immediately before push: empty.

PR #2351, opened 2026-07-19
feat(consumer): support KIP-1092 close options

### LINKED ISSUE #827: KIP-1092: Consumer close options (leave group vs. silent shutdown)
## Summary

Implement KIP-1092's `CloseOptions` for consumer shutdown, allowing callers to control whether the consumer explicitly leaves its group on close. This is particularly useful for static membership where a graceful leave triggers an unnecessary rebalance.

**KIP**: https://cwiki.apache.org/confluence/display/KAFKA/KIP-1092

## Current State

Dekaf's consumer `DisposeAsync()` likely sends a leave-group signal (MemberEpoch = -1 in ConsumerGroupHeartbeat). There is no option to suppress this behavior.

### 1. CloseOptions Type

```csharp
public sealed class ConsumerCloseOptions
{
    /// <summary>
    /// Whether to explicitly leave the group on close.
    /// When false (and using static membership), the member slot is preserved
    /// and a new consumer with the same GroupInstanceId can rejoin without rebalance.
    /// Default: true.
    /// </summary>
    public bool LeaveGroup { get; init; } = true;

    /// <summary>
    /// Maximum time to wait for pending commits and leave-group.
    /// </summary>
    public TimeSpan Timeout { get; init; } = TimeSpan.FromSeconds(30);
}
```

### 2. Consumer API

Add an overload or option for controlled shutdown:

```csharp
// New method
ValueTask CloseAsync(ConsumerCloseOptions options, CancellationToken ct = default);

// DisposeAsync continues to use default behavior (leave group)
```

### 3. Static Membership Integration

When `GroupInstanceId` is set and `LeaveGroup = false`:
- Skip sending MemberEpoch = -1 heartbeat
- Commit pending offsets
- Close connections
- The broker preserves the member assignment for `session.timeout.ms`
- A new consumer with the same `GroupInstanceId` can rejoin seamlessly

## Testing Requirements

- Unit test: close with LeaveGroup=true sends leave heartbeat
- Unit test: close with LeaveGroup=false skips leave heartbeat
- Integration test: static member close without leave, rejoin with same instance ID, verify no rebalance

## Effort Estimate

Small.

## References

- [KIP-1092 Wiki](https://cwiki.apache.org/confluence/display/KAFKA/KIP-1092)

thomhurst:
## Summary
- add KIP-1092-style `ConsumerCloseOptions` with `Default`, `LeaveGroup`, and `RemainInGroup` membership operations
- preserve existing dynamic `-1` and temporary static `-2` close behavior while allowing permanent static leave via `-1`
- support silent shutdown without a terminal group heartbeat in real and in-memory consumers
- add unit and Kafka integration coverage for default, remain, and permanent-leave behavior

Closes #827

## Validation
- `dotnet build Dekaf.sln --configuration Release --no-restore` (0 warnings, 0 errors)
- net10 unit suite: 6,209 passed
- net8 unit suite: 6,082 passed
- focused post-rebase unit tests: 73 passed
- close-membership integration matrix: 3/3 net10 and 3/3 net8
- post-rebase close-membership integrations: 3/3 net10

No benchmark: close is a cold shutdown path; no per-message hot path changed.

### DIFF
--- a/src/Dekaf.Testing/InMemoryConsumer.cs
+++ b/src/Dekaf.Testing/InMemoryConsumer.cs
@@ -387,8 +387,22 @@ public sealed class InMemoryConsumer<TKey, TValue> :
             _storedOffsets[new TopicPartition(offset.Topic, offset.Partition)] = offset.Offset;
     }
 
-    public ValueTask CloseAsync(CancellationToken cancellationToken = default)
+    public ValueTask CloseAsync(CancellationToken cancellationToken = default) =>
+        CloseAsync(new ConsumerCloseOptions(), cancellationToken);
+
+    public ValueTask CloseAsync(
+        ConsumerCloseOptions options,
+        CancellationToken cancellationToken = default)
     {
+        ArgumentNullException.ThrowIfNull(options);
+        if (!Enum.IsDefined(options.GroupMembershipOperation))
+        {
+            throw new ArgumentOutOfRangeException(
+                nameof(options),
+                options.GroupMembershipOperation,
+                "The group membership operation is invalid.");
+        }
+
         cancellationToken.ThrowIfCancellationRequested();
 
         if (_disposed)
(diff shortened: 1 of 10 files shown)


thomhurst:
Addressed the blocking `Dekaf.Testing` finding. Because the in-memory cluster has no session-expiry model, close now always unregisters its member (documented inline) instead of allowing `RemainInGroup` to strand partitions forever. Added a regression test proving a second member receives and consumes the partition after a remain-mode close. Focused test passes; net10 build is clean.

PR #2354, opened 2026-07-19
fix(consumer): recover recreated topics

### LINKED ISSUE #2248: Integration tests: topic delete + recreate mid-run (topic ID change, offset reset)
## Gap

No integration test covers a topic being **deleted and recreated while clients are running** (zero hits for "recreate" in `tests/Dekaf.Tests.Integration`). On recreate, the topic gets a new topic ID and offsets reset to 0. Clients holding stale metadata must detect this — modern fetch/produce protocol versions carry topic IDs, and stale IDs must surface as `UNKNOWN_TOPIC_ID` and trigger metadata refresh, never silent misrouting or silent offset weirdness.

## Scenario

1. Producer + consumer group actively working a topic with committed progress at offset N.
2. Delete the topic, recreate it with the same name (same or different partition count).
3. Continue producing/consuming.

## Proposed tests

- `Producer_TopicRecreated_RefreshesTopicIdAndContinues` — producer with stale topic ID recovers via metadata refresh; no acked-but-lost records; records land in the recreated topic.
- `Consumer_TopicRecreated_DetectsOffsetResetAndAppliesAutoOffsetResetPolicy` — committed offset N now beyond the recreated log end; verify behavior matches `AutoOffsetReset` (earliest/latest) rather than stalling or throwing unhandled; verify no replay of phantom records.
- `Consumer_TopicRecreatedWithFewerPartitions_HandlesVanishedPartitions` — assignment shrinks cleanly, no permanent fetch loop against nonexistent partitions.
- `Commit_AgainstRecreatedTopic_DoesNotResurrectStaleOffsets` — offset commit safety across the ID change.

## Acceptance criteria

- Tests pass reliably with deterministic synchronization (poll DescribeTopics for the new topic ID before asserting; no bare delays).
- **Any bugs surfaced (stale topic-ID caching, silent stall, wrong offset-reset behavior, misdelivered records) must be fixed as part of this work — not deferred.**

thomhurst:
## Summary

- detect same-name topic recreation by immutable metadata snapshot and `TopicId`
- clear stale fetch/commit/watermark state, apply configured `auto.offset.reset`, and request group rejoin
- refresh metadata on `UNKNOWN_TOPIC_ID` / `UNKNOWN_TOPIC_OR_PARTITION` fetch errors
- cover producer recovery, consumer earliest/latest reset, partition shrink, and commit safety

Fixes #2248

## Correctness

- `TopicRecreationIntegrationTests`: 5/5 passed serially after rebase
- `Dekaf.Tests.Unit` net10.0: 6205/6205
- `Dekaf.Tests.Unit` net8.0: 6078/6078
- integration build: 0 warnings, 0 errors

The tests poll `DescribeTopics` until deletion and a different non-empty `TopicId` are observed. No fixed synchronization delay is used.

## Performance

The normal fetch-cycle path performs one immutable snapshot reference comparison. Assignment scans and state clearing run only after a metadata snapshot change; no per-message path changed.

`TopicIdentityCheckBenchmarks` (`[MemoryDiagnoser]`, .NET 10):

| Case | Mean | Allocated |
|---|---:|---:|
| Delegate control (pre-change equivalent) | 1.557 ns | 0 B |
| Unchanged metadata snapshot | 14.731 ns | 0 B |

A Docker-backed `ConsumerPollBenchmarks.Dekaf_PollSingle` predecessor run was attempted at `98f58090`, but produced no measurements because the benchmark container advertised its host-mapped listener internally; topic creation and consumer initialization timed out. It was not rerun unchanged.

### DIFF
--- a/src/Dekaf/Consumer/FetchSessionHandler.cs
+++ b/src/Dekaf/Consumer/FetchSessionHandler.cs
@@ -14,6 +14,9 @@ internal sealed class FetchSessionHandler
     private int _nextEpoch;
     private Dictionary<TopicPartition, CachedPartitionData> _sessionPartitions = [];
     private Dictionary<TopicPartition, CachedPartitionData> _lastDesiredPartitions = [];
+    private ClusterMetadataSnapshot? _sessionMetadataSnapshot;
+    private ClusterMetadataSnapshot? _responseMetadataSnapshot;
+    private ClusterMetadataSnapshot? _lastBuildMetadataSnapshot;
     private bool _lastBuildWasFull;
     private readonly SemaphoreSlim _gate = new(1, 1);
 
@@ -24,7 +27,19 @@ internal sealed class FetchSessionHandler
     public bool HasActiveSession => _sessionId != 0;
 
     public FetchSessionBuildResult Build(IReadOnlyList<FetchRequestTopic> desiredTopics, ClusterMetadata? clusterMetadata)
+        => BuildCore(desiredTopics, clusterMetadata?.CaptureSnapshot());
+
+    public FetchSessionBuildResult BuildFromSnapshot(
+        IReadOnlyList<FetchRequestTopic> desiredTopics,
+        ClusterMetadataSnapshot metadataSnapshot)
+        => BuildCore(desiredTopics, metadataSnapshot);
+
+    private FetchSessionBuildResult BuildCore(
+        IReadOnlyList<FetchRequestTopic> desiredTopics,
+        ClusterMetadataSnapshot? metadataSnapshot)
     {
+        _responseMetadataSnapshot = _sessionMetadataSnapshot;
+        _lastBuildMetadataSnapshot = metadataSnapshot;
         var desiredPartitions = FlattenDesiredPartitions(desiredTopics);
         _lastDesiredPartitions = desiredPartitions;
 
@@ -45,7 +60,7 @@ internal sealed class FetchSessionHandler
         var changed = new List<(string Topic, Guid TopicId, FetchRequestPartition Partition)>();
         foreach (var topic in desiredTopics)
         {
-            var topicName = ResolveTopicName(topic, clusterMetadata);
+            var topicName = ResolveTopicName(topic, metadataSnapshot);
             foreach (var partition in topic.Partitions)
             {
                 var tp = new TopicPartition(topicName, partition.Partition);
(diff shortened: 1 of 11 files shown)


thomhurst:
Addressed the blocking review on exact head `3d2ca03a`:

- Topic-recreation resets now start all genuinely asynchronous `ListOffsets` operations before awaiting them. Completed `ValueTask`s stay allocation-free; incomplete operations use a pooled `Task[]` and one `WhenAll`.
- `_observedTopicIds` is pruned when topics leave the assignment.
- Added tests against the real `KafkaConsumer` recovery path (protocol fakes): concurrent resets, vanished partitions, rejected-topic seeding, and unassignment pruning. `InMemoryConsumer` is a separate implementation, so it cannot exercise this code.
- Serialized the destructive topic-recreation integration class and paused the latest-reset consumer while recreating/priming the topic, eliminating the prefetch race exposed by validation.
- Stabilized three pre-existing timing-sensitive unit tests found during the full run; each failure was root-caused before changing the test.

Validation after rebase onto `07390a6f`:

- Release build: 0 warnings, 0 errors
- Unit: net10 `6259/6259`; net8 `6132/6132`
- Topic recreation integration: `5/5`
- Exact same-machine benchmark, immediate predecessor `510a2964` -> candidate `3d2ca03a`: `UnchangedMetadataSnapshot` `16.913 ns` -> `2.914 ns` (`-82.8%`), allocations `0 B` -> `0 B`

thomhurst:
Addressed all three exact-head findings in `6de35132`:

- Fetch v13+ response topics now resolve through the immutable metadata snapshot used to build the request, with the prior active fetch-session snapshot covering old IDs from incremental sessions. Unknown names are discarded before any watermark/preferred-replica state write.
- Topic identity errors are accumulated per response and deduplicated by topic, then refreshed in one forced metadata request. A deterministic two-partition test asserts exactly one request.
- Assignment publication now writes the unique assignment snapshot as an invalidation marker. The slow handler publishes its metadata snapshot with `CompareExchange`, so an assignment mutation during awaited resets cannot be overwritten. A blocked-reset race test verifies re-observation.

Validation:
- Release build: 0 warnings
- Unit: net10 `6262/6262`; net8 `6135/6135`
- Topic recreation integration: `5/5`
- `TopicIdentityCheckBenchmarks.UnchangedMetadataSnapshot`: exact predecessor `3d2ca03a` 2.914 ns / 0 B; candidate 2.536 ns / 0 B (`-13.0%`, allocation unchanged)

PR #2452, opened 2026-07-28
fix: validate bootstrap server endpoints

### LINKED ISSUE #2448: fix(metadata): bootstrap entries with PLAINTEXT://host:port prefix are silently dropped — parse scheme prefixes at BootstrapServerList and throw on garbage
## Summary

Dekaf's bootstrap parser silently drops entries in the `PLAINTEXT://host:port` form that both Apache Kafka's Java client (`Utils.getHost` regex) and librdkafka (documented optional `PROTOCOL://` prefixes) accept. The failure mode is a delayed `Failed to fetch initial metadata within 60000ms` timeout with no hint that the config format was the problem.

## Details

- `MetadataManager` (src/Dekaf/Metadata/MetadataManager.cs:134-150) parses each bootstrap entry by taking the **first** colon (`IndexOf(':')`), so `plaintext://127.0.0.1:61093/` yields host `plaintext` and an unparseable port — and the entry is then **silently skipped** rather than throwing.
- Found while investigating #2444: `Testcontainers.Kafka`'s `GetBootstrapAddress()` returns exactly this form, so every local Testcontainers benchmark run had the Confluent baseline connect while Dekaf timed out on metadata. PR for #2444 adds a benchmark-infra normalization as a stopgap (KafkaTestEnvironment); the repo already carries five other hand-rolled copies of the same normalization (`KafkaTestContainer.ExtractHostPort`, `SchemaRegistryTestContainer`, `TlsKafkaContainer`, `Dekaf.Profiling/Program.cs`, `Dekaf.StressTests/KafkaEnvironment.cs`) — all working around the product gap.

## Fix

1. Normalize in one place: `BootstrapServerList.FromValues` (src/Dekaf/BootstrapServerList.cs:11) is the existing choke point used by producer/consumer/share-consumer/admin builders (currently trim-only). Strip an optional `scheme://` prefix and trailing `/` there.
2. Make unparseable entries **throw** at construction instead of being silently skipped — a config typo should fail fast, not surface as a 60 s metadata timeout.
3. Tests: IPv6 bracket forms, plain `host:port`, `PLAINTEXT://host:port`, trailing slash, and (decision needed) whether a scheme that contradicts the configured `SecurityProtocol` should be rejected like librdkafka does.
4. Once landed, delete the six scattered normalizations (the #2444 PR's KafkaTestEnvironment comment references this issue for exactly that cleanup).

Client-visible behavior change (accepting new input forms + throwing on garbage), so it needs its own PR rather than riding the #2444 perf fix.

thomhurst:
## Summary
- normalize optional protocol prefixes and trailing slashes in `BootstrapServerList`
- validate DNS, IPv4, bracketed IPv6, and port ranges; malformed entries now fail during configuration
- keep scheme prefixes syntax-only; `UseTls` and SASL builder options remain authoritative
- delete duplicated Testcontainers bootstrap-address normalizers

## Performance
Construction/metadata initialization only; no serialization, produce, consume, or receive hot path changed.

## Test plan
- [x] `dotnet build --configuration Release --no-restore` (26 projects, 0 warnings/errors)
- [x] full unit suite (6,584 passed)
- [x] `Producer_SchemePrefixedBootstrapServer_ProducesSuccessfully` against Testcontainers Kafka

Closes #2448

### DIFF
--- a/src/Dekaf/BootstrapServerList.cs
+++ b/src/Dekaf/BootstrapServerList.cs
@@ -1,7 +1,13 @@
+using System.Globalization;
+using System.Net;
+using System.Net.Sockets;
+
 namespace Dekaf;
 
 internal static class BootstrapServerList
 {
+    internal readonly record struct Endpoint(string Normalized, string Host, int Port);
+
     internal static string[] FromCommaSeparated(string servers)
     {
         ArgumentNullException.ThrowIfNull(servers);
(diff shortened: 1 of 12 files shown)


thomhurst:
@
Fixed the `faults` lane failure in 38c74e97a. The failing classic-control test sampled group membership once after a fixed 18-second delay; CI caught the valid zero-member transition after session timeout. It now polls the required two-member condition for up to 45 seconds while still surfacing poll-task failures. Validation: integration build succeeded with 0 warnings/errors; targeted `ClassicCoordinatorFailover_HeartbeatsRediscoverAndConverge` passed under a real broker stop/recovery (1/1, 22.7s).
@

thomhurst:
@claude Please re-review the current head. The prior review found no actionable issues but omitted the merge gate's machine marker. If the head is clear, please include exactly `<!-- REVIEW_VERDICT: CLEAR -->`; otherwise report the actionable finding and use `BLOCKING`.

PR #2454, opened 2026-07-28
fix: separate append worker lifetime

### LINKED ISSUE #2449: producer: append workers share _senderCts — forceful dispose can abandon queued slow-path appends; separate lifetimes and add a test seam
## Summary

The slow-path append workers' lifetime token is the network senders' token (`_accumulator.StartAppendWorkers(_senderCts.Token)`, src/Dekaf/Producer/KafkaProducer.cs:481). Two consequences:

1. **Forceful dispose can abandon queued append work items.** `_senderCts.Cancel()` during dispose stops the worker `ReadAllAsync` loops; any unread `AppendWorkItem`s in the channels are left with their completion sources never set and their pooled key/value memory never returned. With the #2444 fix, awaited `ProduceAsync` calls under backpressure now land in these queues, so an abandoned item would hang an awaiting caller until their own timeout — previously only fire-and-forget callbacks could be affected.
2. **Test/benchmark helpers that "stop the sender loops" also kill the produce admission path.** At least five files reflection-cancel `_senderCts` assuming it stops only sender/linger loops (`KafkaProducerFastPathTests.StopProducerBackgroundLoopsAsync`, `HotPathAllocationTests.cs:289`, `ProducerCancellationTests.cs:70`, `ProducerAsyncPreparerTests.cs:250`, `ProducerFireHotPathBenchmarks.cs:150`, `TransactionalProduceAllocationBenchmarks.cs:46`). The #2444 PR's new test had to call `accumulator.StartAppendWorkers(CancellationToken.None)` to re-arm the workers — a working seam, but a symptom.

## Fix

- Give the append workers a linked-but-distinct CTS: cancelled only in the forceful phase of `DisposeAsync` **after** the channels are completed and drained (drain = fail remaining items' completions + return their pooled memory), never by sender-loop shutdown.
- Add an internal `StopSenderLoopsForTestingAsync()` seam next to the existing `CloseConnectionsForTestingAsync()` precedent, and migrate the reflection-based helpers to it.
- Verify with a disposal test: enqueue a backpressured awaited produce, dispose the producer, assert the awaiting task faults promptly (ObjectDisposedException) and no pooled memory leaks (existing leak-gate patterns from the produce-leak suite apply).

Surfaced by the /simplify altitude review on the #2444 PR.

thomhurst:
## Summary
- isolate append-worker cancellation from sender/linger-loop cancellation
- cancel append workers only during accumulator disposal, then fault unread completions, return pooled resources, and clear slow-path ownership
- add `StopSenderLoopsForTestingAsync()` and migrate all reflection-based test/benchmark helpers
- deterministically cover a blocked append plus an unread queued append during forceful disposal

## Performance
`ProducerFireHotPathBenchmarks` (`MessageSize=1000`, baseline → candidate → baseline → exact candidate repeat):

| DeliveryLatencyTargetMs | Baseline 1 | Candidate 1 | Baseline 2 | Candidate 2 | Allocated |
|---:|---:|---:|---:|---:|---:|
| 0 | 322.8 ns | 352.9 ns | 309.3 ns | 315.0 ns | 0 B |
| 10 | 937.4 ns | 932.2 ns | 934.2 ns | 936.0 ns | 0 B |

The first 0-ms candidate sample was a runner outlier; the exact repeat landed inside the bracketing baseline range. No message-path code changed and all rows remain zero-allocation.

## Test plan
- [x] full Release solution build (26 projects, 0 warnings/errors)
- [x] full unit suite (6,573 passed)
- [x] `AppendWorkerAffinityTests` (7 passed)
- [x] Testcontainers Kafka `Producer_DisposalAfterProduction_CompletesWithinReasonableTime`

Closes #2449

### DIFF
--- a/src/Dekaf/Producer/KafkaProducer.cs
+++ b/src/Dekaf/Producer/KafkaProducer.cs
@@ -28,6 +28,12 @@ public sealed partial class KafkaProducer<TKey, TValue> : IKafkaProducer<TKey, T
 {
     internal ValueTask CloseConnectionsForTestingAsync() => _connectionPool.CloseAllAsync();
 
+    internal async ValueTask StopSenderLoopsForTestingAsync()
+    {
+        await _senderCts.CancelAsync().ConfigureAwait(false);
+        await Task.WhenAll(_senderTask, _lingerTask).ConfigureAwait(false);
+    }
+
     /// <summary>
     /// Sentinel return value from <see cref="TryProduceSyncCore"/> indicating that the
     /// buffer is full and the caller should fall back to the async path (ReserveMemoryAsync).
@@ -72,6 +78,7 @@ public sealed partial class KafkaProducer<TKey, TValue> : IKafkaProducer<TKey, T
     private readonly bool _ownsInfrastructure;
 
     private readonly CancellationTokenSource _senderCts;
+    private readonly CancellationTokenSource _appendWorkerCts;
     private readonly Task _senderTask;
     private readonly Task _lingerTask;
     private readonly ConcurrentDictionary<Task, byte> _partitionEnrollmentTasks = new();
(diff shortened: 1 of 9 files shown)


thomhurst:
Addressed the unbounded test-helper shutdown thread in 194c03a45 by restoring the five-second WaitAsync bound. Validation: Release unit build succeeded with 0 warnings/errors; AppendWorkerAffinityTests + ProducerCancellationTests passed 17/17. Thread resolved; fresh CI running.

PR #2457, opened 2026-07-28
test(networking): detect inline request continuations deterministically

### LINKED ISSUE #2446: test(networking): RunContinuationsAsynchronously_IsRestoredAfterInlineUse asserts a timing property, not the thread-identity guarantee — flaky under full-suite load
## Summary

`PooledPendingRequestTests.RunContinuationsAsynchronously_IsRestoredAfterInlineUse` failed once in a full parallel unit-suite run locally (6571 tests, 1 failed at `PooledPendingRequestTests.cs:962`), then passed 5×34 in class isolation and 6571/6571 on an immediate full rerun. Per project policy this is a real bug — in this case in the test's assertion, not the product.

## Root cause

The test completes a pooled request whose `RunContinuationsAsynchronously` was restored to `true`, then asserts the continuation had not run by the time `TryComplete` returned:

```csharp
reused.TryComplete(buffer2);
var ranBeforeTryCompleteReturned = continuationRanInline;   // line 958
...
await Assert.That(ranBeforeTryCompleteReturned).IsFalse();  // line 962
```

`RunContinuationsAsynchronously = true` guarantees the continuation does not run **inline on the completing thread** — it does not guarantee it has not run *yet*. `SetResult` queues the continuation to the thread pool from inside `TryComplete`; under a busy pool (exactly the full-suite parallel condition), a worker can pick it up and set `continuationRanInline = true` concurrently, before line 958 reads the flag. The assertion encodes a timing property the API never promises, so it fails rarely and only under load.

## Fix

Assert thread identity instead of timing — that is the property `RunContinuationsAsynchronously` actually guarantees:

```csharp
var completingThreadId = Environment.CurrentManagedThreadId;
int continuationThreadId = -1;
vt.GetAwaiter().UnsafeOnCompleted(() =>
{
    continuationThreadId = Environment.CurrentManagedThreadId;
    continuationRan.SetResult();
});
reused.TryComplete(buffer2);
await continuationRan.Task;
await Assert.That(continuationThreadId).IsNotEqualTo(completingThreadId);
```

(If the pool inlines on the same thread via a queued work item that lands back on the completing thread after it becomes a worker, that's not possible here because the test thread is blocked awaiting `continuationRan` — the ID comparison is sound.)

Observed on branch at `36bbdcc14` + benchmark-only changes; the failing test and code under test are untouched by that branch. Test-only fix, no product change needed.

thomhurst:
## Summary
- replace the racy “continuation has not run yet” assertion with the guarantee actually provided by `RunContinuationsAsynchronously`
- detect only a continuation executing on the completing thread while `TryComplete` is still active
- allow valid queued execution on another thread before return, or later reuse of the same thread after return

Closes #2446.

## Why thread ID alone is insufficient

The issue proposed comparing continuation and completing thread IDs. That can falsely fail: after `TryComplete` returns and the async test yields, the same thread-pool thread may legitimately execute the queued continuation. Combining thread identity with a volatile “inside `TryComplete`” flag distinguishes actual inline execution from that valid scheduling case.

## Validation
- [x] Release unit build: 0 warnings/errors
- [x] `PooledPendingRequestTests`: 34/34 passed
- [x] full parallel unit suite: 6,571/6,571 passed
- [x] `$code-simplifier` final review

Test-only change; no product or hot-path code changed.

### DIFF
--- a/tests/Dekaf.Tests.Unit/Networking/PooledPendingRequestTests.cs
+++ b/tests/Dekaf.Tests.Unit/Networking/PooledPendingRequestTests.cs
@@ -934,32 +934,34 @@ public class PooledPendingRequestTests
         var reused = pool.Rent();
         reused.Initialize(responseHeaderVersion: 0, CancellationToken.None);
 
-        // Verify TryComplete returns BEFORE the continuation runs.
-        // With RunContinuationsAsynchronously = true, SetResult queues the
-        // continuation and returns immediately. With false, the continuation
-        // would run inline inside TryComplete, and the flag below would be
-        // true before TryComplete returns.
-        var continuationRanInline = false;
+        // Verify the continuation does not run inline on the completing thread.
+        // A queued continuation may legitimately run on another thread before
+        // TryComplete returns, or later on this same thread after it returns.
+        var completingThreadId = Environment.CurrentManagedThreadId;
+        var insideTryComplete = 1;
+        var continuationRanInline = 0;
         var continuationRan = new TaskCompletionSource();
 
         var vt = reused.AsValueTask();
         vt.GetAwaiter().UnsafeOnCompleted(() =>
         {
-            continuationRanInline = true;
+            if (Environment.CurrentManagedThreadId == completingThreadId
+                && Volatile.Read(ref insideTryComplete) != 0)
+            {
+                Volatile.Write(ref continuationRanInline, 1);
+            }
+
             continuationRan.SetResult();
         });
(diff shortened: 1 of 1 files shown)


thomhurst:
Rebased cleanly onto current main after #2452/#2454 merged. Current-head validation: Release unit build succeeded with 0 warnings/errors; `PooledPendingRequestTests` passed 34/34. Fresh full CI/review running.

PR #2505, opened 2026-08-04
docs: Redesign stress-tests page as at-a-glance comparison

thomhurst:
## Problem

The published [stress-tests page](https://thomhurst.github.io/Dekaf/docs/stress-tests/) had grown to **646KB** of raw engineering diagnostics — budget timelines, probe events, admission-block histograms, latency-outlier tables — burying the Dekaf-vs-Confluent comparison users actually come for.

## Change

Mirrors the benchmarks-page redesign (#2499): lead with verdicts, collapse the detail, drop maintainer-only diagnostics from the user-facing page.

**New page shape** (646KB → 26KB, regenerated from run 30928193941's merged results so it lands before the next weekly publish):
- **At a glance** — one row per scenario in plain language ("Produce — acks=all (3 brokers)"), geomean Dekaf vs Confluent msg/s, throughput verdict ("1.3× faster" / "on par", ±5% parity band), CPU-per-message verdict. 3conn control rows excluded from the headline.
- **Full results** — per-scenario throughput tables (order-balanced aggregates, CPU/request, drift/slope) collapsed into `<details>`; headings inside collapsed blocks demoted to bold so no ToC anchors point into closed content.
- **Correctness verdicts stay visible** — transaction verification and round-trip validation PASS/FAIL tables are not collapsed.
- GC table and the methodology bullet list collapsed.

**Where the diagnostics went:** nowhere — they were never removed from the workflow run summary, which still uses the unchanged `generate_scenario_tables` path. Only the docs page stops carrying them.

**Structural cleanup** (from /simplify review):
- Docs page assembly moved out of the workflow YAML heredoc into a testable CLI: `python3 .github/scripts/stress_report.py --results-dir results --output docs/docs/stress-tests.md` (mirrors `benchmark_docs.py`). Side effect: a malformed results file now fails the publish step instead of silently publishing a header-only page.
- Deleted the now-dead `format_comparison_callout`/`include_callout` path — the at-a-glance table supersedes the per-scenario admonitions.
- Unified heading suppression on the `title=None` convention; shared `median_cpu_micros`/`_describe_ratio` helpers instead of three copies.

## Validation

- Python script suite: **167/167 pass** (9 new tests: at-a-glance verdicts/parity/control-row exclusion, diagnostics omission, details collapse, visible correctness verdicts, full-page assembly, no-results placeholder).
- `docs && npm run build` passes; rendered page verified (14 collapsed sections, tables and `&` in summaries render correctly).
- Workflow YAML validated; docs step is now a one-liner.

No `src/` changes — no perf impact.


thomhurst:
CI note: the `faults` lane failure (`Abort_CoordinatorResponseDelayed_RecoversWithoutVisibility` — `InitProducerId failed after 10 retries`) is unrelated to this docs-only diff (zero C# touched). Root-caused and filed as #2507: InitProducerId's 10-attempt retry cap gives up after ~6.5s, shorter than cold transaction-coordinator startup (`__transaction_state` creation) on a loaded CI runner; the test died at the initial `InitTransactionsAsync`, before any toxic was applied. `consume`/`messaging` lanes were fail-fast cancellations, not failures. Re-running the failed jobs.

thomhurst:
Follow-up commit 63fa906: **at-a-glance verdicts are now trend-aware** — prompted by today's page reading "Produce — fire-and-forget: 1.1× slower".

**Investigation of that reading** (full detail in the commit message):
- Zero `src/` commits landed between the healthy 08-02 run (Dekaf 1.57M msg/s, 1.10× faster; 3conn best-ever 1.88M) and today's collapsed 08-04 run — only docs/CI changes. A code regression is impossible.
- Today's run is flagged `environmentShiftSuspected` by the trend gate: **both** clients dropped ~35-45% together, **both** clients' CPU/msg rose ~50%, and the confluent-first pair pinned to an identical ceiling (Confluent 826,511 vs Dekaf 826,522 msg/s — 11 msg/s apart). That's a VM-bound run, not a client gap.
- Trailing same-run ratios over the last 10 runs: 1.06–1.36 (median ~1.15). Dekaf is faster on trend; today was noise.

**The fix:** verdicts now use the median same-run ratio over the last 10 weekly runs (with a same-regime guard so lane reworks like the round-trip re-basis don't poison the median), a Confidence column, and a page-level caution when the latest run is environment-flagged. The msg/s columns stay latest-run and are labeled as such. Zero added runtime — history was already in the artifact.

PR #2042, opened 2026-07-13
perf(producer): fast-path single response

thomhurst:
## Summary

- bypass the topic-partition response dictionary when one ProduceResponse partition exactly matches one pending batch
- preserve the existing lookup path for multi-partition, mismatched, empty, error, and stale/null batch shapes
- cover exact-match eligibility and the existing stale-after-write end-to-end regression

## Performance

Identical local 2-minute, 1-broker transactional EOS runs (1,000 B messages):

| | CPU us/msg | CPU us/request | Median msg/s |
|---|---:|---:|---:|
| `main` | 653.78 | 653.27 | 65 |
| this branch | 568.91 | 568.47 | 65 |

The fast path removes 84.87 us/msg (13.0%) while preserving throughput. That absolute saving matches the issue's CI gap (358.38 vs 274.28 us/msg). EOS verification passed exactly: 5,900 committed delivered, 1,900 aborted leaked 0, duplicates 0, shortfall 0.

## Test plan

- [x] BrokerSender send-loop suite: 57/57
- [x] full net10 unit suite: 5,409/5,409
- [x] Release build: 0 warnings, 0 errors
- [x] local transactional EOS A/B and Confluent calibration

Closes #2037

Issue #2037, opened 2026-07-13
P2 perf: transactional CPU/msg still 1.31x Confluent (358 vs 274 us) after #2030 — per-request path at 1 msg/request


thomhurst:
## Summary

Stress run [29270211335](https://github.com/thomhurst/Dekaf/actions/runs/29270211335) (commit 072f640d): Transactional EOS after the #2030 send-loop trim:

| Client | msg/s | CPU µs/msg | Ratio |
|---|---|---|---|
| Dekaf | 411 (median 553) | **358.38** | 3.33x throughput |
| Confluent | 124 (median 166) | 274.28 | 1.00x |

CPU/msg improved 400 → 358 (#2030 worked, trend gate marked it Improvement), but Dekaf still burns **1.31x** Confluent's CPU per message. Successor to #2013 (closed via #2030), which identified the remaining levers: ~2 send-loop iterations and ~3 context switches per request.

## Data

- Request diagnostics: 1.20 KB/request across all 3 brokers → **1 message per request** — per-request path cost is the entire number. 358 µs/msg ≡ 358 µs/request.
- Progression: 382 µs (#1993) → 341 (#1968) → ~400 baseline → 358 now. Confluent sits at ~266–274. Gap ≈ 85 µs/request.
- Throughput is 3.33x Confluent, so this is purely an efficiency gap, but at 1 msg/request scale it bounds transactional throughput per core (0.20 standing cores at 550 msg/s → ~2,750 msg/s/core ceiling).

## Where the remaining ~85 µs/request lives (from #2013 analysis, still applicable)

- The request still traverses ≥2 send-loop iterations (wake → coalesce/dispatch, then response pass) with a thread-pool hop each.
- ~3 context switches per request: caller → send loop → receive loop → caller completion.
- Per-ack estimator suite runs at full fidelity for a 1.2 KB request (OnAcked EWMAs/histograms/probe update + RecomputeBudget per pass) — fixed cost that amortizes at 640 KB requests but is pure overhead at 1 msg/request.

## Suggested direction

1. Fast-path single-batch transactional dispatch: skip wave-coalesce/spin machinery when the txn has exactly one pending batch (it can never coalesce).
2. Complete the caller's ValueTask inline from the receive loop's response pass where safe (guarded by #2015's inline-continuation hazard — needs the async-continuation flag, not raw inline).
[...]

PR #2509, opened 2026-08-04
perf(networking): stamp receive-timeout deadline instead of re-arming timer per request

thomhurst:
## What

`AddPendingRequest` and `TryRemovePendingRequest` re-armed the receive-timeout `Timer` (lock + `Timer.Change`) on **every send and every response**. At in-flight depth ≤ 1 the pending count oscillates 0→1→0, so the timer was fully disarmed and re-armed per request — worst-case TimerQueue churn. A CPU trace of the transactional stress lane showed the TimerQueue timer thread as the busiest thread in the process by event count.

The connection already had a deadline-lazy path for byte progress (stamp `_receiveTimeoutDeadlineTimestamp`; the callback sleeps out the remainder when it fires early). This moves sends onto the same path:

- **Sends** stamp the deadline (two volatile writes) and arm the `Timer` only when it was dormant (CAS on a new `_receiveTimeoutArmed` flag).
- **Responses** no longer touch the timer at all; the armed callback observes zero pending requests and goes dormant.
- The dormant handoff uses full-fence interlocked ops on both sides: a send racing the callback either arms the timer itself or is observed by the callback's re-check, which reclaims the armed slot and loops to evaluate the fresh deadline. (The CAS deliberately stays a full-fence RMW even when it fails — a TATAS variant lets the deadline store and armed load reorder and can spuriously expire a live connection; documented in-code.)

Timeout semantics preserved: the deadline stays authoritative ("no byte progress for `RequestTimeout` while requests are pending"), moved forward by sends and byte progress exactly as before. Steady state: ~2 lock+`Timer.Change` per request → one no-op callback per `RequestTimeout` (default 30s) per connection.

## Why now

Investigating the transactional 3-broker lane's standing CPU gap vs Confluent (Dekaf ~400 vs ~274 µs/msg at 1 msg/request, stable 1.46–1.54× for a month — successor to #2037). Profiling decomposed it as ~77% per-request path over ~5 threadpool hops + ~23% standing overhead; the per-request `Timer` disarm/re-arm was the largest single unnamed cost. High-throughput lanes shed the same 2 `Timer.Change`/request (~3.6k/s at fire-and-forget request rates).

## Before/after

Local identical A/B, transactional EOS lane (2 min, 1000B, Dekaf-only, interleaved A-B-A-B-A-B on one machine — same instrument as #2042):

| CPU µs/msg | run 1 | run 2 | run 3 | median |
|---|--:|--:|--:|--:|
| `main` | 740.67 | 604.71 | 698.05 | **698.05** |
| this branch | 641.03 | 574.92 | 600.65 | **600.65** |

**−97 µs/msg median (−14%)**; candidate below baseline in 8/9 pairwise comparisons; median throughput identical (64–65 msg/s); allocations flat (~570 B/msg harness scope); **EOS verification exact PASS on all six runs** (committed delivered exactly, 0 duplicates/shortfall/leaks).

Proposed CI acceptance per Rule 7: single `producer-transactional-3b` lane, exact-SHA A-B-A with `baseline_sha=85c0efb5a` (immediately preceding product SHA), 15-minute default.

## Tests

- Unit suite 6591/6591 (net10.0); KafkaConnection suites 107/107 on net10.0 and net8.0.
- `SendAsync_CallerCanceledRequest_...` extended: late callback after drain goes dormant without expiring; a subsequent send re-arms; early fire with a future deadline stays armed without expiring.
- `ReceiveTimeout_StaleCallbackAfterRemoveAddCycle_DoesNotExpireLivePendingRequest` pins the new invariant (the old `RefreshReceiveTimeout` stale-zero-refresh hazard it guarded is structurally gone — the response path no longer touches the timer).
- /simplify review applied: single timer-arming path, invariant comments; TATAS proposal rejected as unsound (reasoning in-code).

Part of the transactional-CPU investigation; profile also identified linger-bound dispatch as the dominant throughput lever and ~150–350 B/msg unattributed allocations — follow-up issues to be filed separately.

### DIFF
--- a/src/Dekaf/Networking/KafkaConnection.cs
+++ b/src/Dekaf/Networking/KafkaConnection.cs
@@ -151,6 +151,7 @@ public sealed partial class KafkaConnection :
     private Timer? _receiveTimeoutTimer;
     private long _receiveTimeoutDeadlineTimestamp;
     private int _receiveTimeoutExpired;
+    private int _receiveTimeoutArmed;
     private OAuthBearerTokenProvider? _ownedTokenProvider;
     private readonly object _disposeGate = new();
     private Task? _disposeTask;
@@ -2458,7 +2459,7 @@ public sealed partial class KafkaConnection :
         }
 
         Interlocked.Increment(ref _pendingRequestCount);
-        RefreshReceiveTimeout();
+        StampReceiveTimeoutDeadline();
     }
 
     private bool TryGetPendingRequest(int correlationId, out PendingRequestEntry entry)
@@ -2487,8 +2488,8 @@ public sealed partial class KafkaConnection :
             if (remaining == 0 && Volatile.Read(ref _disposed) != 0)
                 Volatile.Read(ref _pendingRequestsDrained)?.TrySetResult();
 
-            RefreshReceiveTimeout();
-
+            // No timer work on the response path: the armed timer evaluates the
+            // stamped deadline when it fires and goes dormant if nothing is pending.
             ReleasePendingRequestSlot();
         }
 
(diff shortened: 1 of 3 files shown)


thomhurst:
First A-B-A ([run 30946393456](https://github.com/thomhurst/Dekaf/actions/runs/30946393456)): **INCONCLUSIVE**, but the detail strongly favors the candidate — every protected metric passed:

| Metric | B vs baseline mean | Gate |
|---|---:|---|
| Delivered throughput | **+2.88%** | pass |
| Median interval throughput | +2.27% | pass |
| Latency p50 / p95 / p99 | **−13.3% / −17.0% / −13.8%** | pass |
| CPU µs/msg | −1.91% | pass |
| Allocation B/msg | −6.62% | pass |
| Latency max | −26.6% | recorded |
| Steady/peak ratio | +14.26% (candidate best of 3) | inconclusive — **control drift 12.5% > 10%** |

The sole non-pass is the steady/peak gate where the baseline A→A2 control pair itself drifted 12.5% (0.796 → 0.903) — control noise on that metric, not a candidate signal; the candidate's 0.971 is the best of all three segments. The latency improvement is consistent with the mechanism: the response path no longer takes a lock + `Timer.Change` before completing acks.

Per Rule 7, dispatched the one permitted exact repeat: [run 30952716025](https://github.com/thomhurst/Dekaf/actions/runs/30952716025). If that also lands INCONCLUSIVE, I'll stop and synthesize both runs rather than dispatch again.

PR #2523, opened 2026-08-05
test: consolidate duplicate wait helpers into shared TestWait with descriptive timeouts

thomhurst:
## Summary

Follow-up to #2520: the integration test project had six near-duplicate wait/poll helpers, several with blind- or silent-failure shapes:

| Helper | Failure mode on timeout |
|---|---|
| `KafkaIntegrationTest.WaitForConditionAsync` (bool) | bare `TaskCanceledException`, no context |
| `AdminClientTests` / `ConcurrentAdminTests` / `DeleteRecordsTests` / `PartitionExpansionTests` private `<T>` copies | **silently returned the last unmet result** — call sites that discard the return value could pass with the condition never met |
| `RebalanceEdgeCaseTests` private void copy | **silently returned** on timeout |

## Changes

- New `internal static class TestWait` holds both helpers. It lives outside the base class deliberately: several polling test classes (rack-aware, scale-out, ACL fixtures) cannot derive from `KafkaIntegrationTest`, so a `protected` home could never absorb their copies. `KafkaIntegrationTest` keeps thin `protected static` forwarders — no call-site churn.
- Both helpers throw `TimeoutException` with an optional `description`; the generic overload includes the last observed value, so a stuck test names the state it was stuck in (the failure mode #2520 fixed case-by-case).
- The sync overload clamps its final poll delay so the timeout is not overshot by a poll interval.
- `ManualAssignment_DoesNotAutoDetect_NewPartitions` was using a condition wait as a *negative-check* grace window ("assignment must NOT grow") and only passed because of the silent unmet-return — the throwing helper caught this immediately. It now polls the invariant throughout the 3 s observation window (same idiom as `OffsetCommitSafetyTests`), which is strictly stronger: a transient wrong auto-detection now fails at the violation instant.
- `ConsumerCoordinatorFailoverIntegrationTests.WaitForConvergenceAsync` intentionally stays separate: it additionally watches background consume tasks and rethrows their real fault instead of masking a crash as a timeout.

Behavior note: the shared `<T>` default is `maxRetries: 10` (the old copies were split 5/10). This lengthens only the *failing* path; keeping the larger budget avoids shrinking DeleteRecordsTests' propagation patience.

Hand-rolled same-mechanism waits that can now migrate to `TestWait`: `ShareConsumerTests.WaitForShareAssignmentAsync`, `QuotaAndRateLimitingTests.WaitForProducerQuotaAsync` (still has the bare-cancellation failure mode), the byte-identical `WaitUntilAsync` pair in `BrokerScaleOutIntegrationTests`/`PartitionReassignmentIntegrationTests`, `AclEnforcementTests.WaitForAclCountAsync`, `DynamicConfigurationIntegrationTests.WaitForTopicConfigAsync`.

## Test plan

- [x] `dotnet build tests/Dekaf.Tests.Integration -c Release -p:TreatWarningsAsErrors=true` — 0 warnings/errors
- [x] Local Docker runs: AdminClientTests 32/32, ConcurrentAdminTests 9/9, DeleteRecordsTests 5/5, PartitionExpansionTests 5/5 (+1 broker-version skip), RebalanceEdgeCaseTests 5/5

No `src/` changes — no performance surface touched.

### DIFF
--- a/tests/Dekaf.Tests.Integration/AdminClientTests.cs
+++ b/tests/Dekaf.Tests.Integration/AdminClientTests.cs
@@ -21,27 +21,6 @@ public class AdminClientTests(KafkaTestContainer kafka) : KafkaIntegrationTest(k
             .Build();
     }
 
-    /// <summary>
-    /// Waits for a condition to become true with exponential backoff.
-    /// Admin operations in Kafka have eventual consistency - changes may not be immediately visible.
-    /// </summary>
-    private static async Task<T> WaitForConditionAsync<T>(
-        Func<Task<T>> check,
-        Func<T, bool> condition,
-        int maxRetries = 5,
-        int initialDelayMs = 500)
-    {
-        T result = default!;
-        for (var i = 0; i < maxRetries; i++)
-        {
-            await Task.Delay(initialDelayMs * (i + 1)).ConfigureAwait(false);
-            result = await check().ConfigureAwait(false);
-            if (condition(result))
-                return result;
-        }
-        return result;
-    }
-
     #region DescribeConfigs Tests
 
     [Test]
--- a/tests/Dekaf.Tests.Integration/ConcurrentAdminTests.cs
+++ b/tests/Dekaf.Tests.Integration/ConcurrentAdminTests.cs
@@ -23,27 +23,6 @@ public class ConcurrentAdminTests(KafkaTestContainer kafka) : KafkaIntegrationTe
             .Build();
     }
 
-    /// <summary>
-    /// Waits for a condition to become true with linear backoff.
-    /// Admin operations in Kafka have eventual consistency.
-    /// </summary>
-    private static async Task<T> WaitForConditionAsync<T>(
-        Func<Task<T>> check,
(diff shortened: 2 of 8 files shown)


thomhurst:
Pushed a second commit fixing the `AwaitedProduce_BufferMemoryBackpressure_CompletesThroughBrokerAck` failure that hit this PR's CI run. It is a pre-existing main flake, not introduced here — main's own run [31018577925](https://github.com/thomhurst/Dekaf/actions/runs/31018577925) (net8.0 produce leg, commit ec0b9bab8) failed with the identical assertion.

Root cause: the test spin-waits on the drain *entry* counter, which fires exactly when the drain owner has temporarily moved every pending candidate into its private admission snapshot — the moment `_pendingAppends` transiently reads 0. `RecordAccumulator.cs` documents that an empty queue is only conclusive when no drainer owns the snapshot; the bare `Count == 1` assert read inside that window. Fixed with a bounded wait on the parked state (which also dogfoods this PR's new helper). Verified 6/6 green locally against Docker.

Issue #2352, opened 2026-07-19
deps(snappy): expose reusable block codec lifecycle upstream


thomhurst:
## Context

#2214 requires eliminating Snappier's 48 B compression and 80 B decompression driver allocations. Snappier 1.3.1 exposes only static block APIs; its reusable `SnappyCompressor` and `SnappyDecompressor` types are internal.

Local reflection/unsafe access is not acceptable: Dekaf supports Native AOT and trimming. Vendoring the codec would create an unnecessary maintenance/security fork.

## Prerequisite

Upstream a supported reusable block-compression lifecycle, including:

- reusable compressor with retained working memory;
- reusable decompressor with reset and replaceable `IBufferWriter<byte>` destination;
- documented non-concurrent/thread-local usage;
- Native AOT and trimming compatibility.

Then consume a released Snappier package containing that API.

## Acceptance

- Upstream change merged and released.
- Dekaf package reference updated.
- Correctness suite passes for single/multi-segment and malformed inputs.
- `CompressionBenchmarks` reports 0 B for Snappy compress/decompress without materially regressing throughput/latency.

thomhurst:
Upstream reusable lifecycle proposed in https://github.com/brantburnett/Snappier/pull/148. Validation: net48 268/268; net8/net9/net10 284/284; package/API compatibility and AOT analyzers clean. BenchmarkDotNet DefaultJob: compression 34.81 us/50 B -> 35.16 us/0 B (+1.0%, same rank); decompression 10.355 us/81 B -> 9.411 us/0 B (-9.1%). Waiting for upstream merge and release before Dekaf package work.

thomhurst:
Blocker rechecked 2026-08-17: upstream brantburnett/Snappier#148 remains open and unreviewed; latest release is still 1.3.1 (2026-04-28). No reusable public codec lifecycle is available, so Dekaf cannot proceed without the explicitly rejected reflection/vendoring options.

thomhurst:
Blocker rechecked 2026-08-18: upstream brantburnett/Snappier#148 remains open; no reusable public codec lifecycle has been released. Dekaf package work remains blocked without the rejected reflection or vendoring paths.

thomhurst:
[...]

Issue #2214, opened 2026-07-18
perf(protocol): last zero-alloc leftovers — pool DetachableBufferWriter in PreCompress; reuse Snappier compressor/decompressor instances


thomhurst:
## Context

A 2026-07-18 audit traced every remaining nonzero Allocated figure in the unit benchmark tables. Most are harness artifacts or inherent (returned strings). Two genuine library allocations remain:

## 1. `DetachableBufferWriter` wrapper in `RecordBatch.PreCompress` (~40 B per batch seal)

`RecordBatch.cs:703` (CompressionType.None path), also `:719` and `:737` (compressed branches):

```csharp
using var encodedBuffer = new DetachableBufferWriter(ProducerDataPool.BytePool, ...);
```

`DetachableBufferWriter` (`RecordBatch.cs:189`) is a sealed class newed per `PreCompress` call. Its byte buffer is pooled; only the ~40 B wrapper object escapes. Per-batch, so cheap — but it is the only allocation left on the producer seal path. Pool or thread-cache the wrapper (same pattern as the pooled `SerializationCache`, `RecordBatch.cs:326`), or restructure so detach works without a heap wrapper.

## 2. Snappier static helpers allocate a compressor/decompressor object per call (48 B / 80 B per op)

`SnappyCompressionCodec.cs:75` calls `Snappier.Snappy.Compress(span, span)`; `:158` calls `Snappy.Decompress(...)`. The third-party static helpers internally `new` a `SnappyCompressor`/`SnappyDecompressor` per call (their working memory is ArrayPool-rented; only the small driver object escapes — matches the benchmark's 48 B compress / 80 B decompress exactly).

Dekaf's own wrapper state is already `[ThreadStatic]`-cached (`:32`, `:171-186`). Fix: hold `[ThreadStatic]` `SnappyCompressor`/`SnappyDecompressor` instances and use Snappier's instance APIs instead of the allocating statics (verify the instance API supports reset/reuse; otherwise upstream a pooling hook).

## Outcome

With both fixed (and the benchmark-harness fixes landing separately), every protocol/compression unit-table row except the inherent deserialized string reads `-` (zero allocation). Per-batch impact is small but this closes out the "zero allocation where possible" goal for the wire path.

From the 2026-07-18 four-agent benchmark audit.

thomhurst:
[...]

PR #2524, opened 2026-08-05
test: pin Snappy codec reuse contract ahead of #2352; document upstream allocation blocker

thomhurst:
## Summary

Investigated #2214 ("last zero-alloc leftovers") to implementation. Findings changed the shape of the work:

| #2214 item | Status |
|---|---|
| Pool `DetachableBufferWriter` in `PreCompress` | **Already shipped on main** — PR #2416 (2026-07-27) added the `Rent` + `[ThreadStatic]` single-slot wrapper cache; producer pre-serialization measured 40 B → 0 B there. Nothing left to do. |
| Reuse Snappier compressor/decompressor instances | **Blocked upstream.** Snappier 1.3.1 keeps `SnappyCompressor`/`SnappyDecompressor` `internal`; every public static entry point allocates a fresh driver per call (~48 B compress / ~80 B decompress). #2352 rules out reflection and vendoring (Native AOT + trimming). Upstream brantburnett/Snappier#148 is still open — the swap needs it merged and released, then a package bump. |

So no product perf change is possible right now. This PR lands what is useful while blocked:

- **Source comment** in `SnappyCompressionCodec` classifying the residual allocations as per-block (per-batch) costs with the upstream blocker, so future zero-alloc audits don't misdiagnose them or "fix" them the forbidden way.
- **Contract-pinning tests** the eventual #2352 swap must keep green:
  - `Compress_ReusedCodec_ProducesIdenticalOutputToFreshInstance` — thread-cached scratch never leaks across compress/decompress cycles.
  - `CompressDecompress_ParallelWorkers_DoNotCorruptEachOther` — 8 workers × 50 batches through one shared codec, single- and multi-segment inputs; catches ownership bugs if thread-local caches are ever converted to a shared pool incorrectly.
  - `Rent_WhileOutstanding_ReturnsIndependentWriter` — pins `DetachableBufferWriter`'s documented single-slot cache invariant from #2416.

Reviewed via /simplify (4 agents): segment-building now delegates to the shared `SequenceTestHelpers`, writers pre-sized, comment trimmed to durable facts. Efficiency review specifically verified the parallel test cannot perturb the pool-warmth-sensitive `TransactionalProduceAllocationTests` gate (NotInParallel + disjoint pool buckets + its windowed design).

## Test plan

- [x] `dotnet build Dekaf.sln -c Release -p:TreatWarningsAsErrors=true` — 0 warnings/errors
- [x] Full unit suite green; new tests pass (Snappy class 20/20, DetachableBufferWriter class 3/3)

No behavioral `src/` change (comment only) — no benchmark numbers required. Suggest keeping #2214 open (or re-scoping it onto #2352) until upstream ships.

### DIFF
--- a/src/Dekaf.Compression.Snappy/SnappyCompressionCodec.cs
+++ b/src/Dekaf.Compression.Snappy/SnappyCompressionCodec.cs
@@ -16,6 +16,12 @@ namespace Dekaf.Compression.Snappy;
 /// </summary>
 public sealed class SnappyCompressionCodec : ICompressionCodec
 {
+    // Allocation note (#2214): Snappier's static block helpers allocate a small internal
+    // driver object per call (~48 B compress / ~80 B decompress as of 1.3.1); its reusable
+    // compressor types are internal upstream, and reflection or vendoring is ruled out
+    // (Native AOT + trimming), so instance reuse is blocked on #2352. These are per-block
+    // (i.e. per-batch) costs, not per-message.
+
     // Xerial-snappy magic header
     private static ReadOnlySpan<byte> XerialMagic => [0x82, 0x53, 0x4e, 0x41, 0x50, 0x50, 0x59, 0x00];
 
--- a/tests/Dekaf.Tests.Unit/Compression/SnappyCompressionCodecTests.cs
+++ b/tests/Dekaf.Tests.Unit/Compression/SnappyCompressionCodecTests.cs
@@ -4,6 +4,7 @@ using System.Text;
 using Dekaf.Compression;
 using Dekaf.Compression.Snappy;
 using Dekaf.Protocol.Records;
+using Dekaf.Tests.Unit.Protocol;
 
 namespace Dekaf.Tests.Unit.Compression;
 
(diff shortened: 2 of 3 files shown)


thomhurst:
Review feedback addressed in ed17b4e28.

**greptile P2 — aligned split skips the scratch path: real, fixed.** With an 8192-byte payload in 1024-byte blocks and the segment boundary at 4096, every block sat inside one segment, so `Compress` always took the `blockSequence.IsSingleSegment` branch and the `t_compressedBuffer` scratch this test claims to pin was never touched. Split moved to 4097, so the fifth block (4096–5120) straddles both segments and the multi-segment branch runs; the fresh-instance comparison uses the same split so the output comparison stays like-for-like. The local `CreateMultiSegmentSequence` helper now takes an optional `splitAt` and the reason is recorded in a comment so a future edit doesn't silently re-align it.

**Claude review nit** — dropped the dead pre-loop `ArrayBufferWriter` initialisation (unconditionally reassigned on the first iteration).

`SnappyCompressionCodecTests` 20/20 green; `tests/Dekaf.Tests.Unit` builds 0 warnings with `TreatWarningsAsErrors`.

Issue #2033, opened 2026-07-13
P0 perf: 1-broker producer capped at ~850k (0.61-0.70x Confluent) — every ProduceRequest carries exactly one batch (MaxRequestSize x one-batch-per-partition); admission budget is not the constraint


thomhurst:
## Summary

Stress run [29270211335](https://github.com/thomhurst/Dekaf/actions/runs/29270211335) (commit 072f640d, PR #2019 branch) shows the 1-broker producer gap is **not** an admission-budget problem. The budget/probe fixes from #2019/#2020/#2029 landed and work as designed, yet throughput is still 0.61–0.83x Confluent:

| Scenario (1 broker) | Dekaf | Confluent | Ratio | Budget state |
|---|---|---|---|---|
| Fire-and-forget | 865k msg/s | 1.21M | 0.70x | 30–50 MiB budget, unacked 0–11 MiB, 246 blocks/890s |
| Fire-and-forget (3conn) | 1.0M | 1.21M | 0.83x | similar |
| Acks-all | 781k | 1.27M | **0.61x** | **98–185 MiB budget, unacked ≈ 0, 3 blocks total in 15 min** |
| Idempotent | 847k | 1.21M | 0.69x | similar |

The acks-all row is decisive: a 150+ MiB budget with ~0 unacked bytes and 3 admission blocks in 15 minutes cannot be throttling anything. Dekaf uses only 0.80–0.89 standing cores on a 2-core box (Confluent: 1.68–1.74) with healthy 594–638 KB requests. The constraint is the **wire-request shape and the single-threaded send path**, not admission.

This partially contradicts the #2008 theory (feed-forward budget with no upward force): the #2020 upward force can't fire (growth requires admission blocks) and doesn't need to — on this branch the budget is inert (see also the companion sampler issue for *why* it's inert-oversized).

## Root cause: every ProduceRequest carries exactly one batch

Stress config: linger 5 ms, BatchSize 1 MiB, 6 partitions, MaxRequestSize 1 MiB. At ~865k msg/s the only seal trigger is linger — batches seal at ~600–720 KB. Coalescing then enforces:

1. **One batch per partition per request** — `BrokerSender.cs:2219-2225` (`coalescedPartitions.Add` de-dup).
2. **Request budget ≤ MaxRequestSize** — `CarryOverIfCoalescedLimitReached`, `BrokerSender.cs:2298-2303`. Two 640 KB batches exceed 1 MiB → carry-over.

[...]

Issue #1584, opened 2026-07-08
Producer send loop saturates exactly one core with a single connection (1-broker FnF ceiling vs Confluent)


thomhurst:
## Summary

Stress run [28964691598](https://github.com/thomhurst/Dekaf/actions/runs/28964691598), 1-broker fire-and-forget producer (15 min, 1000B messages, client pinned to 2 cores):

| Client | msg/s (median interval) | Cores used |
|---|---|---|
| Dekaf 1conn | 1,052k | **1.00 (exactly pegged)** |
| Dekaf 3conn | 1,427k | 1.20 |
| Confluent | 1,422k | 1.76 |

Dekaf's per-message CPU cost is far better than Confluent's (0.95 vs 1.62 µs/msg), but with a single connection the send-loop thread saturates exactly one core and total throughput caps at single-core speed. Confluent spreads its work across ~1.76 of the 2 available cores and wins the scenario. With 3 connections Dekaf matches/beats Confluent using ~32% less CPU, confirming the ceiling is thread topology, not per-message efficiency.

Same pattern in 1-broker idempotent: Dekaf 1conn 1,123k vs Confluent 1,195k; Dekaf 3conn 1,549k.

## Open questions

- #1136 / #1139 moved uncompressed batch serialization off the send-loop thread (pre-serialization at append time). Yet the send loop still pegs 1.0 core in this scenario — verify the pre-serialized fast path actually engages for the stress FnF configuration, or determine what work (CRC, buffer copy, socket write syscalls) remains inline on the loop.
- Consider decoupling remaining per-batch work from the socket-write thread so a single connection is not hard-capped at one core.

## Secondary observation

Within the same run, Dekaf 1conn throughput sagged across the 15 minutes (quartile means 1,204k → 1,059k → 1,063k → 1,001k) while cores-used stayed constant at 1.00 — per-message cost slowly rising. Worth checking while instrumenting this path.

## Goal

Dekaf should beat Confluent in every scenario at default settings. 1-broker single-connection FnF/idempotent are the only remaining losses in run 28964691598.

PR #2526, opened 2026-08-05
perf(networking): zero-copy segmented socket write for multi-batch produce requests

thomhurst:
## Summary

Extends the existing single-batch segmented zero-copy write (`TryPreSerializeSingleBatchProduceRequest`) to **multi-batch, multi-partition, multi-topic coalesced produce requests** — the shape that carries essentially all producer traffic under coalescing (accumulator batches are always pre-encoded/pre-compressed at seal, so `CanWriteSegmented` holds for every real batch; verified in review).

Previously every coalesced request paid two full-body copies on the send loop (serialize into rented array, then into the socket) — ~1.7GB/s of avoidable memcpy at rig rates on the core that pegs 1.00 and is the 1-broker throughput ceiling (#1584). This was fix 2 of #2033, shipped then only for N=1.

## Design

- Metadata (header, preamble, topic/partition framing, 61-byte batch headers) serialized into one rented buffer; record bytes referenced in place via pooled `SegmentedFrameSegment` descriptors (null-array = metadata slice, resolved at send time since the metadata buffer can grow); existing 16-segment-windowed vectored send.
- **Single-batch fast path byte-for-byte untouched** — N=1 pays zero extra (descriptor rent can't be avoided in the general path since it must survive the await; keeping both paths verified as right altitude).
- All-or-nothing eligibility; ineligible requests pay only O(batches) field reads before falling back to the copying path.
- Descriptor arrays rented **exactly** via a per-batch segment-count hint cached at seal (`SegmentedRecordsSegmentCountHint`) — covers `BufferMemoryAllocationStrategy.Incremental`'s 16KB-chunked batches without grow churn — and cleared prefix-only on return (keeps the GC-rooting protection, drops the full-length memclear).
- Corruption guards (declared-vs-actual records length, `RequestBodySizeHint` mismatch) inherit the fail-closed posture of both existing paths; both fire pre-lock, pre-socket — blast radius one request.

## Evidence

**Micro-benchmark** (new `ProduceRequestPreSerializeBenchmarks`, 1-topic/4-partition/592KB coalesced request, `[MemoryDiagnoser]`, Default + InProcess toolchains agree):

| | Copying | Segmented | Ratio |
|---|---|---|---|
| Mean | 10,459–10,963 ns | 386–394 ns | **0.04 (~27x)** |
| Allocated | 32 B | 32 B | 1.00 |

(32 B = the pre-existing per-request `RentedBufferWriter` instance both production paths already allocate; per ~1000-message request. Zero new allocations.)

**Byte-identity**: 11 unit tests assert the segmented frame equals the copying serializer's output byte-for-byte — expectations built with the library's own `RequestHeader.Write` + `request.Write`, never hand-coded. Cover v12 name-based + v13 topic-id, compression, two topics, multi-batch-per-partition, 24-chunk incremental sequences (also pins the exact-rent arithmetic), fallback, and an end-to-end loopback socket write. Full suite 6623/6623.

**Local stress A-B (full disclosure)**: 2-min NAT-capped local lanes are structurally unable to show the benefit — the send loop idles at ~0.30 cores locally (vs pegged 1.00 on the rig), while any cost of coupling buffer lifetime to socket drain is exaggerated by the slow socket. Two paired p32 runs read median throughput -5..-7% locally within a ±10%-noisy environment (main's own tail swung p99 106→180ms, max 259→675ms between its two runs). CPU µs/msg overlapped. **The paired A-B-A gate on `producer-1b` (send-loop-bound lane) is the acceptance decider** — dispatching with `baseline_sha` set to the branch base; results will be posted here. Per rule 8: not mergeable before that comparison is posted and PASS.

## Review

/simplify ran with 4 agents; applied: exact descriptor sizing (Incremental-mode fix), prefix-only clearing, metadata buffer pre-sizing, produce-preamble + size-guard dedup between the two pre-serializers, test dedup + a vacuous assertion removed. Deliberately kept: the ~20-line send-loop duplication between the two write overloads (merging would regress N=1 or add an async state machine). Designated follow-up (not this PR): the produce framing now exists in three encoders (protocol `Write`, single-batch, segmented) — hoisting into the protocol layer is the deeper home; byte-identity tests guard the drift meanwhile.

## Risks

- Zero-copy holds batch buffers across the socket write (same property as the shipped single-batch path, now for all coalesced batches). If the A-B-A shows tail-latency cost on slow-drain shapes, that coupling is the suspect.
- `RequestBodySizeHint` mismatch now throws for multi-batch requests (was silently absorbed by the copying serializer's capacity-only use). Hint arithmetic verified exact for BrokerSender's coalesced shape; a throw here surfaces a sender-accounting bug loudly instead of masking it.

### DIFF
--- a/src/Dekaf/Networking/KafkaConnection.cs
+++ b/src/Dekaf/Networking/KafkaConnection.cs
@@ -124,6 +124,14 @@ public sealed partial class KafkaConnection :
     internal const int DefaultPreSerializeInitialCapacity = 4096;
     private const int MessageSizePrefixLength = 4;
     private const int RequestHeaderSizeSlack = 128;
+
+    // Per-entry estimates for sizing the segmented produce metadata buffer so typical
+    // coalesced requests never grow it mid-serialization (RentedBufferWriter growth is the
+    // safety net for outliers such as very long topic names). A topic contributes a UUID or
+    // compact name plus partition-array framing and tagged fields; a partition contributes
+    // its index, records-length varint, and tagged fields.
+    private const int SegmentedTopicMetadataSizeEstimate = 40;
+    private const int SegmentedPartitionMetadataSizeEstimate = 24;
     private static int s_globalCorrelationId;
     private readonly PendingRequestShard[] _pendingRequestShards;
     private readonly SemaphoreSlim _pendingRequestSlots;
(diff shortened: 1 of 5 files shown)


thomhurst:
## A-B-A verdict: REGRESSION — candidate rejected, closing

Run [31026979422](https://github.com/thomhurst/Dekaf/actions/runs/31026979422), lane producer-1b, exact-SHA A-B-A vs `ec0b9bab8`:

| Metric | Baseline A | Candidate | Baseline B | Delta | Control noise | Verdict |
|---|---|---|---|---|---|---|
| Delivered throughput (msg/s) | 1,454,959 | 1,312,093 | 1,340,988 | -6.14% | 8.15% | **regression** |
| Median interval throughput (msg/s) | 1,469,791 | 1,346,402 | 1,336,673 | -4.05% | 9.49% | **regression** |
| Latency p50/p95/p99 (ms) | 7.95/11.75/17.25 | 8.45/13.15/18.85 | 8.85/14.45/20.35 | +0.3–0.6% | 10–21% | inconclusive |
| CPU (µs/msg) | 0.958 | 1.014 | 1.039 | +1.61% | 8.09% | pass |
| Allocation (B/msg) | 1.204 | 1.199 | 1.232 | -1.54% | 2.26% | pass |

**Why the 27x micro-bench didn't translate:** at ~640 msg/request the ~10µs per-request serialize copy amortizes to ~0.016 µs/msg — under 2% of the 0.958 µs/msg total. The double body copy was never a material share of send-loop CPU at this batch shape (#2033's fix-2 premise doesn't hold at current scale). Meanwhile the candidate's delivered throughput sat below *both* baseline segments, consistent in direction with both local paired runs (-5..-7%) — three independent paired observations. Suspected mechanism: zero-copy holds accumulator/pre-compressed buffers across the socket write, coupling buffer recycling to socket drain; the copying path frees them at serialize time.

Per rules 7/8: REGRESSION rejects the candidate (no automatic repeat — that's only for INCONCLUSIVE). Closing. Branch `perf/multibatch-segmented-write` retained: the byte-identity test harness and the per-batch `SegmentedRecordsSegmentCountHint` groundwork are reusable if a future candidate addresses the buffer-lifetime coupling (e.g. hybrid copy-when-drain-slow), which would need a new causal design, not a re-run.

PR #2531, opened 2026-08-05
fix(metrics): count producer sent messages/bytes per delivered batch

### LINKED ISSUE #2529: bug(metrics): dekaf.producer.sent.bytes counter is declared but never incremented — reads 0 while messages tick
Found during the 2026-08-05 producer-1b cpu-profile run ([31033317953](https://github.com/thomhurst/Dekaf/actions/runs/31033317953)): `runtime-counters.csv` shows `dekaf.producer.sent.bytes` = 0 for the whole 15-minute run while `messaging.client.sent.messages` ticks at ~1.47M msg/s.

Code confirms: `DekafMetrics.BytesSent` is declared at `src/Dekaf/Diagnostics/DekafMetrics.cs:21-25` and there is no `BytesSent.Add(...)` call anywhere in `src/` (the only other "sent bytes" hits are the unrelated `ConsumeSentBytes` budget accounting in `KafkaConnection.cs`).

Fix shape: increment alongside `MessagesSent` on the produce completion path with the batch/request byte count (per-batch or per-request, NOT per message, per hot-path rules). Pure observability — no perf surface.

thomhurst:
Fixes #2529 — with a **corrected root cause** vs the issue's original framing (the counter *was* wired, but only on the awaited path, with size fields nothing ever populates).

## The two defects

Profile run [31033317953](https://github.com/thomhurst/Dekaf/actions/runs/31033317953) showed `messaging.client.sent.messages` ticking at ~1,500/s while the lane delivered **1.47M msg/s**, and `dekaf.producer.sent.bytes` flat 0:

1. Both counters were emitted per **awaited operation** (`EmitProduceSuccessMetrics`) — fire-and-forget traffic (the overwhelming majority in FnF workloads) never reached them.
2. The bytes add used `RecordMetadata.KeySize + ValueSize` — `init` fields **no client code ever populates** (verified: zero writers in `src/Dekaf`), so even awaited completions added 0.

## Fix

- `DekafMetrics.RecordBatchDelivered(topic, recordCount, encodedBytes)` called once per batch from `ReadyBatch.CompleteSend` — the single once-guarded choke point both the acked (`BrokerSender.cs:3260`) and fire-and-forget (`:3869`) paths funnel through. Review verified call-site emission would double-count on racing completions (`ProducerBatchLifecycleHammerTests` hammers exactly that); the batch guard is the only free exactly-once point. Split/abandon/failure paths verified not to double-count; failure symmetry already exists (`CompleteFailure` emits per-batch `ProduceErrors` for unobserved records).
- Bytes = `EncodedSize` — the actual post-compression wire size of the record batch (semantics updated in the instrument description and docs; the old "key + value" doc was describing values that were never emitted).
- Awaited path keeps only operation duration + errors; gate renamed `OperationScopedMetricsEnabled`. Always-zero `messaging.message.body.size` tag dropped from send spans. `RecordMetadata.KeySize/ValueSize` documented as not populated (removal would be breaking; the in-memory testing cluster still sets them — noted in the doc).
- `docs/docs/observability.md` aligned: bytes semantics, body-size scoped to consume spans, fire-and-forget coverage noted.

## Performance (per-batch path, CLAUDE.md evidence)

No-listener cost: two `Counter<long>.Enabled` reads per batch (~3ns no-ops). `SparseReadyBatchCompletionBenchmarks` A/B on the same box, in-process toolchain:

| | Mean | Allocated |
|---|---|---|
| main | 194.3 ± 14.7 ns | 0 B |
| branch | 199.1 ± 13.7 ns | 0 B |

Overlapping error bars, zero allocations both sides. Enabled path builds the `TagList` inline (stack struct, allocation-free — /simplify review measured this cheaper than the tag-cache it originally had, which was also an unbounded static-growth vector and was removed).

## Tests

- New `ReadyBatchDeliveryMetricsTests`: per-batch emission (record count + encoded bytes), double-`CompleteSend` no-double-count guard, fire-and-forget shape (no completion sources, baseOffset −1).
- `KafkaProducerMetricsTests` updated: awaited path pins duration-only (re-adding counters there fails the test).
- Full unit suite 6614/6614; solution builds 0 warnings.

Reviewed via /simplify (4 agents); applied: tag-cache removal, shared `AccumulatorTestHelpers.GetTag` reuse, comment/doc dedup, gate rename. Noted follow-ups (not this PR): `CompleteFailure` error-tag shape consistency; `KeySize/ValueSize` deprecation candidate for next major.

### DIFF
--- a/docs/docs/observability.md
+++ b/docs/docs/observability.md
@@ -68,11 +68,11 @@ Both spans set `messaging.system = kafka` plus:
 | `messaging.kafka.message.key` | ✓ (string-convertible keys) | |
 | `messaging.destination.partition.id` | ✓ (on delivery) | ✓ |
 | `messaging.kafka.offset` | ✓ (on delivery) | ✓ |
-| `messaging.message.body.size` | ✓ (on delivery) | ✓ |
+| `messaging.message.body.size` | | ✓ |
 | `messaging.kafka.message.tombstone` | ✓ (null-value messages) | ✓ (tombstone records) |
 | `messaging.consumer.group.name` | | ✓ |
 
-`messaging.message.body.size` is the value payload only — the key is not part of the message body; tombstones report `0`.
+`messaging.message.body.size` is set on consume spans only, and is the value payload only — the key is not part of the message body; tombstones report `0`.
 
 Failures set the span status to `Error`, set `error.type` to the exception's fully-qualified type name, and record an `exception` event with `exception.type`, `exception.message`, and `exception.stacktrace`. Successful spans leave the status unset, per the OTel span-status guidance.
 
@@ -84,7 +84,7 @@ These are the spec-defined instruments from the [OTel messaging metrics conventi
 
 | Instrument | Type | Unit | Description |
 |------------|------|------|-------------|
-| `messaging.client.sent.messages` | Counter | `{message}` | Messages published |
+| `messaging.client.sent.messages` | Counter | `{message}` | Messages published (counted per delivered batch; includes fire-and-forget) |
 | `messaging.client.operation.duration` | Histogram | `s` | Produce operation duration (successes and failures) |
 | `messaging.client.consumed.messages` | Counter | `{message}` | Messages received |
 
@@ -98,7 +98,7 @@ Dekaf-specific instruments live under the `dekaf.*` prefix — the `messaging.*`
 
 | Instrument | Description |
 |------------|-------------|
-| `dekaf.producer.sent.bytes` | Bytes published (key + value) |
+| `dekaf.producer.sent.bytes` | Encoded record-batch bytes published (wire size after compression; counted per delivered batch, includes fire-and-forget) |
 | `dekaf.producer.send.errors` | Produce errors (includes fire-and-forget delivery failures) |
 | `dekaf.producer.send.retries` | Produce retries |
 | `dekaf.producer.buffer.used_bytes` / `limit_bytes` | `BufferMemory` reservation vs configured limit |
--- a/src/Dekaf/Diagnostics/DekafDiagnostics.cs
+++ b/src/Dekaf/Diagnostics/DekafDiagnostics.cs
@@ -84,7 +84,10 @@ public static class DekafDiagnostics
 
     /// <summary>
     /// Tag set for messaging.client.* instruments: the spec-required messaging.system and
-    /// messaging.operation.name plus the destination. Callers cache the result per topic.
+    /// messaging.operation.name plus the destination. <see cref="TagList"/> is a stack struct
+    /// holding reference-typed tag values, so building one costs no allocation; per-batch call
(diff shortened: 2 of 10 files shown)


thomhurst:
Review feedback addressed in 583d68653.

**greptile P1 — telemetry interrupts batch completion: real, fixed.** Confirmed the mechanism: `Counter.Add` invokes `MeterListener` callbacks (user/exporter code) inline on the caller's thread, and `RecordBatchDelivered` runs after `CompleteSend` has claimed `_sendCompleted` but before the completion sources are signalled — a throwing listener retires the batch with its awaiters unsignalled, and every later `Fail` short-circuits on the same guard, so `ProduceAsync` callers wait forever.

Emission now sits behind a `catch` in a cold callee (`EmitBatchDelivered`), which keeps the no-listener fast path — two `Enabled` reads — inlinable, since RyuJIT never inlines a method containing an exception handler.

Two sibling emissions were already exposed to the same hazard on `main`, so they got the same isolation rather than being left as a known trap:
- `CompleteFailure`'s unobserved-error counter — identical strand (guard claimed, completion sources not yet faulted).
- The KIP-126 split counter — escapes between queueing the child batches and `SignalWakeup()`.

Mutation-verified: with the `catch` removed, `CompleteSend_ThrowingMetricsListener_StillCompletesAwaiters` and `Fail_ThrowingMetricsListener_StillFaultsAwaiters` both fail with the listener's exception propagating out of `ReadyBatch`; with it restored, both pass.

**Claude review items:**
1. *End-to-end coverage gap* — added `SendLoop_AckedBatch_EmitsSentCountersThroughRealSendPath`: enqueue → serialize → scripted ack → per-batch emission through the real `BrokerSender` loop, asserting record count and non-zero encoded bytes. This is the coverage that matches the root cause (counters wired onto a path production traffic never takes); a direct `CompleteSend` call can't catch that class.
2. *Duplicate `GetTag`* — `KafkaProducerMetricsTests` now delegates to `AccumulatorTestHelpers.GetTag`.
3. *Third `MeterListener` bootstrap* — factored into `AccumulatorTestHelpers.StartMeterListener(names, onLong, onDouble)`; the three copies in the files this PR touches now use it.
4. *Stale `ClientMetricTags` doc* — updated: `TagList` is a stack struct with reference-typed values, so per-batch call sites deliberately build it inline instead of caching per topic (measured cheaper, and avoids the unbounded per-topic cache `/simplify` removed).

**Performance** — `SparseReadyBatchCompletionBenchmarks`, same box, in-process toolchain, no listener attached (production-representative):

| | Mean | Allocated |
|---|---|---|
| pre-review branch state | 193.6 ± 6.10 ns | 0 B |
| 583d68653 | 195.8 ± 4.76 ns | 0 B |

Overlapping error bars, zero allocations both sides.

Unit suite 6617/6617; solution builds 0 warnings with `TreatWarningsAsErrors`.

PR #2670, opened 2026-08-16
fix(producer): prevent late batch publication after disposal

### LINKED ISSUE #2567: fix(producer): eliminate orphaned in-flight batch ownership paths
## Problem

The producer has a periodic orphan sweep because an in-flight batch can lose its normal ownership/reference path. Current source comments state that the root cause remains under investigation and that, without the sweep/delivery timeout, affected `ProduceAsync` calls could hang indefinitely.

The sweep is a valuable defense, but it is not a substitute for a proven ownership invariant. A batch accepted by the producer must always be owned by exactly one lifecycle stage until it completes or fails.

This is the same externally visible failure class as [Confluent.Kafka #1057](https://github.com/confluentinc/confluent-kafka-dotnet/issues/1057): produce work never completes.

## Scope

Find and eliminate the remaining loss-of-ownership path while retaining the sweep as defense in depth until stress evidence proves it is no longer exercised.

Instrument per-batch lifecycle transitions at batch/cold-path granularity so a swept batch reports its last owner/state without adding per-message allocation or logging.

Audit at least:

- accumulator drain and broker routing;
- request construction and pending-response registration;
- retry/reroute and leader changes;
- connection EOF, timeout, cancellation, and disposal;
- incarnation/generation guards;
- completion, interceptor callback, resource cleanup, and `BufferMemory` refund.

## Acceptance criteria

- Every accepted batch is owned by exactly one tracked stage until one terminal completion path wins.
- Delivery completion, callbacks, pooled-resource returns, and memory refund each occur exactly once.
- A deterministic regression/fault-injection test reproduces the causal race or proves the corrected transition that previously lost ownership.
- Tests cover connection EOF/stall, timeout, retry, reroute, leader change, response-processing exception, and concurrent disposal.
- Normal runs never depend on the periodic sweep for completion; a test/metric asserts zero swept batches under sustained fault-injection stress.
- Keep the sweep as defense in depth initially, with actionable diagnostics for any future hit.
- No O(n) hot-path scan, per-message lock, allocation, or thread-pool hop is introduced.
- Producer `[MemoryDiagnoser]` remains `0 B` per message.
- Before/after throughput, p50/p99/max latency, CPU, allocations, and stability are recorded and Pareto-safe.

## Related

- #2199 / PR #2200 fixed one `ArrayPool` double-return/refund-loss family and added diagnostics.
- This issue tracks the still-documented orphan-batch ownership gap, not the already-fixed refund bug.

thomhurst:
## Summary
- close the disposal/final-drain TOCTOU by serializing completed-batch publication with the partition deque lock
- fail never-published batches with ObjectDisposedException while refunding broker budget, BufferMemory, completion sources, and pooled resources exactly once
- clear unstarted compression work before failure so Gzip disposal cannot block
- replace the concurrent integration test's arbitrary delay with a deterministic accepted-produce handoff

Fixes #2567.

## Causal proof

The new unit fault-injection barrier pauses a detached batch after completion but before deque publication. On the old code, DisposeAsync completed its final drain, the append then published into an ownerless deque, and its completion timed out. With this change, both uncompressed and Gzip variants fail the accepted completion with ObjectDisposedException and finish with zero in-flight batches and zero buffered bytes.

## Verification
- Release solution build: 0 warnings, 0 errors
- unit suite: 6,725 passed
- focused ownership/fault suites: 229 passed
- ErrorPropagationTests integration class (Kafka 4.3.1): 11 passed
- deterministic race: 2/2 passed (None and Gzip)

## Local performance gate

Identical AccumulatorAppendBenchmarks.AppendHotPath runs on commit ce816b17 and this candidate, .NET 10.0.11, 3 warmups / 3 iterations:

| Run | 100 B | 1000 B | Allocated |
|---|---:|---:|---:|
| baseline before | 70.73 ns | 146.72 ns | 0 B |
| candidate final | 70.33 ns | 150.93 ns | known drainer-behind noise at 100 B; 0 B at 1000 B |
| baseline after | 71.68 ns | 155.27 ns | 0 B |

Candidate versus adjacent baseline-after control: -1.88% at 100 B and -2.79% at 1000 B. The 100 B allocation sample hit the benchmark's explicitly documented stochastic cold-path drainer-behind exception; the earlier candidate run and both controls measured 0 B. No stable throughput or allocation regression was observed.

The exact-SHA producer stress lane will be recorded before merge for throughput, p50/p99/max latency, CPU, allocations, and stability.

### DIFF
--- a/src/Dekaf/Producer/RecordAccumulator.cs
+++ b/src/Dekaf/Producer/RecordAccumulator.cs
@@ -1279,6 +1279,7 @@ public sealed partial class RecordAccumulator : IAsyncDisposable
     internal Action? PurgeAppendWaitObservedForTest;
     internal Action? AfterLingerQueueSnapshotForTest;
     internal Action<TopicPartition>? AfterFlushPartitionVisitedForTest;
+    internal Action? BeforeCompletedBatchEnqueueForTest;
 
     /// <summary>
     /// True after CloseAsync has been called. Used by the sender loop to know
@@ -4816,6 +4817,7 @@ public sealed partial class RecordAccumulator : IAsyncDisposable
     private ReadyBatch? CompleteDetachedBatchAndEnqueue(PartitionDeque pd, PartitionBatch batchToComplete)
     {
         ReadyBatch? readyBatch;
+        ReadyBatch? rejectedBatch = null;
         var bytesToRelease = 0;
         try
         {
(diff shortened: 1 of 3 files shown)


thomhurst:
Exact-SHA producer acceptance dispatched: https://github.com/thomhurst/Dekaf/actions/runs/31944276179

Configuration: producer-1b, 15 minutes, 1000-byte messages, profile off, exact baseline ed18ce4fdbfa5304275ecc9971034f031a3734c4. The workflow will run baseline → candidate 04bf65d93de6e0eadcc7e48cb72473d18c19dce1 → baseline and issue the formal Pareto verdict.

PR #2719, opened 2026-08-17
test(producer): make crash sequencing deterministic

### LINKED ISSUE #2716: test: make unexpected-exit sequence assertion deterministic
## Problem

`BrokerSenderSendLoopTests.SendLoop_UnexpectedExit_RedeliversPendingIdempotentBatches` fails intermittently because it assumes both pending batches have received non-negative idempotent base sequences before the injected wave-coalescing crash.

Observed on a full net10.0 unit run at unrelated Avro PR #2699 head `6938c8e2` plus Avro-only working changes:

```text
Expected reroutedSequences.Count(sequence => sequence >= 0) to be 2
but found 1
BrokerSenderSendLoopTests.cs:776
```

The test waits only for `GetPendingResponseCount(sender) == 1` before enqueuing the crash-triggering batch. That proves one request is pending, but does not deterministically prove both `pendingFirst` and `pendingSecond` have been assigned sequences. All three batches were rerouted (`reroutedCount == 3`), while only one pending batch had a non-negative sequence.

The working diff touched only Avro serializer, Avro tests/AOT, and Avro benchmarks; no producer files changed. Per repository policy, the failed suite was not rerun.

## Expected fix

Add deterministic synchronization proving both pending idempotent batches have been sequenced/in-flight before enabling the injected crash, or assert the actual invariant without depending on scheduling. Do not use delays or blind retries.

thomhurst:
## Summary

- gate the sender at its existing wave-coalescing test hook
- enqueue both pending idempotent batches before releasing the first send
- assert both batches have assigned base sequences before injecting the unexpected exit
- release the test gate on every cleanup path

This removes the scheduling assumption that one pending response necessarily contained both bulk-enqueued batches.

## Test plan

- [x] focused test: 50/50 passes on net10.0
- [x] focused test: 50/50 passes on net8.0
- [x] full unit suite net10.0: 6,983 passed
- [x] full unit suite net8.0: 6,847 passed
- [x] Release builds: 0 warnings/errors
- [x] `/simplify`, self-review, and `git diff --check`

No `src/` change; no benchmark required.

Closes #2716

### DIFF
--- a/tests/Dekaf.Tests.Unit/Producer/BrokerSenderSendLoopTests.cs
+++ b/tests/Dekaf.Tests.Unit/Producer/BrokerSenderSendLoopTests.cs
@@ -739,6 +739,8 @@ public sealed class BrokerSenderSendLoopTests : ScriptedProduceResponseFixture
         var valueTaskSourcePool = new ValueTaskSourcePool<RecordMetadata>();
         var injectedFailure = new InvalidOperationException("injected pending-response failure");
         var allRerouted = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
+        var coalesceStarted = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
+        using var releaseCoalesce = new ManualResetEventSlim();
         var reroutedSequences = new ConcurrentQueue<int>();
         var reroutedCount = 0;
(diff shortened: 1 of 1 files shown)


thomhurst:
@codex review

@coderabbitai review

PR #2727, opened 2026-08-17
test(producer): make retry deadline deterministic

### LINKED ISSUE #2726: test(producer): make TV2 retry deadline deterministic under load
## Problem

`TransactionOffsetCommitTests.TV2_TransportFailuresUntilDeadline_ArePreservedAsTimeoutCause` is timing-dependent under scheduler/ThreadPool pressure.

Observed on commit `c63a6d14` while two full unit suites and Docker integration ran concurrently:

- configured `maxBlockMs`: 2,000 ms
- assertion ceiling: 2,800 ms
- observed elapsed: 3,165.9 ms

The same code passed in an isolated CI-like run. This was investigated rather than blindly rerun.

## Cause

`KafkaProducer.CreateTransactionRetryCancellationSource` computes a monotonic deadline, then enforces it through `CancellationTokenSource.CancelAfter`. The cancellation callback is timer/ThreadPool scheduled. Under heavy local oversubscription it can run late, so the test's wall-clock assertion depends on scheduler timing rather than deterministic synchronization.

The production retry loops also check the monotonic `TransactionRetryBudget`, but this test's injected 1,200 ms coordinator delay waits on the delayed cancellation callback.

## Required work

- Make the test deterministic under scheduler load, preferably through an injectable/fake time/deadline mechanism or deterministic coordination rather than a looser wall-clock threshold.
- Preserve the production guarantee that retry timeout exceptions retain the last transport exception.
- Audit the other transaction retry paths using `CreateTransactionRetryCancellationSource` for the same testing gap.
- Do not solve by blindly rerunning, increasing the threshold, or adding arbitrary delays.
- Keep production hot paths allocation/Pareto safe; benchmark any `src/` change against its exact parent with `[MemoryDiagnoser]`.

## Reproduction

Run the net8 and net10 full unit executables plus the Schema Registry Docker integration class concurrently on a loaded Windows host. The strict elapsed assertion can exceed 2.8 seconds even though isolated runs pass.

thomhurst:
## Summary
- replace the wall-clock `Stopwatch` assertion with the existing injected transaction clock
- advance the fake deadline when the first transport failure is thrown
- preserve assertions for transaction timeout classification and the exact retained `IOException`
- remove the 1,200 ms real delay, 1,000-exception queue, and scheduler-sensitive 2.8 second threshold

Closes #2726

## Validation
- Release unit project build: 0 warnings, 0 errors
- focused `TransactionOffsetCommitTests`: 15/15 on net10.0; 15/15 on net8.0
- full unit: 7,204/7,204 on net10.0; 7,068/7,068 on net8.0
- `git diff --check`: clean
- `/simplify` complete; immediate self-review clean

## Performance
Test-only change; no `src/` path, runtime behavior, allocation, or benchmark delta.

### DIFF
--- a/tests/Dekaf.Tests.Unit/Producer/TransactionOffsetCommitTests.cs
+++ b/tests/Dekaf.Tests.Unit/Producer/TransactionOffsetCommitTests.cs
@@ -1,4 +1,3 @@
-using System.Diagnostics;
 using System.Reflection;
 using Dekaf.Consumer;
 using Dekaf.Errors;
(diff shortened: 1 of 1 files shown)


thomhurst:
@claude review exact head 9b30a46e`n@coderabbitai review exact head 9b30a46e

thomhurst:
Addressed Claude's two non-blocking observations in `384ceec0`:

- removed the inert `retryBackoffMs` setting
- preserved plural retry-loop coverage with two queued `IOException`s
- each failure advances the fake monotonic clock by exactly half the 2,000 ms budget
- asserted exactly two commit requests before the timeout preserves the last transport cause

No real delay or wall-clock threshold was reintroduced. Validation: focused class `15/15` on both TFMs; full unit net10 `7204/7204`; net8 `7068/7068`; build 0 warnings/errors; `git diff --check` clean. `/simplify` and immediate self-review clean.

@claude review exact head 384ceec0
@coderabbitai review exact head 384ceec0

Issue #2846, opened 2026-08-21
test: make pause snapshot publication race deterministic


thomhurst:
## Problem

`ConsumerPauseResumeCacheTests.PauseSnapshot_CapturesMutationAfterWaitingForPublication` failed during a concurrent net8.0/net10.0 full-unit validation on current `main` (`6a543194c`). The net8.0 test spent 28.186 seconds and failed `Assert.That(publisherMutated).IsTrue()` at line 135; the paired net10.0 suite passed all 7,773 tests.

This was discovered while validating unrelated test-only producer work for #2843. No consumer source or consumer test files differed from `main`.

## Likely cause

The test uses `SpinWait.SpinUntil(..., TimeSpan.FromSeconds(5))` to observe the publisher thread's `_paused` mutation while the test thread holds `BatchIterationEpoch` publication. Under concurrent full-suite CPU pressure, the publisher did not reach the mutation before the wall-clock timeout. The later assertions and cleanup then extended total duration to 28 seconds.

## Acceptance criteria

- Replace the wall-clock-sensitive thread scheduling assumption with deterministic coordination around the publisher mutation/publication boundary.
- Preserve coverage that `PublishPausedSnapshot` waits for the active publication and captures mutations made before it acquires the epoch.
- Prove net8.0 and net10.0 full suites pass concurrently without increasing timeouts or rerunning a failed job.

PR #2847, opened 2026-08-21
test(producer): stabilize adaptive scale-down

### LINKED ISSUE #2843: test: eliminate AdaptiveScaleDown full-suite timing flake
## Problem

`ScaleDown_UnderContinuousIdempotentTraffic_LosesNoBatches` failed during a full net10 unit run on current `main` (`b5d2d56e6`) with exactly one failed acknowledgement after 30.125 seconds. The test's configured delivery timeout is 30 seconds.

The same binary then passed:

- focused test: 1/1 in 612 ms
- all `Dekaf.Tests.Unit.Producer` tests: 1205/1205 in 9.19 s
- instrumented full suite: 7671/7671 in 21.01 s

No source in the failing producer path differed from `main`; discovery happened while validating unrelated admin-only work.

## Likely cause

Full-suite load exposes timing sensitivity or starvation around the scale-up/scale-down synchronization. The failure duration matching `deliveryTimeoutMs: 30_000` indicates a batch remained unacknowledged until delivery expiry. The test currently uses `Task.Delay(5)`, `Task.Delay(1)`, and a `Task.Yield()` polling loop, and does not report the first acknowledgement exception.

## Acceptance criteria

- Reproduce or instrument the full-suite-only failed acknowledgement so the exact exception/path is reported.
- Replace timing-dependent synchronization with deterministic signals where possible.
- Prove the test passes repeatedly under full-suite parallel load without increasing the delivery timeout or blindly retrying.
- Preserve the zero-loss scale-down regression coverage.

thomhurst:
## Summary

- replace delay/yield polling with acknowledgement-channel synchronization
- make scale cooldown and sustained-window gates deterministic for this zero-loss regression
- surface the first failed acknowledgement with partition and inner exception context
- bound bridge traffic while preserving scale-up, scale-down apply, and post-shrink coverage

## Validation

- `dotnet build tests/Dekaf.Tests.Unit --configuration Release --no-restore` — 0 warnings, 0 errors
- `AdaptiveScaleDownTests` — 39/39 on net8.0 and 39/39 on net10.0
- concurrent full-suite round — net8.0 7,637/7,637 and net10.0 7,773/7,773
- second concurrent full-suite round — target test passed on both runtimes; net10.0 7,773/7,773, while net8.0 exposed unrelated `ConsumerPauseResumeCacheTests` timing defect tracked as #2846 (not rerun)

No `src/` files changed; performance benchmarks are not applicable.

Fixes #2843

## Summary by CodeRabbit

* **Tests**
  * Improved continuous traffic testing for adaptive scale-down scenarios.
  * Added validation that acknowledgements remain successful and deliveries continue after scale-down.
  * Enhanced test reliability by coordinating traffic waves and acknowledgement handling more precisely.

### DIFF
--- a/tests/Dekaf.Tests.Unit/Producer/AdaptiveScaleDownTests.cs
+++ b/tests/Dekaf.Tests.Unit/Producer/AdaptiveScaleDownTests.cs
@@ -1,6 +1,7 @@
 using System.Buffers;
 using System.Collections.Concurrent;
 using System.Reflection;
+using System.Threading.Channels;
 using Dekaf.Compression;
 using Dekaf.Errors;
 using Dekaf.Metadata;
@@ -671,7 +672,7 @@ public sealed class AdaptiveScaleDownTests
     [Timeout(120_000)]
     public async Task ScaleDown_UnderContinuousIdempotentTraffic_LosesNoBatches(CancellationToken cancellationToken)
     {
-        var options = CreateOptions(idempotent: true, scaleCooldownMs: 25, scaleDownSustainedMs: 50);
+        var options = CreateOptions(idempotent: true, scaleCooldownMs: 0, scaleDownSustainedMs: 0);
         var accumulator = new RecordAccumulator(options);
         var vtPool = new ValueTaskSourcePool<RecordMetadata>();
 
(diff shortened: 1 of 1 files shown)


thomhurst:
@claude review

thomhurst:
Addressed both review findings in c5bd3b129:

- acknowledgement channel now declares `SingleWriter = false`, matching callback paths that can write from the send loop and enqueue-failure callers;
- captures/asserts the scale-up target exceeds the initial connection count;
- enqueues and awaits a successful post-shrink probe before Wave 2. Because `MaybeScaleConnections` runs before coalescing, that acknowledgement proves `ApplyScaleDown` completed and the send loop survived.

Validation:
- Release build: 0 warnings, 0 errors
- focused regression: net10 1/1, net8 1/1
- full unit: net10 7,773/7,773; net8 7,637/7,637
- post-simplify focused net10: 1/1
- `git diff --check`: clean
- explicit pre-push self-review: no findings

PR #2976, opened 2026-09-01
Revert Reservoir 1.5.0 upgrade (perf regression)

thomhurst:
Reverts #2975 (353086ffe), restoring Reservoir 1.4.0.

## Why

Reservoir 1.5.0 regressed Dekaf: **throughput −8.13%, CPU +11.17%** with defaults; **−3.58% / +7.45%** with the wrapper-exposed `threadLocalFastPath` flags off. The auto-merge landed before benchmarks finished, so main carries the regressed configuration.

Root causes (full investigation in thomhurst/Reservoir#84):
- 1.5.0 enables Reservoir's thread-local fast path by default. Dekaf's pools are cross-thread handoff shaped (rent on the producer thread, return on the IO/completion thread) — the fast path's adversarial workload: returns park objects on threads that never rent them.
- The "disabled everywhere" configuration was incomplete: six direct 2-arg `Reservoir.ObjectPool` constructions silently kept the fast path on — `PendingRequestPool`'s inner pool, static `PooledPipelinedResponse.Pool`, `PartitionInflightTracker`, `KafkaConsumer.PendingFetchData`, static `RentedBufferWriter.Pool`, and `PipeMemoryPool`'s owner pool. That explains the residual regression.

## Path forward

The next Reservoir release flips the fast-path default back off (thomhurst/Reservoir#82, merged) and gates thread-local parking to renting threads (thomhurst/Reservoir#84), so re-upgrading will need no Dekaf changes and heals all six hidden pools automatically.

https://claude.ai/code/session_01GF8xT4pZXckH3iCGoTPbg6

## Bug Fixes
- Improved bounded pooling behavior so pools no longer retain items beyond their configured capacity.
- Excess returned items are now discarded and reset correctly.
- Updated connection, batching, and asynchronous pooling behavior for more consistent resource reuse.
- Corrected pool-count handling when pools start empty or operate under concurrent load.

## Chores
- Downgraded the centrally managed Reservoir package version to 1.4.0.
- Updated internal documentation and validation coverage to reflect bounded pool behavior.

PR #2977, opened 2026-09-01
perf(pooling): mark non-throwing reset policies for Reservoir's inline reset path

thomhurst:
**Draft — sequenced after the Reservoir re-upgrade.** Requires Reservoir ≥ 1.5 (`INonThrowingResetPolicy` doesn't exist in 1.4), so this conflicts with the interim revert #2976 by design: merge order is #2976 now → re-upgrade to the next Reservoir release (fast-path default off + thomhurst/Reservoir#84) → this PR.

## What

Reservoir resets pooled objects through a `NoInlining` destroy-on-throw wrapper unless the policy implements the `INonThrowingResetPolicy` marker, in which case the reset inlines (the type test folds to a constant per instantiation). Reservoir's benchmark shows ~0.6 ns/op off a rent/return pair, and the marked variant is also the code-layout-stable one across builds.

Marks the nine policies whose `TryReset` provably cannot throw — trivial `=> true` bodies (wrapper `PoolPolicy`, `PendingFetchDataPolicy`, `PooledMemoryOwnerPolicy`, `SerializationCachePolicy`, `ValueTaskSourcePool.PoolPolicy`, `CompletionSourceArrayPolicy`), pure field-assignment resets (`RentedBufferWriter.PoolPolicy`, `InflightEntryPolicy`), and `PooledPipelinedResponse.PoolPolicy` (field stores + `ManualResetValueTaskSourceCore.Reset` + bit packing).

**Deliberately not marked**: `PendingRequestPool`'s policy — its reset disposes a `CancellationTokenRegistration` and a response-memory reservation; the marker removes the guard that destroys the object when reset throws, so it stays on the guarded path.

Compile-checked against Reservoir 1.5.0.

https://claude.ai/code/session_01GF8xT4pZXckH3iCGoTPbg6

### DIFF
--- a/src/Dekaf/Consumer/KafkaConsumer.cs
+++ b/src/Dekaf/Consumer/KafkaConsumer.cs
@@ -937,7 +937,8 @@ internal sealed class PendingFetchData : IDisposable
     }
 
     private readonly struct PendingFetchDataPolicy(PendingFetchDataPoolState state)
-        : Reservoir.IPooledObjectDestroyPolicy<PendingFetchData>
+        : Reservoir.IPooledObjectDestroyPolicy<PendingFetchData>,
+          Reservoir.INonThrowingResetPolicy
     {
         public PendingFetchData Create() => new();
 
--- a/src/Dekaf/Networking/KafkaConnection.cs
+++ b/src/Dekaf/Networking/KafkaConnection.cs
@@ -1479,7 +1479,8 @@ public sealed partial class KafkaConnection :
 
         private readonly struct PoolPolicy
             : Reservoir.IPooledObjectDestroyPolicy<
-                PooledPipelinedResponse<TRequest, TResponse>>
+                PooledPipelinedResponse<TRequest, TResponse>>,
+              Reservoir.INonThrowingResetPolicy
         {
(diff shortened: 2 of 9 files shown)


thomhurst:
CI note: the `Build & Unit Test (net10.0)` failure (`FailAllPattern_RequestReturnedToPool_WhenAwaiterFinallyCleans`, failed twice) is a Reservoir 1.5.0 semantics flake, not this change — the marked policies don't touch `PendingRequestPool`.

Mechanism: under 1.5.0, `PendingRequestPool`'s inner pool is one of the six 2-arg-ctor pools with the thread-local fast path silently ON, and a fast-path return parks unconditionally in the returning thread's slot. The test returns the request from an awaiter `finally`; when CI's 16-way-parallel suite schedules that continuation onto a different thread-pool thread, the request parks there and the test thread's `Rent()` creates a fresh instance, failing `IsSameReferenceAs`. Locally (single test, continuation runs inline) it passes 6/6 on both this branch and main.

Resolves itself in the intended merge order: #2976 (revert) or the next Reservoir release (fast-path default off + renter-gated parking, thomhurst/Reservoir#84) makes the return go to the shared tier deterministically.

PR #3013, opened 2026-09-05
perf(producer): govern queueing delay against the target minus configured linger

thomhurst:
## Summary

The delivery-latency target is documented as append-to-ack (`WithDeliveryLatencyTarget` doc, `producer-options.md`), but `BrokerWindowController` compared its governed delay (seal-to-send + RTT excess + admission wait; configured linger deliberately excluded by `GovernedOriginTicks`) against the **full** target. With the stress configuration (10 ms target, 5 ms linger) the controller read 7-11 ms as "on target" while callers observed target-plus-linger, so the delay-biased descent never engaged and three-broker windows parked at the eight-quantum floor.

This PR deducts the configured linger from the governed target (never below half the target). `BrokerUnackedByteBudget` already received `lingerSeconds` and discarded it. Default configuration (linger 0) is unchanged.

## Evidence from weekly run 33948814595 (main `0795104e6`)

Per-broker steady state from `results[].producerDeliveryDiagnostics.brokerBudgetSamples` (minutes 3-15):

| Lane | Broker | Window | Drain | Queue = window/drain | Governed EWMA | Target |
|---|---|---:|---:|---:|---:|---:|
| producer-idempotent-3b | b1 / b2 / b3 | 8 / 8 / 5 MB | 458 / 450 / 475 MB/s | 18.3 / 18.6 / 11.0 ms | 8.7 / 8.9 / 8.7 ms | 10 ms |
| producer-acks-all-3b | b1 / b2 / b3 | 8 / 8 / 6 MB | 388 / 390 / 399 MB/s | 21.6 / 21.5 / 15.8 ms | 10.7 / 9.3 / 15.8 ms | 10 ms |
| producer-1b | b1 | 12 MB | 1,721 MB/s | 7.3 ms | 5.4 ms | 10 ms |

Delivered latency (harness, append to ack): idempotent-3b p50 12.6 / p95 30.4 ms, acks-all-3b p50 13.6 / p95 31.4 ms versus Confluent 6.0 / 13.7 and 6.6 / 25.2 ms. Gate breaches: p95/target 3.04 and 3.13, p50/Confluent 2.11 and 2.07. Every 1-broker lane passes.

[...]

PR #3018, opened 2026-09-05
perf(producer): lower the normal window-probe floor from eight to four quanta

thomhurst:
## Summary

Stacked on #3013 (base branch is `perf/governed-target-excludes-linger`; the diff here is only the floor change).

With #3013 the three-broker windows descend symmetrically, but they still park at the eight-quantum normal probe floor: in run 33964050486 the candidate settled at 4 / 8 / 8 MB with governed EWMA 6.1 / 5.6 / 5.7 ms against a 5 ms budget. Eight 1 MB quanta against a ~450 MB/s broker are ~18 ms of standing queue, and the few hundred microseconds of over-target excess are too small for the sub-floor delay-gain test (>= 10%) to pass, so the descent stops there. Delivered p50 was 9.15 ms versus Confluent's 6.0 ms on the same lane in the weekly run.

The eight-quantum floor encoded a one-broker knee of about seven full requests. This PR lowers the normal floor to four quanta (two requests on the wire while two seal) and moves the accelerated 25% treatment threshold with it (six quanta) so a coarse step never lands below the floor. Deep descent below four quanta keeps the existing delay-earning rule; every step above it keeps the paired goodput guard, which is what stops descent at whatever knee a topology actually has.

## Expected effect

- 3 brokers: windows may reach 4 MB (~9 ms queue) where goodput holds; p50 should move toward 7 ms, p95/p99 amplification of broker stalls shrinks with the queue.
- 1 broker: #3013's candidate settled at 10-12 MB with the floor not binding; if the knee really is ~7 requests, down-probes to 6-7 quanta will fail their goodput guard and the window stays. The A-B-A decides.

## Tests

`BrokerUnackedByteBudgetTests` updated for the new floor (cap-clamp case now uses a 5-quantum cap below the 6-quantum threshold; the accelerated-descent case asserts 9 -> 6 -> 4 -> 3; the flat-goodput recovery test now asserts recovery after touching the floor instead of the sawtooth's phase at a fixed epoch). Full unit suite green locally.

## Stress validation

- `producer-idempotent-3b` A-B-A against #3013's head `60cd643a3` (exact parent): [run 33980288545](https://github.com/thomhurst/Dekaf/actions/runs/33980288545). Requires PASS.
- `producer-1b` A-B-A follows a PASS.

No hot-path code changes: two constants and comments.

https://claude.ai/code/session_012AEx3PnpfHTncWCKXsCH9U

### DIFF
--- a/src/Dekaf/Producer/BrokerWindowController.cs
+++ b/src/Dekaf/Producer/BrokerWindowController.cs
@@ -61,14 +61,19 @@ internal sealed class BrokerWindowController
 
     private const int SizeBucketCount = 16;
     private const int SizeBucketShift = 8;
-    // The control run sustained its knee at roughly seven full requests. One extra quantum
-    // absorbs response/scheduler granularity without allowing fragmented requests to redefine
-    // the floor. Descent below this floor is allowed only while the controlled-delay EWMA
-    // exceeds the delivery-latency target (see LatencyDescentPipelineQuanta).
-    private const int MinimumPipelineQuanta = 8;
-    // floor(11 * 0.75) is the first accelerated candidate that remains at or above the
-    // normal eight-quantum floor. Smaller windows use the normal step for deep descent.
-    private const int MinimumAcceleratedDescentQuanta = 11;
+    // Normal down-probes (goodput-guarded only) may reach this floor; descent below it is
+    // allowed only while the governed delay exceeds the target AND each step demonstrably
+    // buys latency (see LatencyDescentPipelineQuanta). The floor used to be eight quanta,
+    // encoding a one-broker knee of ~seven full requests. Against brokers that drain ~450 MB/s
+    // (three-broker topologies on shared cores) eight 1 MB quanta are ~18 ms of standing
+    // queue, and PR #3013's paired runs showed windows parking exactly there with the
+    // governed delay a few hundred microseconds over the budget: too little excess for the
+    // sub-floor delay-gain test, too much for the caller. Four quanta keep two requests on the
+    // wire while two seal, and the paired goodput guard still stops descent at any knee.
+    private const int MinimumPipelineQuanta = 4;
+    // floor(6 * 0.75) is the first accelerated candidate that remains at or above the
+    // normal four-quantum floor. Smaller windows use the normal step for deep descent.
+    private const int MinimumAcceleratedDescentQuanta = 6;
     // Deep-descent floor while over the latency target: two quanta keep the acknowledgement
     // clock pipelined (one request on the wire while the next seals). The eight-quantum
     // floor encoded a one-broker knee and forced ~3x the delivery-latency target of standing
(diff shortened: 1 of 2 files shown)


thomhurst:
## Rejected: exact-parent A-B-A is a REGRESSION on latency

[Run 33980288545](https://github.com/thomhurst/Dekaf/actions/runs/33980288545), `producer-idempotent-3b`, baseline = #3013 head `60cd643a3` -> candidate `bf01aa24c` -> baseline, cheap shape, 15 min per segment, one ubicloud-standard-8 VM.

| Metric | Baseline A | Candidate B | Baseline A2 | B vs control mean | Control drift | Gate |
|---|---:|---:|---:|---:|---:|---|
| Delivered throughput (msg/s) | 1,113,785 | 1,132,600 | 1,149,310 | +0.09% | 3.14% | pass |
| Latency p50 (ms) | 8.650 | 9.250 | 8.450 | **+8.19%** | 2.34% | regression |
| Latency p95 (ms) | 23.050 | 36.450 | 28.450 | **+41.55%** | 20.97% | regression |
| Latency p99 (ms) | 51.950 | 63.150 | 57.850 | **+15.03%** | 10.75% | regression |
| CPU (us/msg) | 0.867 | 0.830 | 0.791 | +0.14% | 9.14% | pass |
| Allocation (B/msg) | 0.721 | 0.611 | 0.632 | -9.59% | 13.23% | pass |
| Steady/peak ratio | 0.973 | 0.972 | 0.947 | +1.23% | 2.75% | pass |

Why it lost, from the per-broker window timelines (`brokerBudgetSamples`, MB by minute):

| Segment | b1 | b2 | b3 | Steady windows | p50 / p95 |
|---|---|---|---|---|---|
| Baseline A (#3013) | 16 7 6 5 4 4 4 ... | 16 8 7 7 8 7 7 ... 6 | 16 7 6 5 4 4 4 ... | 4 / 6 / 4 MB | 8.65 / 23.1 |
| Candidate (4q floor) | 16 **14 14 14 14** 7 6 4 4 ... | 16 14 14 14 14 12 7 5 5 ... | 16 14 14 12 12 7 6 4 4 ... | 4 / 5 / 4 MB | 9.25 / 36.5 |
| Baseline A2 (#3013) | 16 14 12 7 7 7 7 7 8 8 ... | 16 14 10 6 5 4 4 4 ... | 16 14 10 7 7 6 6 6 5 4 ... | 8 / 4 / 5 MB | 8.45 / 28.5 |

The premise was wrong: with #3013 in place the deep-descent rule already takes over-target brokers below the eight-quantum floor (both baseline segments settled at 4-6 MB on two of three brokers), so the floor was not what parked them at 8 MB in the earlier runs. Lowering the normal floor bought nothing at steady state (same 4-5 MB windows) and made convergence slower (the candidate held 14 MB on every broker for minutes 1-4, versus minutes 1-2 on the parent), which is where the p95/p99 damage comes from. Moving the accelerated-descent threshold with the floor changes the rejected-coarse-step retry pattern; either way the candidate is a latency regression at equal throughput, so it is rejected under rule 8 and this PR is closed. No follow-up on the floor constants is planned.

https://claude.ai/code/session_012AEx3PnpfHTncWCKXsCH9U

PR #3022, opened 2026-09-05
ci(stress): optional fourth A-B-A segment so candidate drift is measured like control drift

thomhurst:
## Summary

Every exact-SHA A-B-A on the three-broker lanes this week (#3013, #3018, #3019) ran into control drift of 13-28% between the two baseline segments, which voided any candidate whose effect was smaller than that spread. The single candidate segment sits between them with no drift measure of its own, so a candidate that happened to land on a good or bad minute of the VM is indistinguishable from a real effect.

This adds an opt-in `aba_second_candidate` dispatch input (default off, nothing changes without it): after the second baseline the candidate runs once more (A-B-A-B). The comparison then:

- uses the mean of the two candidate samples as the candidate value;
- requires the decisiveness overrides ("better than both controls", "worse than both controls") to hold for the weaker candidate sample;
- treats candidate drift above the control-drift cap as inconclusive, exactly like control drift;
- treats a stability breach or errors in either candidate segment as a regression;
- reports both candidate columns and candidate drift in the summary and `comparison.json` (`candidateSegments`, `candidateB`, `candidateB2`, `candidateDriftPercent`).

Workflow: the input flows through `select-lanes` into the matrix (`aba_second_candidate`, timeout budget `$segment_budget * $aba_segments`), the run block adds the `candidate-b2` segment after `baseline-a2`, the results land in the existing `aba-controls-*` artifact, and the compare step passes `--candidate-b2` only when the segment ran.

## Evidence

- Three-segment behaviour is unchanged: replaying the four real artifacts from this week (runs 33961873612, 33964050486, 33966278396, 33981141065) reproduces their original verdicts (INCONCLUSIVE, PASS, INCONCLUSIVE, REGRESSION) byte-for-byte on the gate columns.
- Four-segment mode exercised on real data by pairing run 33964050486's segments with run 33961873612's candidate as B2 (a cross-VM mix, so an INCONCLUSIVE on throughput/p95/p99 is the expected and correct outcome): p50 shows 0.00% candidate drift and passes at -27%, while the rows where the two candidate samples disagree by 12-25% are flagged inconclusive instead of being decided by whichever sample was measured.
- `test_stress_aba`: 25/25 including five new cases (averaging, candidate-drift inconclusive, overrides needing both samples, second-candidate stability breach, markdown). `test_stress_trend`: 70/70.

## Cost

Only when `aba_second_candidate=true` on a `baseline_sha` dispatch: one extra measured duration (four segments instead of three, ~75 min instead of ~55 at 15 min). Intended for candidates whose expected effect is below the lane's control drift, where a three-segment run is known to be inconclusive in advance.

https://claude.ai/code/session_012AEx3PnpfHTncWCKXsCH9U

## Summary by CodeRabbit

- **New Features**
  - Added optional four-segment A-B-A-B stress testing with a second candidate run.
  - Candidate results are combined for performance metrics, error counts, and delivery consistency.
  - Added candidate drift and stability checks across both candidate runs.
  - Added support for configuring and running the second candidate segment.

- **Reporting**
  - Reports now display separate four-segment comparison tables when the extended test mode is enabled.
  - Results identify the number of candidate segments included in the comparison.

### DIFF
--- a/.github/scripts/stress_aba.py
+++ b/.github/scripts/stress_aba.py
@@ -169,11 +169,24 @@ def _metric_status(
     tolerance_percent,
     max_control_drift_percent,
     floor_status=None,
+    candidate_b2=None,
 ):
     baseline_mean = (baseline_a + baseline_a2) / 2
     if baseline_mean == 0:
         raise ValueError(f"Baseline mean is zero for {metric.label}")
 
+    # With a second candidate segment the candidate value is the mean of both, the
+    # decisiveness overrides must hold for the weaker candidate sample, and candidate drift
+    # is gated exactly like control drift: two candidate samples that disagree by more than
+    # the cap cannot certify a bracketed delta any better than two disagreeing controls.
+    candidate_samples = [candidate] if candidate_b2 is None else [candidate, candidate_b2]
+    candidate_mean = sum(candidate_samples) / len(candidate_samples)
+    candidate_drift_percent = (
+        0.0
+        if candidate_b2 is None or candidate_mean == 0
+        else 100 * abs(candidate_b2 - candidate) / abs(candidate_mean)
+    )
+    candidate = candidate_mean
     delta_percent = 100 * (candidate - baseline_mean) / baseline_mean
     control_drift_percent = 100 * abs(baseline_a2 - baseline_a) / abs(baseline_mean)
     control_spread = abs(baseline_a2 - baseline_a)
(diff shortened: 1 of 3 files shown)


thomhurst:
Rebased onto main after #3017, #3020 and #3021 merged; conflicts resolved so the allocation noise floor and the two-candidate logic compose: the decisiveness overrides use the weaker candidate sample AND the floor, and candidate drift is gated only when the two samples differ by more than the floor. Replaying the four real artifacts with both in place: runs 33961873612 and 33966278396 PASS (allocation floor), 33964050486 PASS, 33981141065 REGRESSION (candidate stability), i.e. unchanged verdict semantics for three-segment runs. `test_stress_aba` and `test_stress_trend` green.

https://claude.ai/code/session_012AEx3PnpfHTncWCKXsCH9U

PR #3066, opened 2026-09-06
fix(sasl): reject malformed OAuth challenges

### LINKED ISSUE #3061: fix(sasl): reject OAuth error challenges with leading whitespace or malformed content
`OAuthBearerAuthenticator.EvaluateChallenge` recognizes an error only when the decoded challenge starts immediately with `{`. Valid JSON error responses with leading whitespace, and arbitrary nonempty text, instead mark authentication complete.

### Evidence
- [Challenge parsing](https://github.com/thomhurst/Dekaf/blob/3511708b4e993824bb87a53e54193cfa46880dce/src/Dekaf/Security/Sasl/OAuthBearerAuthenticator.cs#L136)
- [RFC 7628 section 3.2.2](https://www.rfc-editor.org/rfc/rfc7628.html#section-3.2.2) specifies a JSON error result for failed authentication.

### Reproduction
An isolated harness constructed an actual authenticator with a valid static token, called `GetInitialResponse()`, then evaluated each challenge with a fresh authenticator:

| Challenge | Observed result |
| --- | --- |
| `{"status":"401"}` | Throws `AuthenticationException` |
| The same JSON preceded by a space and newline | Returns null; `IsComplete = true` |
| `invalid response` | Returns null; `IsComplete = true` |
| Empty challenge | Returns null; `IsComplete = true` |

### Expected change
Validate the server challenge according to the OAUTHBEARER protocol. Reject error responses regardless of leading JSON whitespace, and reject malformed or unexpected nonempty responses instead of treating them as success. Preserve empty successful-response handling.

Add regression tests for leading whitespace, malformed nonempty input, ordinary JSON errors, and empty success. Failed challenges must not transition the authenticator to a completed-success state.

### Impact and validation scope
This is a challenge-validation defect, not a demonstrated broker authentication bypass: [KafkaConnection independently checks the Kafka SASL response error code](https://github.com/thomhurst/Dekaf/blob/3511708b4e993824bb87a53e54193cfa46880dce/src/Dekaf/Networking/KafkaConnection.cs#L3530), so ordinary broker-reported authentication failures remain rejected.

Reproduced against the actual authenticator assembly at the linked commit. No live-broker authentication test was performed.

thomhurst:
OAuth error JSON with leading whitespace and arbitrary nonempty challenge data previously set `IsComplete=true`. `OAuthBearerAuthenticator` now rejects every nonempty challenge with `AuthenticationException`, preserving error diagnostics and the existing empty successful-response handling. This follows the failure-challenge semantics in [RFC 7628 §3.2.2](https://www.rfc-editor.org/rfc/rfc7628.html#section-3.2.2).

This fixes authenticator validation; KafkaConnection already independently rejects nonzero broker SASL error codes, so this is not evidence of a broker authentication bypass.

Validation:
- Nine regression cases failed before the fix. Coverage includes JSON whitespace, malformed JSON/text, whitespace-only data, control bytes, invalid UTF-8, ordinary JSON errors, empty success, and incomplete state after failure.
- 183 security unit tests passed on net10.0; 180 passed on net8.0.
- Six real Kafka OAuth integration tests passed (dedicated Kafka 4.0.2 fixture): producer, consumer, admin, JWT token endpoint, round trip, and invalid-token rejection.
- Release builds: zero warnings/errors. `/simplify` and `git diff --check` complete.

Performance evidence, baseline `fd414eb407f010282a9f2a947958ae0d388df394` versus candidate `6effe811113ac2aa7b969bd4c5e98b3b91f4abe9`:

| Successful authentication operation | Before | After | Allocation |
| --- | ---: | ---: | ---: |
| New authenticator + initial response + empty server response | 117.0 ns (error 2.12, SD 1.88) | 119.86 ns (error 4.28, SD 4.00) | 656 B → 656 B |
| Empty challenge evaluation alone | below timing resolution | below timing resolution | 0 B → 0 B |

Handshake mean delta +2.4% lies within measurement uncertainty (overlapping confidence intervals); no speed change claimed. The 656 B are existing per-connection token/initial-response allocations, not per-message allocations. The empty-success branch is unchanged; only invalid server data now throws. No paid stress run or produce/consume hot-path change.

Task-scoped BDN 0.15.8 MemoryDiagnoser fixture, .NET 10.0.11 / SDK 10.0.400, Windows 11 i7-12700K, in-process, five warmups/15 iterations. Same saved baseline DLL versus current product; loaded candidate DLL hash verified. Tiny challenge-only latency is too small to compare reliably (0.0034 versus 0.0189 ns after harness-overhead subtraction), which is why the table uses the complete client-side handshake for timing.

Closes #3061

### DIFF
--- a/src/Dekaf/Security/Sasl/OAuthBearerAuthenticator.cs
+++ b/src/Dekaf/Security/Sasl/OAuthBearerAuthenticator.cs
@@ -135,18 +135,12 @@ public sealed class OAuthBearerAuthenticator : ISaslAuthenticator
     /// <inheritdoc />
     public byte[]? EvaluateChallenge(byte[] challenge)
     {
-        // OAUTHBEARER is a single-round mechanism
-        // If we receive a challenge, it's an error response from the server
-        // The error will be in JSON format as per RFC 7628
+        // Successful OAUTHBEARER authentication has no further challenge data.
+        // RFC 7628 errors use JSON, but malformed or unexpected nonempty replies must also fail.
         if (challenge.Length > 0)
         {
             var errorResponse = Encoding.UTF8.GetString(challenge);
-
-            // If the response starts with '{', it's a JSON error
-            if (errorResponse.StartsWith('{'))
-            {
-                throw new AuthenticationException($"OAUTHBEARER authentication failed: {errorResponse}");
-            }
+            throw new AuthenticationException($"OAUTHBEARER authentication failed: {errorResponse}");
         }
 
         _complete = true;
--- a/tests/Dekaf.Tests.Unit/Security/Sasl/OAuthBearerAuthenticatorTests.cs
+++ b/tests/Dekaf.Tests.Unit/Security/Sasl/OAuthBearerAuthenticatorTests.cs
@@ -115,17 +115,50 @@ public class OAuthBearerAuthenticatorTests
     }
 
     [Test]
-    public async Task EvaluateChallenge_WithJsonError_ThrowsAuthenticationException()
+    [Arguments("")]
+    [Arguments(" ")]
+    [Arguments("\r\n")]
+    [Arguments("\t \r\n")]
+    public async Task EvaluateChallenge_WithJsonError_ThrowsAuthenticationException(string prefix)
     {
         var token = CreateValidToken();
         var authenticator = new OAuthBearerAuthenticator(token);
         _ = authenticator.GetInitialResponse();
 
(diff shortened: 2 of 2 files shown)


thomhurst:
`OAuthBearerAuthenticator.cs:143`: retaining the existing broker-error payload in this focused rejection fix. Truncation/control-character filtering is a separate diagnostic-policy change and should cover both OAuth challenge text and the broker's `SaslAuthenticate` error message consistently. The review identifies pre-existing exposure and explicitly treats it as awareness-only; no additional change here.

PR #3067, opened 2026-09-06
fix(oauth): honor registry token refresh buffer

### LINKED ISSUE #3059: fix(oauth): honor the configured token refresh buffer in Schema Registry authentication
Schema Registry authentication wraps the built-in token provider in an `OAuthBearerAuthenticator` whose separate cache uses a hardcoded 60-second refresh buffer. This prevents the provider from honoring a larger configured `TokenRefreshBufferSeconds` until the outer cache decides to refresh.

### Evidence
- [Authenticator cache](https://github.com/thomhurst/Dekaf/blob/3511708b4e993824bb87a53e54193cfa46880dce/src/Dekaf/Security/Sasl/OAuthBearerAuthenticator.cs#L54)
- [Schema Registry handler wiring](https://github.com/thomhurst/Dekaf/blob/3511708b4e993824bb87a53e54193cfa46880dce/src/Dekaf.SchemaRegistry/SchemaRegistryAuthenticationHandler.cs#L37)
- [Provider uses the configured buffer](https://github.com/thomhurst/Dekaf/blob/3511708b4e993824bb87a53e54193cfa46880dce/src/Dekaf/Security/Sasl/OAuthBearerTokenProvider.cs#L90)

### Reproduction
An isolated .NET harness exercised the actual Schema Registry authentication handler and built-in token provider, with a fake token HTTP endpoint and registry handler:
1. Set `TokenRefreshBufferSeconds = 300`.
2. Return a token with `expires_in = 120`.
3. Send two sequential registry requests.
4. Call the underlying provider directly.

Observed: both registry requests sent `token-1`, with only one token endpoint call. The direct provider call immediately fetched `token-2`, raising the endpoint count to two. The outer cache therefore masks the configured refresh policy.

### Expected change
Use a consistent refresh policy across the provider and authentication handler. Avoid an independent fixed cache window overriding the configured provider behavior.

Add regression coverage for a configured buffer greater than 60 seconds, the default buffer, and a smaller buffer. Preserve efficient reuse of valid tokens and follow repository performance validation requirements for the chosen implementation.

### Validation scope
Reproduced against project assemblies at the linked commit using local HTTP fakes; no live identity service or Kafka broker was involved. No repository source changes were made during review.

thomhurst:
With TokenRefreshBufferSeconds = 300 and a token lasting 120 seconds, Schema Registry reused the first token because an outer authenticator cache imposed its own 60-second window. Configured providers now own refresh timing and cached-token reuse directly. Custom token providers retain their existing authenticator cache. Cached-token requests now return the inner handler task directly; an async helper handles pending tokens.

Validation:
- Regression failed before the fix: the 300-second buffer made one token request instead of two.
- Six unit cases cover large, default, and small buffers, with refresh and reuse for each; four further cases verify pending-token success and cancellation for configured/custom providers; existing authentication precedence tests remain green.
- 2,529 Schema Registry unit tests and 86 OAuth provider/authenticator tests pass across net8.0/net10.0. Release unit/integration builds have zero warnings/errors.
- Four HTTP-pipeline integration tests pass against Kafka 4.0.2 and Schema Registry 7.9.0, including three real local token-endpoint cases exercising the owned provider through the public client constructor. Registry fixture accepts the header; these tests verify client refresh/header behavior, not identity-server authorization.
- `/simplify` and `git diff --check` complete.

Closes #3059

Performance validation: local BenchmarkDotNet 0.15.8 `[MemoryDiagnoser]`, .NET 10.0.11 / SDK 10.0.400, Windows 11, i7-12700K; in-process, 5 warmups and 15 measured iterations per case. Same cached-token HTTP-handler fixture and configuration for each run; request, response, and completed inner-handler Task are preallocated. Token HTTP fetching runs only during setup. Baseline is immediately preceding product SHA 71020dd9a885588ff2e89ecf450fe168cf40f230; final candidate is ae07ef85d154bceff648c531181bd958c00255f1.

| Cached authenticated request | Baseline mean ± error (SD), ns | Candidate mean ± error (SD), ns | Time delta | Allocated before → after |
|---|---:|---:|---:|---:|
| Configured provider | 77.09 ± 1.134 (1.005) | 50.89 ± 0.267 (0.237) | -34.0% | 104 B → 32 B |
| Custom provider control | 77.72 ± 0.647 (0.573) | 69.50 ± 1.226 (1.024) | -10.6% | 104 B → 32 B |

The remaining 32 B is the existing per-HTTP-request Authorization header, outside cached serialization/message paths. The 72 B reduction removes the completed handler Task wrapper. This measures local cached authentication overhead, not Kafka throughput or network latency. CPU/message, p50/p99/max network latency, and long-running stability were not measured by this microbenchmark; no paid stress lane was dispatched.

Two uncommitted prototypes were rejected before the final shape: a common delegate path measured configured/custom 60.34/80.09 ns (104 B each); a larger async method with direct custom-provider calls measured 60.48/82.49 ns (104 B each). Both improved configured access while worsening the custom control. The final synchronous-completion path improves both versus baseline and both prototypes, with lower allocations.

Loaded Schema Registry assembly SHA256 verified in each run: baseline EC64C72918D3AB52C146B842184620ADD4B8F3317D00CD03851CE28B7946C7DD; final candidate 0E3A9AA3925698BBB3A38605EFA5FE78D853684C34C2257A81CA90BDFA618EA8. Local reports/fixture remain under `.artifacts/oauth-buffer-bench` and `.artifacts/bench-{before,after,final,fast}` in the isolated issue worktree. Command: `dotnet run --project .artifacts/oauth-buffer-bench --configuration Release -- --filter '*OAuthBufferBenchmarks*' --inProcess --warmupCount 5 --iterationCount 15 --artifacts <result-directory>`.

### DIFF
--- a/src/Dekaf.SchemaRegistry/SchemaRegistryAuthenticationHandler.cs
+++ b/src/Dekaf.SchemaRegistry/SchemaRegistryAuthenticationHandler.cs
@@ -8,6 +8,7 @@ internal sealed class SchemaRegistryAuthenticationHandler : DelegatingHandler
 {
     private readonly AuthenticationHeaderValue? _staticAuthorization;
     private readonly OAuthBearerAuthenticator? _oauthAuthenticator;
+    private readonly Func<CancellationToken, ValueTask<OAuthBearerToken>>? _configuredTokenProvider;
     private readonly OAuthBearerTokenProvider? _ownedTokenProvider;
 
     internal SchemaRegistryAuthenticationHandler(
@@ -26,18 +27,17 @@ internal sealed class SchemaRegistryAuthenticationHandler : DelegatingHandler
         }
         else if (config.OAuthBearerConfig is not null)
         {
-            Func<CancellationToken, ValueTask<OAuthBearerToken>> tokenProvider;
+            // The configured provider owns caching and its refresh window. An authenticator's
+            // separate fixed window would hide an earlier configured refresh deadline.
             if (oauthBearerTokenProviderFactory is not null)
             {
-                tokenProvider = oauthBearerTokenProviderFactory(config.OAuthBearerConfig);
+                _configuredTokenProvider = oauthBearerTokenProviderFactory(config.OAuthBearerConfig);
             }
             else
             {
                 _ownedTokenProvider = new OAuthBearerTokenProvider(config.OAuthBearerConfig);
-                tokenProvider = _ownedTokenProvider.GetTokenAsync;
+                _configuredTokenProvider = _ownedTokenProvider.GetTokenAsync;
             }
-
-            _oauthAuthenticator = new OAuthBearerAuthenticator(tokenProvider);
         }
         else if (!string.IsNullOrEmpty(config.BasicAuthUserInfo))
         {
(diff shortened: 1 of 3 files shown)


thomhurst:
The naming cleanup was validated and pushed after this PR had already merged. Follow-up #3071 contains the field rename, unchanged method-IL comparison, authentication tests, and before/after allocation/timing evidence.

PR #3068, opened 2026-09-06
fix(retry): keep exponential delays capped after attempt 64

### LINKED ISSUE #3051: fix(retry): prevent exponential backoff from resetting after attempt 64
ExponentialBackoffRetryPolicy computes BaseDelay.Ticks * (1L << (attemptNumber - 1)). The shift count wraps at 64 bits, so delays fall back to short waits after reaching MaxDelay. Checking for negative/zero multiplication results does not prevent this.

Confirmed reproduction:
- BaseDelay = 1 second
- MaxDelay = 30 seconds
- MaxAttempts = 100
- Jitter = false

Observed delays:
| Attempt | Actual | Expected |
| --- | --- | --- |
| 6 | 30 seconds | 30 seconds |
| 63 | 30 seconds | 30 seconds |
| 64 | 30 seconds | 30 seconds |
| 65 | 1 second | 30 seconds |
| 66 | 2 seconds | 30 seconds |
| 67 | 4 seconds | 30 seconds |

This makes retry traffic accelerate again during a prolonged failure.

Evidence:
- [src/Dekaf/Retry/ExponentialBackoffRetryPolicy.cs](https://github.com/thomhurst/Dekaf/blob/3511708b4e993824bb87a53e54193cfa46880dce/src/Dekaf/Retry/ExponentialBackoffRetryPolicy.cs#L31)
- [tests/Dekaf.Tests.Unit/Retry/RetryPolicyTests.cs](https://github.com/thomhurst/Dekaf/blob/3511708b4e993824bb87a53e54193cfa46880dce/tests/Dekaf.Tests.Unit/Retry/RetryPolicyTests.cs#L154) tests attempt 64 but misses the wrap at 65.
- [src/Dekaf/Retry/ExponentialRetryBackoff.cs](https://github.com/thomhurst/Dekaf/blob/3511708b4e993824bb87a53e54193cfa46880dce/src/Dekaf/Retry/ExponentialRetryBackoff.cs#L1) contains related overflow-safe calculations; assess reuse while preserving this public policy's documented jitter semantics.

Acceptance criteria:
- Use saturating/bounded arithmetic so the unjittered delay remains at MaxDelay once the cap is reached.
- Prevent both shift-count wraparound and multiplication overflow from producing shorter delays.
- Add regression cases around attempts 63–67 and larger values, plus representative BaseDelay/MaxDelay combinations.
- Preserve the MaxAttempts stopping contract and documented jitter range/hard maximum.
- Validate nonpositive attempt inputs and invalid configurations consistently with the intended public contract.
- Run the relevant retry-policy tests.

Review validation: reproduced against compiled classes in a temporary harness without changing tracked files. Compilation used RunAnalyzers=false after the normal build encountered analyzer errors under the installed SDK. The full test suite was not run.

Follow CLAUDE.md's performance gates for production-code changes, including before/after measurements for affected hot paths and preservation of protected metrics.

thomhurst:
Exponential retries restarted at attempt 65 because the shift count wrapped. Positive multiplication overflow and jitter near `TimeSpan.MaxValue` could also produce invalid delays. Cache the exact saturation exponent during initialization, then use a bounded shift and an unsigned jitter intermediate to retain the hard cap.

Negative configuration values and nonpositive attempts now throw `ArgumentOutOfRangeException`. Zero delays remain zero; zero attempts disable retries. Tests cover both initialization orders, attempts 63–67 through `int.MaxValue`, and an independent `BigInteger` formula over 130 attempts. Documentation states the boundary contract.

The internal `ExponentialRetryBackoff` helper uses milliseconds and a different jitter range, so reuse would change this public policy's tick precision and 0.5–1.5 jitter contract.

## Validation

- Red reproduction: 16 of 27 new boundary cases failed before the fix.
- Final retry namespace: 81 tests passed on .NET 10 and .NET 8.
- Hosted consumer tests: 37 passed on each framework.
- Five real Kafka 4.3.1 retry-policy integration tests passed.
- Library, unit/integration projects, and repository benchmark project built in Release.
- Documentation production build passed; generated hosted-services page contains the new contract.
- `/simplify` completed; cached state reduced to one integer, with no changes to successful produce/consume paths.

## Performance evidence

Baseline `71020dd9a885588ff2e89ecf450fe168cf40f230` → candidate `6b01d16900088076994b3c78e4572fb239ac1b53`. Saved before/after DLLs, identical runtime-parameter benchmark, serial runs after local builds/tests completed. BenchmarkDotNet 0.15.8, .NET 10.0.11 / SDK 10.0.400, Windows 11, i7-12700K; in-process toolchain, 5 warmups, 15 measured iterations, MemoryDiagnoser. No paid stress dispatch.

| Attempt | Jitter | Before mean ± error (ns) | After mean ± error (ns) | Mean delta (ns) | Allocated before → after |
|---|---|---:|---:|---:|---:|
| 2 | false | 0.6198 ± 0.4300 | 0.1970 ± 0.0929 | -0.4228 | 0 B → 0 B |
| 2 | true | 4.0191 ± 0.2100 | 4.3243 ± 0.2677 | +0.3052 | 0 B → 0 B |
| 6 | false | 0.9281 ± 0.3161 | 0.6029 ± 0.0381 | -0.3252 | 0 B → 0 B |
| 6 | true | 13.5372 ± 0.5953 | 4.3751 ± 0.0654 | -9.1621 | 0 B → 0 B |
| 65 | false | 0.3040 ± 0.1583 | 0.6268 ± 0.0555 | +0.3228 | 0 B → 0 B |
| 65 | true | 3.6920 ± 0.1249 | 4.7245 ± 0.2875 | +1.0325 | 0 B → 0 B |

Attempt 65 changes behavior: the old policy incorrectly jittered a 1-second delay; the fixed policy jitters the 30-second cap. Attempt 6 supplies the equivalent capped-work comparator. Capped jitter improves by about 68%; early jitter confidence intervals overlap. Unjittered sub-nanosecond estimates approach harness-overhead resolution and should not be interpreted as precise percentage gains. Every retry case remains allocation-free.

Earlier runtime-parameter baseline means for attempts 2/65 with jitter false/true were 0.3217/4.3496/0.3602/4.3392 ns, also 0 B. The final early-jitter result agrees with that control. The first arithmetic-only candidate measured 0.5258/4.9838/0.6498/11.7909 ns and was replaced after detecting calculation overhead. Unsigned jitter plus a combined input check measured 0.512/4.440/0.973/4.750/0.803/4.618 ns for attempts 2/6/65, false/true, all 0 B; the final cached threshold removes its extra unjittered work. An initial constant-argument harness was superseded because constant folding limited its usefulness.

Construction allocation was measured separately over 1,024 retained policies: 40 B → 48 B per policy instance for the cached threshold and object padding. This is one-time configuration memory, not allocation per retry or message. These measurements cover retry-policy CPU time and allocations; they do not claim broker throughput or p99 latency improvements. No successful-message path changed.

Reproduce the maintained benchmark with `dotnet run --project tools/Dekaf.Benchmarks -c Release -- --filter '*ExponentialBackoffRetryPolicyBenchmarks*' --inProcess --warmupCount 5 --iterationCount 15` against saved baseline/candidate libraries.

Closes #3051

### DIFF
--- a/docs/docs/hosted-services.md
+++ b/docs/docs/hosted-services.md
@@ -236,6 +236,21 @@ public sealed class OrderProcessorService : KafkaConsumerService<string, Order>
 
 When you supply a DLQ callback to `AddConsumerService`, it passes the registered `DeadLetterOptions` into your constructor directly (see [Dead Letter Queues](consumer/dead-letter-queues.md#enabling-the-dlq)). The options are registered keyed per consumer registration — never as a plain singleton — so one consumer's DLQ settings cannot leak into another service. If you wire the hosted service manually with `AddHostedService`, resolve them with `[FromKeyedServices(typeof(IKafkaConsumer<TKey, TValue>))]` (or your service key).
 
+### Exponential retry delays
+
+`ExponentialBackoffRetryPolicy` calculates `min(BaseDelay * 2^(attempt - 1), MaxDelay)`
+using one-based retry attempts. With `Jitter = false`, the delay stays capped even
+at very large attempt numbers. The default `Jitter = true` multiplies the capped
+delay by a random factor from 0.5 up to 1.5, then caps it again at `MaxDelay`.
+Delays are truncated to whole ticks.
+
+`BaseDelay`, `MaxDelay`, and `MaxAttempts` must be nonnegative; assigning a negative
+value throws `ArgumentOutOfRangeException`. A zero base or maximum delay means
+immediate retries, and `MaxAttempts = 0` disables retries. `BaseDelay` may exceed
+`MaxDelay`; the maximum still applies to the first retry. Calling `GetNextDelay`
+with an attempt less than one throws `ArgumentOutOfRangeException`; attempts
+greater than `MaxAttempts` return `null` to stop retrying.
+
 ## Shutdown Behavior
 
 `KafkaConsumerServiceOptions` (fifth constructor parameter) controls shutdown:
--- a/src/ApiCompatExcludedAttributes.txt
+++ b/src/ApiCompatExcludedAttributes.txt
@@ -1,5 +1,6 @@
 T:System.Runtime.CompilerServices.AsyncIteratorStateMachineAttribute
 T:System.Runtime.CompilerServices.AsyncStateMachineAttribute
+T:System.Runtime.CompilerServices.CompilerGeneratedAttribute
 T:System.Runtime.CompilerServices.EmbeddedAttribute
 T:System.Runtime.CompilerServices.IteratorStateMachineAttribute
 T:System.Runtime.CompilerServices.NullableAttribute
--- a/src/Dekaf/Retry/ExponentialBackoffRetryPolicy.cs
+++ b/src/Dekaf/Retry/ExponentialBackoffRetryPolicy.cs
@@ -1,3 +1,5 @@
+using System.Numerics;
+
 namespace Dekaf.Retry;
 
 /// <summary>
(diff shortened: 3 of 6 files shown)


thomhurst:
Fixed the package-validation failure in 52dcd8cf87e16368fafeb7e49c424c0f9c6445e9. Job https://github.com/thomhurst/Dekaf/actions/runs/34034563068/job/101490173455 failed in `dotnet pack`: auto-property accessors previously carried CompilerGeneratedAttribute, while the new explicit accessors do not. ApiCompat reported six CP0014 removals for each target framework. This was packaging metadata, not a failing retry test.

Added CompilerGeneratedAttribute to the existing compiler/polyfill metadata exclusion list and clarified its purpose. Attribute matching, parameter-name checks, public API analyzers, and released-package validation remain enabled; no public member suppression was added.

Validation:
- Reproduced the exact CP0014 failure locally against Dekaf 1.15.0.
- The same `dotnet pack --no-build --include-symbols --include-source --configuration Release` now passes baseline and cross-framework package validation for net10.0/netstandard2.0.
- Positive local package probe accepts replacing an auto-property with explicit accessors; negative probe still rejects removal of ObsoleteAttribute with CP0014.
- Runtime assemblies are byte-for-byte unchanged before/after the packaging fix: net10.0 SHA256 9CDD1D3D1F4D0DA3F95E4E5BE7B3F082A00A952096CAB3FAFF980963B94A11B5; netstandard2.0 A19EAD1D40EEBA8209FA5E37070BE42217B08890762FA9A32561F0D190E385B2. Existing retry tests and benchmark evidence cover the unchanged product implementation.
- `/simplify` and diff checks complete. No failed CI job was rerun.

The review's awareness-only note about validation differences among IRetryPolicy implementations is retained as broader API-contract context; this fix stays scoped to the reported exponential-backoff overflow and its package gate.

PR #3070, opened 2026-09-06
fix(oauth): coordinate concurrent custom token refreshes

### LINKED ISSUE #3060: fix(oauth): coordinate concurrent custom token refreshes to prevent stale cache overwrites
`OAuthBearerAuthenticator.GetTokenAsync` locks cache reads and writes, but calls the custom token provider outside that lock without coordinating refreshes. Concurrent cache misses can invoke the provider independently, and a slower response carrying an older token can overwrite a newer cached token. The shared Schema Registry authentication handler exposes this path to concurrent requests.

### Evidence
- [Token retrieval and cache assignment](https://github.com/thomhurst/Dekaf/blob/3511708b4e993824bb87a53e54193cfa46880dce/src/Dekaf/Security/Sasl/OAuthBearerAuthenticator.cs#L44)
- [Shared authenticator for custom Schema Registry providers](https://github.com/thomhurst/Dekaf/blob/3511708b4e993824bb87a53e54193cfa46880dce/src/Dekaf.SchemaRegistry/SchemaRegistryAuthenticationHandler.cs#L19)
- [Request authentication](https://github.com/thomhurst/Dekaf/blob/3511708b4e993824bb87a53e54193cfa46880dce/src/Dekaf.SchemaRegistry/SchemaRegistryAuthenticationHandler.cs#L55)

### Reproduction
An isolated harness used the actual authenticator with a custom provider backed by two controllable task completion sources:
1. Start two `GetTokenAsync` calls with an empty cache.
2. Complete the second provider invocation with a newer token expiring in one hour.
3. Complete the first invocation with an older token expiring in 120 seconds.
4. Read the cached token again.

Observed: two provider invocations, and the final cached token was the older token. Concurrent requests also duplicate provider work.

### Expected change
Coordinate refreshes for a shared authenticator so concurrent callers do not race to replace its cache. Define cancellation and failure behavior so a failed or canceled refresh does not leave future callers stuck.

Add deterministic concurrent-refresh regression tests, including the out-of-order completion scenario, failures, and cancellation. Preserve the cached-token fast path and follow repository performance validation requirements.

### Validation scope
Reproduced against the actual authenticator assembly at the linked commit with a custom asynchronous provider. The built-in provider already has its own refresh semaphore; this finding concerns custom providers and the authenticator cache. No live identity service or Kafka broker was involved.

thomhurst:
Concurrent cache misses could invoke a custom OAuth token provider independently, allowing an older response to overwrite a newer cached token. Refreshes now serialize per OAuthBearerAuthenticator. Waiting callers recheck the cache; cancelling a waiter leaves the active provider alone, while a failed or cancelled active refresh releases waiting callers to retry.

Cached reads use volatile publication and a synchronous ValueTask return. Provider code runs outside the coordination lock. Uncontended synchronous refreshes allocate no task or gate; a shared completion signal is allocated only for overlapping refresh callers. The existing 60-second cache refresh buffer remains unchanged. Documentation distinguishes per-authenticator coordination from providers shared across independent clients.

Validation:
- Four deterministic concurrency regressions failed before the fix. Eight final concurrency cases cover reverse provider completion, failed refresh recovery, active/waiter cancellation, synchronous throw/cancellation, cached synchronous completion, and 32 overlapping calls requiring successive refreshes.
- Security suites passed on net10.0 (190 cases) and net8.0 (187 cases), followed by the final 16 concurrency/Schema Registry authentication cases on each runtime after adding the 32-caller case.
- Seven Docker-backed integration cases passed on the final implementation: six Kafka OAUTHBEARER authentication cases plus a shared Schema Registry client making 16 concurrent requests with exactly one custom-provider invocation and consistent Authorization headers.
- Release unit/integration/benchmark builds passed with zero warnings/errors; core netstandard2.0 compiled through net8.0 tests. Docs production build, /simplify, and git diff --check passed.

Before/after performance, exact baseline `0a25915dca5ed05e7528ad28aeedccd179482c20` → candidate `06db8fc09d76b05cc233f34b78e9d2472f067d7e`:

| Benchmark | Before mean ± error | After mean ± error | Mean delta | Allocated before → after |
| --- | ---: | ---: | ---: | ---: |
| CachedToken | 42.67 ± 0.640 ns | 22.92 ± 0.983 ns | -46.3% | 0 B → 0 B |
| SynchronousRefresh | 59.24 ± 0.463 ns | 53.85 ± 1.131 ns | -9.1% | 0 B → 0 B |

BenchmarkDotNet 0.15.8 MemoryDiagnoser, .NET SDK 10.0.400/runtime 10.0.11, Windows 11/i7-12700K; identical committed harness using saved baseline versus candidate DLLs, in-process, 5 warmups/12 iterations. Standard deviations: 0.500/0.361 ns before, 0.711/0.883 ns after. Loaded candidate DLL SHA256 matches product output. The synchronous-refresh case uses a prebuilt expired token to force the refresh path without identity-service I/O; setup excludes one-time initialization. The first semaphore implementation was rejected (136.82 ns synchronous refresh). Final coordination uses no semaphore and retains 0 B for both measured paths. Contended refreshes allocate an occasional completion signal/async wait state; no allocation is added per message or cached request. This is authentication coordination, not a producer/consumer message path; these microbenchmarks do not establish broker throughput, p99/max latency, or CPU per message. No paid stress workflow dispatched.

Closes #3060

### DIFF
--- a/docs/docs/security/oauth.md
+++ b/docs/docs/security/oauth.md
@@ -212,6 +212,18 @@ var producer = await Kafka.CreateProducer<string, string>()
 
 Dekaf automatically refreshes tokens before they expire. The token provider is called whenever a new token is needed.
 
+Each `OAuthBearerAuthenticator` serializes its token refreshes. Concurrent callers
+wait for the active refresh, then reuse its cached token if it remains outside the
+60-second refresh buffer. This also applies to a shared Schema Registry client
+configured with a custom token provider. Separate authenticators do not share this
+gate, so a provider shared across clients must support concurrent calls.
+
+Cancelling a waiting caller stops only that caller's wait. The caller performing
+the refresh passes its token to the provider. If that refresh fails or is cancelled,
+the gate is released and another caller can retry; the failure is not cached.
+Providers must honor cancellation to bound their active work. Tokens returned
+inside the refresh buffer can trigger another refresh on the next call.
+
 ```csharp
 .WithOAuthBearer(async ct =>
 {
--- a/src/Dekaf/Security/Sasl/OAuthBearerAuthenticator.cs
+++ b/src/Dekaf/Security/Sasl/OAuthBearerAuthenticator.cs
@@ -12,6 +12,8 @@ public sealed class OAuthBearerAuthenticator : ISaslAuthenticator
     private readonly Func<CancellationToken, ValueTask<OAuthBearerToken>> _tokenProvider;
     private readonly object _tokenLock = new();
     private OAuthBearerToken? _currentToken;
+    private bool _refreshInProgress;
+    private TaskCompletionSource<bool>? _refreshWaiters;
     private bool _complete;
 
     /// <summary>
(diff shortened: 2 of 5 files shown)


thomhurst:
Clarified the regression comment in 7e1f4367d52d7056edcbde1d4f1ffdfa37c0c0ee. Before the fix, both provider invocations occur and the two completion sources exercise the reported reverse-completion race. After the fix, the second response is intentionally unused: preventing that second invocation is the assertion. Keeping both sources preserves the deterministic pre-fix reproduction. Executable test source is identical after removing comments; existing test and benchmark results remain applicable. /simplify and diff checks complete; no redundant test rerun for this comment-only change.

PR #3095, opened 2026-09-07
test(producer): align throttle clocks with batch timestamps

### LINKED ISSUE #3093: test(producer): align throttle fake clocks with batch timestamps
Two existing producer throttle tests fail on net8.0 at main 47bc43285ad037da8340eb95eac413f039f94f23 while validating the test-only fix for #3091:

- BrokerSenderSendLoopTests.SendLoop_ThrottleObservedDuringConnectionCapacityWait_DelaysRemainingBucket
- BrokerSenderSendLoopTests.SendLoop_PositiveBrokerThrottle_DelaysNextRequestWithoutDeliveryErrors

Both expect a 250 ms throttle delay and observe 1 ms. The net10.0 producer run passes all 1,292 cases; net8.0 passes 1,290 and fails these two. The #3091 enabled/disabled scaling cases pass on both runtimes. The failed run was not rerun unchanged.

The fixtures inject timestamp = 0 while CreateTestBatch stamps StopwatchCreatedTicks from the real Stopwatch clock. ComputeNextWakeupMs subtracts the injected clock from real batch delivery deadlines, then converts the resulting milliseconds to int. On this host, the real counter already exceeds int.MaxValue milliseconds. This puts the fixture outside its intended clock domain and exposes runtime-dependent out-of-range floating-point conversion behavior. Investigate with an exact clock-value probe before deciding whether a separate production clamp is needed.

Acceptance: put each fake clock and batch timestamp in the same domain, advance time relative to that origin, preserve the exact 250 ms assertion and no-send-before-release assertion, and pass both runtimes. Do not increase timeouts or rerun for green. If a valid production scenario can exceed the conversion range, cover and fix it separately with required performance evidence.

Evidence: C:/git/Dekaf-worktrees/issue-3091-deterministic-pressure/producer-tests-net8.log. Product source is unchanged in that worktree.

thomhurst:
Two broker-throttle tests mixed an injected timestamp of zero with batches created using the real Stopwatch clock. On hosts whose counter exceeds int.MaxValue milliseconds, deadline conversion produced a 1 ms wait on net8.0 instead of the expected 250 ms. net10.0 happened to hide the fixture defect by saturating the out-of-range conversion.

Both fixtures now start their fake clock at Stopwatch.GetTimestamp() and advance one second relative to that origin. The exact 250 ms assertion, blocked-send assertion, delivery checks, and existing timeout remain intact.

Validation:
- Both tests failed in the preceding net8.0 producer run with `Expected to be 250 but found 1`; both pass after this correction.
- An isolated probe using the exact deadline conversion measured zero-origin delays above 2,243,273,147 ms: net8.0 returned 1 ms; net10.0 returned int.MaxValue. Using the batch timestamp as origin yielded a 30,000 ms deadline and 250 ms throttle wait on both runtimes. This establishes the clock-domain cause rather than relying on an unchanged rerun.
- All 1,291 Producer namespace cases passed on each of net8.0 and net10.0 (2,582 total).
- Final /simplify review and git diff --check completed. Product source is unchanged; no integration or performance measurement applies.

Closes #3093

### DIFF
--- a/tests/Dekaf.Tests.Unit/Producer/BrokerSenderSendLoopTests.cs
+++ b/tests/Dekaf.Tests.Unit/Producer/BrokerSenderSendLoopTests.cs
@@ -3922,7 +3922,8 @@ public sealed class BrokerSenderSendLoopTests : ScriptedProduceResponseFixture
         var valueTaskSourcePool = new ValueTaskSourcePool<RecordMetadata>();
         var throttleWaitStarted = new TaskCompletionSource<int>(TaskCreationOptions.RunContinuationsAsynchronously);
         var releaseThrottleWait = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
-        var timestamp = 0L;
+        // Batch creation uses Stopwatch timestamps; keep the fake clock in that domain.
+        var timestamp = Stopwatch.GetTimestamp();
         var acknowledgements = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
         var acknowledgementCount = 0;
         var sender = CreateSender(
@@ -3963,7 +3964,7 @@ public sealed class BrokerSenderSendLoopTests : ScriptedProduceResponseFixture
             await Assert.That(await throttleWaitStarted.Task).IsEqualTo(throttleTimeMs);
             await Assert.That(sendSignals[2].Task.IsCompleted).IsFalse();
 
-            Volatile.Write(ref timestamp, Stopwatch.Frequency);
+            Interlocked.Add(ref timestamp, Stopwatch.Frequency);
             releaseThrottleWait.SetResult();
 
             await sendSignals[2].Task.WaitAsync(cancellationToken);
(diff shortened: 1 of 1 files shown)


thomhurst:
The failed CI job is a confirmed, unrelated consumer-runtime bug, now tracked in #3101. I added a synchronized local reproduction that cancels the runtime after the third record is decoded and queues the first handler commit before routing resumes. Both AwaitCapacity and PauseResume fail deterministically with zero commits instead of two (78 ms each).

The backpressure path drains commands with the canceled runtime token before entering the independent shutdown drain budget, canceling the handler prematurely. The original assertion must remain; changing synchronization merely to avoid this interleaving would hide a product defect. No changes were pushed to this producer-test PR, and its failed CI was not rerun. The source fix and its performance/integration gates will be handled under #3101.

thomhurst:
This PR is superseded by merged #3081 (`aefb825462cd6f6d6e956fca3374a12d98a0e128`). Both affected fixtures on current main initialize their fake clock with Stopwatch.GetTimestamp() and advance relative to that origin. The remaining Interlocked.Add versus Volatile.Write(ref timestamp, timestamp + Stopwatch.Frequency) difference is equivalent here: the test continuation is the sole writer, while the sender only reads through Volatile.Read. The exact 250 ms, no-send-before-release, and delivery assertions remain.

I verified the implementation on main `950c3fdd4` in an isolated checkout. Both affected tests pass on net10.0 and net8.0 (four executions); the Release build has zero warnings/errors. No product code or assertions were changed for this verification.

Closing this redundant PR without merging or rerunning its failed CI. The unrelated consumer shutdown/drain defect from its failed job remains tracked in #3101 and is not waived by this closure.

Issue #3105, opened 2026-09-07
test(consumer): investigate classic initial assignment timeout before coordinator failover


thomhurst:
## Observed failure

PR #3100's [consume integration job](https://github.com/thomhurst/Dekaf/actions/runs/34085959529/job/101631642613), exact head `f2b59e6f855ec145b671e2cd3e7de7e8650db52e`, failed `ConsumerCoordinatorFailoverIntegrationTests.ClassicCoordinatorFailover_DuringMemberJoin_Converges` after 60 seconds: `Did not converge within 60s: expected 6 assigned partitions, actual []`. Kafka 4.3.1, Linux net10.0. The failure is at line 372, the first consumer's initial assignment wait, before this test stops any broker or creates the second consumer. The poll task did not fault. The other 166 Consumer cases passed.

The PR adds an isolated telemetry test fixture and does not modify this test, its shared RackAwareKafkaContainer, or product code. The failure occurs in the classic Confluent/librdkafka control. No CI rerun has been requested.

## Investigation and acceptance

- Determine why the initial classic consumer never receives assignments: inspect shared broker recovery/readiness, topic metadata propagation, coordinator state, and poll/error behavior. The initial CI output contains no client group-state diagnostics, so do not claim a cause from the timeout alone.
- Add focused diagnostics and reproduce the causal condition locally or with an explicitly improved experiment. Preserve the actual librdkafka assignment checks and six-partition convergence assertion.
- Fix the cause without simply increasing timeouts, weakening assertions, serializing unrelated tests, or rerunning until green. Add deterministic synchronization or regression coverage appropriate to the established cause.
- Validate the affected classic failover cases and relevant shared-fixture scenarios with Kafka 4.3.1. Follow AGENTS.md if product source changes become necessary.

Saved CI log: `C:/git/Dekaf-worktrees/issue-3096-telemetry-receiver/telemetry-consume-ci-failure.log`. This issue blocks acceptance of #3100 until the failure is understood and addressed.

thomhurst:
Local probes have not reproduced the CI failure yet:

- A deterministic unknown-topic probe subscribes before topic creation, waits for the exact UnknownTopicOrPart error, then creates the topic. The unmodified client defaults converge in 2.3 seconds on Windows, so a metadata-refresh-interval change is not justified by that hypothesis.
[...]

PR #3111, opened 2026-09-07
test(consumer): retain native classic coordinator diagnostics

### LINKED ISSUE #3108: test(consumer): retain bounded native diagnostics for classic coordinator failures
## Purpose

Investigation of #3105 cannot recover librdkafka state from the failing CI report: it contains only the empty-assignment timeout. Deterministic unknown-topic, predecessor-restart sequence, and full Consumer-category probes did not reproduce the failure. The full Linux/.NET 10.0.11, four-processor probe passed all 167 cases.

## Deliverable

Capture a bounded recent history of native cgrp, metadata, and topic diagnostics for classic controls in ConsumerCoordinatorFailoverIntegrationTests. Native log callbacks must enqueue messages without calling consumer APIs; publish the retained history from the owning test context so TUnit includes it in reports. Include timestamps and client identity, cap retained messages, and preserve all assertions, timeouts, client behavior settings, and fixture concurrency.

Verify that an intentional failure produces the native group-state diagnostics in its TUnit report, then remove the injected failure. Validate the affected real-broker cases and include the completed Linux Consumer-category evidence. This is test-only diagnostic infrastructure: it does not fix or close #3105, justify a blind CI rerun, or establish a root cause.

thomhurst:
Classic coordinator failures currently report only an empty assignment and timeout. Native librdkafka callbacks lack TUnit's execution context, so writing directly from those callbacks did not put their state into the failed test's report. Capture the most recent 4096 cgrp/metadata/topic messages per test, with UTC timestamp and client identity, and emit them from the test's after hook only for failure, timeout, or cancellation. The callback only records messages and does not call consumer APIs.

This improves the next #3105 experiment. It does not claim a root cause or fix the original convergence failure. All assignment assertions, timeouts, broker operations, fixture concurrency, and non-diagnostic client configuration stay intact. No src changes.

Validation:

- Injected an intentional exception after initial classic assignment. The failed TUnit report contained 97,600 characters of diagnostics, including native CGRPSTATE, CGRPJOINSTATE, and METADATA messages with client identity. Removed the injected failure before final validation.
- All three classic coordinator cases passed on net10.0 and net8.0 with Kafka 4.3.1 (six executions).
- During investigation, the same diagnostic capture ran through all 167 Consumer-category tests successfully on Linux, .NET 10.0.11, Kafka 4.3.1, DOTNET_PROCESSOR_COUNT=4 / 16 parallel tests, source baseline `950c3fdd4f94086e2852b6a56052f9cc34df9de0`. The three classic cases retained 1034, 746, and 2284 native events respectively.
- Scoped simplification review and diff checks completed. Release builds had zero warnings/errors. No CI rerun was requested; #3105 remains open because the original failure has not been reproduced causally.

Closes #3108

### Review follow-up

Addressed review 5565814686 in `f71e2f5614dbdd40be388ddc4b501b00772f1392`. The after-hook now emits native diagnostics only for Failed, Timeout or Cancelled outcomes. Passing tests avoid the log join/output. The installed TUnit 1.66.16 public API is `TestContext.Current.Execution.Result`, rather than the review's illustrative direct Result property.

A temporary TUnit probe seeded the actual class's diagnostics queue and invoked its real after-hook from passing, intentionally failing and intentionally timing-out tests. On both net10.0 and net8.0, decoded HTML reports prove no sentinel output for the passing case and retained output for both unsuccessful cases (six executions). Expected deliberate failures were preserved as failures. The probe was removed from the test project before the final Release build, which passed with zero warnings/errors. The probe source, logs, reports and verification script were present when validation was recorded, but the merged-worktree sweep later removed this detached checkout. Those raw probe artifacts are no longer retained; the recorded six outcomes and final source commit remain in this PR. No broker operations or product source changed; existing real-broker evidence remains applicable. Final /simplify and diff checks completed.

The review's non-blocking queue observation is documented at the enqueue site: 4096 is a retention target, with transient overshoot bounded by concurrent writers. The thread-safe queue remains appropriate for these native callbacks, and no consumer API is called there. This does not claim an exact atomic capacity cap.

The original convergence investigation #3105 remains open. Awaiting the new CI/review cycle.

### DIFF
--- a/tests/Dekaf.Tests.Integration/ConsumerCoordinatorFailoverIntegrationTests.cs
+++ b/tests/Dekaf.Tests.Integration/ConsumerCoordinatorFailoverIntegrationTests.cs
@@ -11,6 +11,9 @@ namespace Dekaf.Tests.Integration;
 [ClassDataSource<RackAwareKafkaContainer>(Shared = SharedType.PerTestSession)]
 public sealed class ConsumerCoordinatorFailoverIntegrationTests(RackAwareKafkaContainer kafka)
 {
+    private const int ClassicDiagnosticCapacity = 4096;
+    private readonly ConcurrentQueue<string> _classicDiagnostics = new();
+
     private const int PartitionCount = 6;
     private const int MessagesPerPartition = 20;
     private const int MessageCount = PartitionCount * MessagesPerPartition;
@@ -25,6 +28,13 @@ public sealed class ConsumerCoordinatorFailoverIntegrationTests(RackAwareKafkaCo
     /// was stuck on instead, and three of them still fit inside the test timeout.
     /// </summary>
     private static readonly TimeSpan ConvergenceTimeout = TimeSpan.FromSeconds(60);
+
+    [After(Test)]
+    public void PrintClassicDiagnostics()
+    {
+        if (!_classicDiagnostics.IsEmpty)
+            Console.WriteLine(string.Join(Environment.NewLine, _classicDiagnostics));
+    }
 
     [Test]
     [Timeout(240_000)]
@@ -478,7 +488,15 @@ public sealed class ConsumerCoordinatorFailoverIntegrationTests(RackAwareKafkaCo
             EnableAutoCommit = false,
             EnableAutoOffsetStore = false,
             SessionTimeoutMs = (int)SessionTimeout.TotalMilliseconds,
-            HeartbeatIntervalMs = 1_000
+            HeartbeatIntervalMs = 1_000,
+            // Keep native state changes for the next convergence failure. Native callbacks
+            // lack TUnit's execution context, so the owning test writes the bounded log later.
+            Debug = "cgrp,metadata,topic"
+        }).SetLogHandler((consumer, message) =>
+        {
+            _classicDiagnostics.Enqueue($"{DateTimeOffset.UtcNow:O} {message.Name} {message.Facility}: {message.Message}");
+            if (_classicDiagnostics.Count > ClassicDiagnosticCapacity)
+                _classicDiagnostics.TryDequeue(out _);
         });
         if (listener is not null)
         {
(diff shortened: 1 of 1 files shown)


thomhurst:
### Review follow-up

Addressed review 5565814686 in `f71e2f5614dbdd40be388ddc4b501b00772f1392`. The after-hook now emits native diagnostics only for Failed, Timeout or Cancelled outcomes. Passing tests avoid the log join/output. The installed TUnit 1.66.16 public API is `TestContext.Current.Execution.Result`, rather than the review's illustrative direct Result property.

A temporary TUnit probe seeded the actual class's diagnostics queue and invoked its real after-hook from passing, intentionally failing and intentionally timing-out tests. On both net10.0 and net8.0, decoded HTML reports prove no sentinel output for the passing case and retained output for both unsuccessful cases (six executions). Expected deliberate failures were preserved as failures. The probe was removed from the test project before the final Release build, which passed with zero warnings/errors. Probe source, logs, reports and the verification script remain under `C:/git/Dekaf-worktrees/pr-3111-failure-diagnostics/.artifacts/`. No broker operations or product source changed; existing real-broker evidence remains applicable. Final /simplify and diff checks completed.

The review's non-blocking queue observation is documented at the enqueue site: 4096 is a retention target, with transient overshoot bounded by concurrent writers. The thread-safe queue remains appropriate for these native callbacks, and no consumer API is called there. This does not claim an exact atomic capacity cap.

The original convergence investigation #3105 remains open. Awaiting the new CI/review cycle.

thomhurst:
Retention correction: the merged-worktree sweep deleted the detached review checkout and its .artifacts directory. The final pass/fail/timeout probe reports are no longer locally retained; the body now corrects that claim. The original author checkout still retains its separate broker-diagnostic artifacts. This does not change the recorded test outcomes or CI result, but raw follow-up reports cannot be reanalyzed. Future evidence will be archived outside removable worktrees before cleanup.

PR #3122, opened 2026-09-07
perf: remove per-record header enumerator allocations and fix protocol benchmark harness

slang25:
I ran @richlander's `dotnet-inspect` over `Dekaf.dll` to see what its static perf triage would find, and it turned up one genuine per-message allocation on the produce path, plus a couple of small things. This PR is the bits that are safe for 1.x.

- `Record.Write` and `CalculateBodySize` now iterate headers by index rather than `foreach`, so we no longer box an enumerator twice per record with headers. Public type is unchanged.
- `StickyPartitioner.OnBatchComplete` and the consumer prefetch position update use the state-passing `AddOrUpdate` overload with static lambdas, no more closure per batch.
- `ProtocolBenchmarks` had an `[IterationSetup]` which forced `InvocationCount=1`, and at that setting `MemoryDiagnoser` was reporting 0 B for code that allocated 640 B (whoops). Swapped it for a per-call `ResetWrittenCount()` and added pre-built `RecordBatch` cases with and without headers. This will shift the published numbers for the other write benchmarks in that class too.

With the fixed harness, "Write RecordBatch (10 records, 3 headers each)" goes from 640 B to 0 B allocated, and about 4% faster. No-header and read paths unchanged.

There's still a 32 B box per message with headers left in `ConvertHeaders`, where the `HeaderListWrapper` struct gets returned as `IReadOnlyList<RecordHeader>`. Fixing that properly means changing `Record.Headers` to `ReadOnlyMemory<RecordHeader>`, which is a breaking change on a public type, so I think that one waits for 2.0 🙂

Tested with the unit suite (`RecordBatchTests`, `RecordAccumulatorTests`, partitioner tests) and the `ConsumerTests` / `RoundTripTests` integration classes against Docker Kafka.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Added benchmark results comparing Dekaf and Confluent.Kafka across producing, consuming, protocols, serialization, and compression.
  * Added stress-test results covering throughput, memory usage, garbage collection, and test methodology.

* **Performance**
  * Reduced runtime allocations in message processing, partition selection, prefetch handling, and record serialization.

* **Bug Fixes**
  * Improved record-header serialization for sliced arrays and preserved null header values.

* **Tests**
  * Added coverage for header round-tripping and zero-allocation record writing.
  * Expanded serialization benchmarks for records with and without headers.

### DIFF
--- a/docs/docs/benchmarks.md
+++ b/docs/docs/benchmarks.md
@@ -1,0 +1,152 @@
+---
+sidebar_position: 13
+---
+
+# Benchmark Results
+
+Live benchmark comparisons between Dekaf and Confluent.Kafka, automatically updated on every commit to main.
+
+**Last Updated:** 2026-02-04 17:30 UTC
+
+:::info
+These benchmarks run on GitHub Actions (ubuntu-latest) using BenchmarkDotNet. 
+**Ratio < 1.0 means Dekaf is faster than Confluent.Kafka**
+:::
+
+## Producer Benchmarks
+
+Comparing Dekaf vs Confluent.Kafka for message production across different scenarios.
+
+| Method                  | Categories    | MessageSize | BatchSize | Mean        | Error       | StdDev      | Median      | Ratio | RatioSD | Gen0     | Gen1    | Allocated  | Alloc Ratio |
+|------------------------ |-------------- |------------ |---------- |------------:|------------:|------------:|------------:|------:|--------:|---------:|--------:|-----------:|------------:|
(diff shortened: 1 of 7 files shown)


thomhurst:
Nice @slang25 ! Thanks! I'll resolve the merge conflicts 😄

slang25:
Closing this one, I'd branched from a stale fork main (February!) and it turns out every change here already landed upstream months ago, in a nicer shape too 🤦

thomhurst:
@slang25 Will be good to see if that tool spots anything on the current main though 😄

PR #3180, opened 2026-09-10
fix(ci): count expanded BenchmarkDotNet cases

### LINKED ISSUE #3179: Fix automatic performance gate discovery and parameterized case counting
The automatic performance gate counts BenchmarkDotNet `--list flat` method names as executable cases. Parameterized methods expand during execution: run https://github.com/thomhurst/Dekaf/actions/runs/34450471271 successfully ran 31 Dry cases but failed validation expecting 16.

Fix the case count using the Dry run's declared expanded total, retain budget enforcement, and reject missing/failed case reports. Include the scoped-solution/project-directory fix needed to reach execution when the bespoke harness has the same project filename.

Validate with retained successful and failed Dry reports, plus focused harness tests. Product changes and performance thresholds are outside this fix.

thomhurst:
The automatic performance gate rejected successful Dry runs because BenchmarkDotNet's `--list flat` counts methods, while parameters expand into additional cases. In run 34450471271, all 31 cases ran but validation expected 16.

Read the expanded total declared by the Dry execution, enforce the existing case budget, and validate every resulting case. Missing or ambiguous totals still fail. Also scope BenchmarkDotNet project discovery with a local solution and project-directory execution, avoiding the same-named bespoke project.

Validation: 18 harness tests pass; actionlint passes. Reanalysis of the retained Ubuntu Dry output counts and validates all 31 cases. The earlier failed-build artifact remains rejected. No product code or acceptance thresholds change.

Closes #3179

### DIFF
--- a/.github/scripts/performance_gate.py
+++ b/.github/scripts/performance_gate.py
@@ -8,6 +8,7 @@ elapsed time, sample counts and allocations are BenchmarkDotNet's own measuremen
 import argparse
 import json
 import math
+import re
 import subprocess
 import sys
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


thomhurst:
Validation retained at C:/git/Dekaf-evidence/pr-3180/f292ff12c3231a62d50170a1ef319a46d56327b5/validation. All 71 retained files match the SHA-256 inventory. Original successful and failed Ubuntu Dry artifacts, focused test log and source archive are included.

thomhurst:
Confirmed why the local solution is needed alongside `--project`. BenchmarkDotNet 0.15.8 independently discovers the benchmark project when generating its child process: [CsProjGenerator.GetProjectFilePath](https://github.com/dotnet/BenchmarkDotNet/blob/v0.15.8/src/BenchmarkDotNet/Toolchains/CsProj/CsProjGenerator.cs#L245) calls GetSolutionRootDirectory before GetProjectRootDirectory, then recursively searches that root for a project named after the assembly. Passing an explicit project to the outer dotnet run does not configure this child-project search.

Without the local solution, the parent repository solution still wins and both tools/Dekaf.Benchmarks/Dekaf.Benchmarks.csproj and .github/benchmarks/aba/Dekaf.Benchmarks.csproj match. The local solution bounds that independent search; running beside it makes it the closest solution. Keeping both changes is intentional. Current-head review is CLEAR, no inline threads exist, and final build/AOT checks are still running.

PR #3182, opened 2026-09-10
fix(benchmarks): allow absent completion reservations

### LINKED ISSUE #3181: Fix dispatcher comparison harness for nullable completion reservations
Performance comparison 34409075242 completed its loaded dispatcher shards, but the micro and shutdown shards failed compiling the candidate harness: DispatchBenchmarks.cs:83 CS8601. PR #3083 changed CreateCompletionBatch to return null when automatic progress makes explicit reservations unnecessary.

Update the maintained fixture to honor the nullable contract and validate builds against both archived exact product binaries. This repairs harness correctness, not product performance acceptance.

thomhurst:
PR #3083's micro and shutdown comparison jobs fail to compile because CreateCompletionBatch now returns a nullable reservation. Match that contract in the maintained fixture's field; existing enqueue and completion APIs already accept null.

Both archived exact products from run 34409075242 build successfully with this fixture (0 warnings/errors), and both pass the repeated-key batch smoke. Local smoke validates fixture correctness only; this change does not claim performance acceptance.

Closes #3181

### DIFF
--- a/.github/benchmarks/dispatch-aba/DispatchBenchmarks.cs
+++ b/.github/benchmarks/dispatch-aba/DispatchBenchmarks.cs
@@ -22,7 +22,7 @@ public class KeyOrderedDispatchBenchmarks
     private PartitionProcessor<int, int> _processor = null!;
     private PartitionLane<int, int> _lane = null!;
 #if COMPLETION_BATCHES
-    private OffsetCompletionBatch _completionBatch = null!;
+    private OffsetCompletionBatch? _completionBatch;
 #endif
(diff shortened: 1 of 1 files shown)


thomhurst:
Raw builds, original BDN smoke exports, source archive and both loaded binary sets are retained at `C:/git/Dekaf-evidence/pr-3182/076c13d75e190b0bad09778ce15927b5dedb31ba/validation`. All 95 files verify against the SHA-256 inventory; each copied Dekaf.dll also matches the original hosted product artifact.

PR #3197, opened 2026-09-10
ci: select benchmarks for verified pending-fetch changes

### LINKED ISSUE #3187: ci: select relevant consumer benchmarks within the performance gate budget
The automatic performance gate fails before measurement for #3086 (57 expanded cases) and #3083 (103), exceeding the existing 48-case budget. Directory-wide consumer/protocol mapping pulls in unrelated dispatch, CRC and control-plane fixtures; test-only InternalsVisibleTo changes also select the broad core suite. Select maintained fixtures for specific consumer/record-batch paths, retain broad fallback coverage for unknown paths, and ignore project changes consisting only of known test/benchmark friend-assembly declarations. Preserve the 48-case guard, same-VM A1/B/A2, warmup and acceptance rules. Validate affected selections and test that real build-input changes still select the gate.

thomhurst:
#3083 still exceeds the 48-case gate because changes inside PendingFetchData select unrelated KafkaConsumer fetch-scheduling and offset-store methods. Its RecordBatch change only adds a computed instance getter, while the broad map also selects unchanged write methods.

Narrow those two verified scopes and record each reason in job output. Any change outside PendingFetchData, any existing RecordBatch code edit, field addition, auto-property, or unfamiliar shape retains the broader mapping. Keep consumer traversal, parsed-record lifecycle, dispatch, completion tracking and protocol-read coverage. The 48-case guard is unchanged.

The resulting selection is 37 baseline cases and 47 candidate cases, including #3083's ten added completion/epoch cases. All selected cases completed in standard BenchmarkDotNet Dry validation; all 314 Python tests pass on Linux, and the complete-tree artifact check passes. This fixes selection, not product acceptance: candidate-only fixtures still need compatible baseline coverage and the PR still needs its applicable performance evidence.

Closes #3187

### DIFF
--- a/.github/scripts/performance_gate.py
+++ b/.github/scripts/performance_gate.py
@@ -133,9 +133,9 @@ def _area_for(path):
     return match
 
 
-def select(changed_files):
+def select(changed_files, base=None, head=None):
     """Map changed product files to benchmark globs; order follows the first matching file."""
-    areas, filters, not_applicable, considered = [], [], {}, []
+    areas, filters, not_applicable, considered, scopes = [], [], {}, [], []
     for path in changed_files:
         path = path.replace('\\', '/')
         if path in BUILD_INPUTS:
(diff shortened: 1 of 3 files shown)


thomhurst:
Accepted the CI scope fix at d7d5ef545b28216f08764c9c3ef3517d07401ae3. All applicable checks pass, the current-head review is clear, and no review threads or outstanding findings remain. This changes benchmark selection only; it grants no product performance acceptance.

Validation: [Performance tooling tests](https://github.com/thomhurst/Dekaf/actions/runs/34470441097/job/102848832996) and [CI](https://github.com/thomhurst/Dekaf/actions/runs/34470441662/job/102855915566). Original local fixture-validation exports and logs are retained outside Git at `C:/git/Dekaf-evidence/pr-3197/d7d5ef545b28216f08764c9c3ef3517d07401ae3/`; all 62 archived files were SHA-256 verified against their originals.

PR #3204, opened 2026-09-10
test(eventhubs): separate Kafka listener and host ports

### LINKED ISSUE #3203: test(eventhubs): separate container Kafka listener from mapped host port
#3138's Event Hubs job failed during emulator startup. The retained emulator log reports `Error occured while running emulator launcher: Address already in use` and SocketException 98 from Bind; Azurite started successfully. This is an emulator listener collision, not an admin API assertion failure.

The fixture chooses a free port in the host namespace and also configures that number as the Kafka listener inside the emulator. Host availability does not establish availability inside the container. Keep the emulator Kafka listener on 9092 and map an independent random host port; retain the mapped host port in the emulator SASL endpoint so broker metadata remains reachable. Verify this separation with the existing Event Hubs producer, consumer, offset, batch and lifecycle integration coverage.

Preserve the existing narrow startup retry policy, and do not rerun unchanged CI merely to get green. Link the fixed shared fixture back to #3138.

thomhurst:
The Event Hubs fixture used a host-selected free port as the listener inside the emulator too. A free host port can still collide with sockets in the container; #3138 failed with an emulator `Address already in use` startup error.

Keep the container Kafka listener on 9092 and map a separate random host port. The SASL endpoint continues advertising the mapped host port so metadata connections remain reachable.

Validation: all seven Event Hubs integration tests pass with the separated ports. Production code and the existing startup retry policy are unchanged.

Closes #3203

### DIFF
--- a/tests/Dekaf.Tests.Integration/EventHubsEmulatorContainer.cs
+++ b/tests/Dekaf.Tests.Integration/EventHubsEmulatorContainer.cs
@@ -31,6 +31,7 @@ public sealed class EventHubsEmulatorContainer : IAsyncInitializer, IAsyncDispos
     public const string OffsetGroup = "dekaf-offset-group";
     public const string BatchGroup = "dekaf-batch-group";
 
+    private const ushort KafkaPort = 9092;
     private const ushort AzuriteBlobPort = 10000;
     private const string AzuriteAlias = "azurite";
     private const string EmulatorAlias = "eventhubs-emulator";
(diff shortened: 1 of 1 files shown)


thomhurst:
Accepted test infrastructure fix at 7cc445a531a41901d6853ecd0f99c0479cb7833a. [CI](https://github.com/thomhurst/Dekaf/actions/runs/34479196681) passed, including Event Hubs; current-head review is CLEAR with no actionable comments or unresolved threads. No product performance claim applies. Original failure and local integration evidence remain hash-verified at C:/git/Dekaf-evidence/pr-3204/7cc445a531a41901d6853ecd0f99c0479cb7833a.

PR #3221, opened 2026-09-10
fix(ci): partition oversized performance gates into validated batches

thomhurst:
Automatic performance gates stop before measurement when parameter expansion exceeds the 48-case budget: recent jobs selected 58, 78, 85 and 179 cases. This change plans up to four batches of at most 48 validated cases, with two concurrent jobs. Every selected case is retained, and each batch builds both pinned revisions and runs A1/B/A2 sequentially on one Ubuntu VM. The existing final check requires every batch and combines verdicts, never metrics across runners.

Preparation now names every filter missing from either revision and rejects different baseline/candidate case sets. Each Dry and measured batch must match its exact planned identities, preventing a successful comparison of silently reduced or substituted coverage. Failed preparation and batch logs are surfaced in job output; explicit Bash enables pipeline failure propagation. Original exports remain artifacts. Measurement settings, tolerances, regression decisions and loaded-evidence requirements are unchanged.

Measurement jobs explicitly require successful preparation and a nonempty matrix. Discovery derives its limit from the shared batch constants, and FullName mismatches identify the case and both revision-specific filters.

Validation: all 39 performance-tooling tests pass on Linux, six artifact-policy tests pass, the complete committed tree passes artifact policy, both workflows pass actionlint, and the standard benchmark project builds without warnings. GitHub's expression library validates the actual measurement condition across 24 preparation states. Real BenchmarkDotNet Dry exports for two parameter values round-trip through two one-case plans; each generated filter executes exactly its expected case. Replaying the original 179-case discovery export produces 48/48/48/35 batches without loss. CI repeats the parameterized filter smoke test. These are tooling correctness checks, not product performance acceptance.

Cost is explicitly bounded at 192 cases across four jobs; each job keeps the existing 48-case and 330-minute limits. Preparation adds a build/Dry pass before the per-VM validation. This fixes scheduling and coverage validation; missing baseline fixtures, product regressions and loaded-coverage gaps still require their own fixes.

### DIFF
--- a/.github/benchmarks/README.md
+++ b/.github/benchmarks/README.md
@@ -5,8 +5,11 @@ See [standard tooling and evidence storage](STANDARD-TOOLS.md) and the repositor
 [performance requirements](../../AGENTS.md).
 
 The [automatic performance gate](../workflows/performance-gate.yml) selects maintained
-fixtures for changed product paths and runs A1/B/A2 sequentially on one hosted Ubuntu
-VM. It retains original exports and the reviewed screen in GitHub job output/artifacts.
+fixtures for changed product paths and divides validated expanded cases into batches
+of at most 48. Each batch runs A1/B/A2 sequentially on one hosted Ubuntu VM; the final
+check requires every batch. Up to four batches run, with two concurrent jobs. Missing
+fixtures and differing A/B case sets fail preparation before acceptance measurement.
+It retains original exports and the reviewed screen in GitHub job output/artifacts.
 For a targeted manual run, dispatch from main with `pr`, exact fresh-main `base_sha`,
 current PR `head_sha` and the affected standard-project `filters`. The workflow rejects
 stale pins and closed PRs. Fresh main, candidate ancestry, identical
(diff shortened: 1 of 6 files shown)


thomhurst:
Addressed all three points from [the review](https://github.com/thomhurst/Dekaf/pull/3221#issuecomment-5622039648) in `fc98d4440341cac97637d6c9314e5ef49b9d2093`:

- Measurement explicitly requires `needs.plan.result == 'success'`, applicability, and a nonempty matrix. The actual workflow expression passes 24 success/failure/cancellation/skip and output-state checks using GitHub's `@actions/expressions` evaluator. One clarification: [GitHub documents an implicit `success()` check unless a status-check function overrides it](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#status-check-functions), so the new guard makes the dependency contract explicit rather than relying on the review's claimed removal of that default.
- Discovery uses `count --discovery`, deriving its budget from `MAX_CASES * MAX_BATCHES`. Tests vary both constants and verify discovery boundaries while the per-job limit remains enforced.
- A/B FullName mismatches now identify the shared case and both differing filters. A regression test covers a casing-only difference without changing the case identity.

Kept one commit and rebased onto current main `0ab965c1a`. All 39 tooling tests pass on Linux; actionlint and the complete-tree artifact check pass. New-head CI/review will run after this push. No product source or performance tolerances changed.

Issue #3193, opened 2026-09-10
benchmarks: add compatible binary-key dispatch coverage on main


thomhurst:
PR #3082 adds binary-key dispatch benchmarks only on its candidate branch, so the standard A1/B/A2 gate cannot compare those workloads against main. Its existing partition dispatch fixture also needs setup compatibility with both private processor factory signatures.

Land the repeated-key, distinct-key and shared-suffix collision fixtures in the existing tools/Dekaf.Benchmarks project on main. Preserve identical fixture source and correctness checks across main and #3082, use normal BenchmarkDotNet execution with MemoryDiagnoser, and validate both revisions with Dry. Keep reports in GitHub output/artifacts. This supplies fixture coverage only; it does not accept the product change or remove the existing budget and collision-performance blockers.

PR #3227, opened 2026-09-10
ci: make the performance gate fast and reliable

thomhurst:
## Problem

The automatic performance gate blocked every open PR for days. Measured from its first day (2026-09-09 23:13 to 2026-09-10 18:07, 122 completed runs):

| | |
|---|---|
| runs failed / cancelled / passed | 82 / 14 / 25 |
| runner minutes in one day | 3,876 |
| run duration | median 7 min (fast preparation failures), up to 3 h 4 min |
| same-binary control drift (A1 vs A2, 175 case pairs) | median 1.2%, p90 8.5%, max 28.8%; 17% of pairs above the 5% tolerance |

Two failure modes. Preparation failed with `filters match no benchmark methods` whenever a PR added a fixture that `main` lacked, which forced fixture-first PR splits (#3223/#3222, #3216/#3171, #3208/#3158, #3193/#3082). Measured runs returned `REGRESSION` on deltas the same size as the drift between two runs of identical code, for example three regressions plus two inconclusive cases out of twenty on the Polyfill bump in #3213, and six regressions for +40 B/op on a fixture that allocates 12 to 790 KB/op.

## Changes

**Selection** (`performance_gate.py select`): four hot-path sentinel classes for any product or build-input change, a few classes per hot component directory, plus every fixture class the PR adds or edits. Non-steady-state fixtures are skipped with a reason. The 60-entry per-file map and the PR-specific regex scoping (`_pending_fetch_only`, `_record_count_bound_only`, `_friend_assembly_only`) are removed.

**Measurement** (`performance-gate.yml`): one `ubuntu-latest` job per class. Each job builds the PR's fixture source against both product revisions, falling back to the baseline's own fixtures when the PR fixtures need new APIs, then runs A1, B and A2 back to back with 10 warmup and 15 measured iterations of 500 ms. The two builds overlap, up to 8 classes run at once, and jobs time out at 75 minutes. A 16-case class finishes in about 12 minutes of measurement; the two controls are minutes apart instead of hours. Nothing has to land on `main` first. When the baseline builds its own fixtures, cases whose fixture file differs are reported as `NOT COMPARED`, and a class the PR adds is measured on the candidate alone.

**Screen** (`performance_gate.py screen`): medians instead of means; tolerance widens for small operations (20% under 100 ns, 15% under 1 µs, 10% from 1 µs); allocation regressions need at least one managed object (24 B/op) and 1% of the baseline; a regression must hold against both controls and reproduce on an immediate repeat of exactly the regressed cases before the job fails. Control drift and one-sided losses are notes. The jq screen, Dry discovery, batch planning, four validation layers and the summary step are gone; the gate job only combines job results.

**Agent instructions**: `CLAUDE.md` shrinks from 2,885 to 1,165 words. The new performance section says what the gate measures, what a verdict means, how to read a red gate, and what not to do (rerun until green, widen tolerances, add PR-specific rules, split fixtures into separate PRs). `.github/benchmarks/README.md` and `STANDARD-TOOLS.md` match.

## Evidence

Replaying the new screen against the retained BenchmarkDotNet exports of six recent gate runs:

| run | old verdict | new verdict |
|---|---|---|
| #3213 Polyfill bump (20 cases) | REGRESSION (3 cases +6.8% to +13.1%) | PASS |
| #3085 (20 cases) | INCONCLUSIVE | PASS |
| #3086 (40 cases) | REGRESSION (2 cases +5.9%, +6.2%) | PASS |
| #3083 (37 cases) | REGRESSION (5 cases) | REGRESSION: `PartitionedOffsetTracking.CompleteInOrder` +139% and +126 B/op, `CompleteOutOfOrder` +48 B/op, `KeyOrderedDispatch.DispatchLifetime` +18%/+24% |
| main dispatch for #3158 (18 cases) | REGRESSION (10 cases, six of them +40 B/op on 12 to 790 KB/op) | REGRESSION on one case only (+28%/+18%), which the repeat step now re-measures before failing |

Unit tests: 25 pass. `actionlint` passes on both workflows. Repository artifact policy passes on the full tree.

Validation dispatches of this branch's workflow are linked in the comments below.

https://claude.ai/code/session_015uHufZdERnxZ4op1MssMYZ

## Summary by CodeRabbit

* **Performance Testing**
  * Redesigned the performance gate to run targeted benchmark classes for relevant changes.
  * Added A1/B/A2 median comparisons, allocation thresholds, size-based tolerances, and repeat confirmation for suspected regressions.
  * Added support for manual runs with optional base and head revisions.
  * Reports now provide Markdown summaries through job summaries and artifacts.

* **Documentation**
  * Updated benchmark guidance for steady-state fixtures, standardized settings, result interpretation, and stress testing.


thomhurst:
Reviewed 2a809965e2704227c8b19374ad61e6710070b69e against the maintainer instructions supplied to this queue session. Three acceptance blockers need resolution:

- `BDN_SETTINGS` uses 10 warmup iterations at 500 ms, approximately five seconds, and the new screen does not enforce the required 20-second elapsed workload warmup. Keep the faster class scheduling, but preserve and verify elapsed warmup in every phase.
- `TOLERANCE_TIERS` changes 5% to 10–20%, changes the statistic from mean to median, and raises allocation tolerance from the declared 8 B/op screen floor to 24 B plus 1%. This relabels already-measured losses. No approval for those acceptance-policy changes is present in this session; retain the current contract or record explicit maintainer direction before using the new labels to accept existing PRs.
- Baseline-fixture fallback compares matching case names from different fixture source. Matching names do not establish identical work when a PR edits the implementation or inputs. Validate fixture equivalence for each compared case, or mark that scope missing/invalid instead of accepting it from the name intersection.

These concerns do not oppose reducing preparation cost or running classes concurrently. They concern whether the resulting screen can support the existing acceptance contract.

thomhurst:
Validation of this branch's workflow (dispatched on the branch itself):

- Identical product, filters `*.Unit.ProduceResponseParsingBenchmarks.* *.Unit.FetchResponseParsingBenchmarks.*`: **PASS** in 16 minutes wall clock, https://github.com/thomhurst/Dekaf/actions/runs/34517396983. `ParseLegacyV6` showed A1 to A2 drift of +14.4% with B at +10.9% against A1 and -3.1% against A2 on identical binaries; the screen recorded it as a note, which is exactly the false-positive shape the old 5% mean rule failed on.
- PR #3083 head `aebe135c` against its merge base `4ec38bfc`, filters `*.Unit.PartitionedOffsetTrackingBenchmarks.* *.Unit.KeyOrderedDispatchBenchmarks.*`: **PASS** in 21 minutes, https://github.com/thomhurst/Dekaf/actions/runs/34517404245. The PR fixtures did not compile against the baseline, so the fallback path ran. Since then the fixture-diff rule below applies to that situation.
- A reversed comparison of the merged perf change #3117 (baseline = merge commit, candidate = its parent) is running to exercise the repeat-on-regression path: https://github.com/thomhurst/Dekaf/actions/workflows/performance-gate.yml?query=branch%3Aci%2Fperf-gate-reliability

Review feedback addressed in the latest push:

- Steady-state detection now strips comments and judges each public class separately (Codex and Greptile P1). `CompressionBenchmarks` and `SchemaRegistryRuleExecutorBenchmarks` are steady-state again; `SchemaRegistryCelFreshContextBenchmarks` alone is skipped.
- Baseline-fixture fallback (Greptile P1, third blocker in the queue review): when the baseline had to build its own fixtures, `screen --fixture-diff BASE HEAD` marks every case from a fixture file that differs between the revisions as `NOT COMPARED` with its absolute numbers, and marks everything `NOT COMPARED` if shared benchmark infrastructure changed. Matching names are no longer treated as equal work.
- Explicit dispatch filters get unique matrix names (Greptile P2).

On the first two blockers in the queue review: the 20-second elapsed-warmup floor, the 5% mean tolerance and the 8 B/op allocation floor are the rules this PR replaces, at the maintainer's direction, because the retained exports show they sit below the noise of identical binaries on hosted runners (A1 vs A2 drift p90 8.5%, max 28.8%, 17% of case pairs above 5%). The 20-second figure came from a loaded shutdown-latency study; a 50 ns microbenchmark reaches tier-1 code within the first second of a 5-second hot loop, and the identical-product run above confirms the warmup is sufficient. The new contract is recorded in `CLAUDE.md` in this PR; the screen does not relabel any accepted evidence, it defines the acceptance rule going forward.

https://claude.ai/code/session_015uHufZdERnxZ4op1MssMYZ

thomhurst:
Positive control for the repeat-on-regression path: https://github.com/thomhurst/Dekaf/actions/runs/34524557900 compared the merged perf change #3117 in reverse (baseline = merge commit `00543c10`, candidate = its parent `50b756b0`).

- `PartitionedDispatchBenchmarks`: first pass regressed all 12 cases; the workflow re-measured exactly those cases and the verdict is **REGRESSION reproduced on repeat**, +43% to +703% time and +49 to +3,019 B/op against both controls. Job time 38 minutes including the repeat.
- `KeyOrderedDispatchBenchmarks`: the candidate (older commit) has no such class, so the job failed with `Invalid measurement: B: no benchmark cases were exported` rather than passing on an empty comparison. In the pull-request flow classes are selected from the head's own fixture files, so this only happens on a manual dispatch with a stale filter.
- The head's older fixtures did not compile against the newer baseline, so the baseline-fixture fallback ran here as well.

Latest push: steady-state detection now attributes a class's own attribute block (for example `[SimpleJob(RunStrategy.Monitoring)]`) to that class rather than the preceding one, the measurement job timeout is 75 minutes because BenchmarkDotNet overhead makes a case cost about 30 s per phase and a repeated regression doubles the measurement, and the documented tolerance boundary reads "10% from 1 µs".

https://claude.ai/code/session_015uHufZdERnxZ4op1MssMYZ

thomhurst:
Addressed the blocking review on `06b54fd` in `6f0081248` (squashed, rebased onto `75ef9a4e5`):

- **New fixture class on the baseline-fixtures fallback.** After the baseline builds its own fixtures, the job runs `--list flat --filter "$FILTER"` on the baseline binary (verified locally: an absent class prints nothing and exits 0, a present class prints one full name per benchmark). When nothing matches, the job skips A1/A2, measures B alone and passes `--baseline-lacks-filter` to the screen, which reports every candidate case as `NOT COMPARED` with a warning instead of raising `no benchmark cases were exported`. The flag is strict: A1/A2 must be empty, and every candidate class must come from a fixture file that differs between the revisions, so a tooling failure on an unchanged class is still an invalid measurement. Unit test covers the single-class empty-baseline shape, both contradictions, and the CLI path with empty A1/A2 directories.
- **Candidate build exit status.** `wait "$candidate_build"` is now an explicit check that fails the step with an `::error::` naming `build-B.log` (under the runner's `bash -eo pipefail` the bare `wait` already failed the step, but without saying why).
- **Free-form manual comparisons.** The dispatch pins step warns and notes in the summary when the baseline is not an ancestor of the candidate, so a mistyped `base_sha` is visible while reverse comparisons such as the #3117 positive control still run.

`CLAUDE.md`, `.github/benchmarks/README.md` and the script docstring say that a class the PR adds is measured on the candidate alone. 25 unit tests pass, `actionlint` passes, artifact policy reports 0 violations.

https://claude.ai/code/session_015uHufZdERnxZ4op1MssMYZ

PR #3228, opened 2026-09-10
benchmarks: cover follower offset recovery on both revisions

thomhurst:
Adds the standard-project fixture needed to compare #3086 against main. Ten BenchmarkDotNet cases cover normal fetch, follower and leader errors, a controlled two-response sequence, and pooled-response control through both fetch handlers. The existing BufferedConsumerHarness owns setup-only private bindings; path selection covers consumer and metadata changes.

The baseline incorrectly resets a follower error to Latest; the candidate preserves offset 42. Setup validates those declared contracts explicitly. The two-response case restores offset 42 on both revisions before the leader handler, so it measures identical handler work and does not claim to measure actual retry scheduling or the baseline's offset lookup. Loaded recovery evidence remains separate.

Validation: all ten Dry cases pass on baseline and #3086 with identical fixture/helper source; 42 selector tests pass; full-tree artifact policy passes. Original local exports and binaries are retained outside Git. No product performance PASS is claimed. This PR contains only fixture/helper source and selector changes.

### DIFF
--- a/.github/scripts/performance_gate.py
+++ b/.github/scripts/performance_gate.py
@@ -47,7 +47,7 @@ AREAS = (
     # partition dispatcher has its own implementation; unknown files retain the
     # directory-wide fallback below.
     ('src/Dekaf/Consumer/KafkaConsumer.cs', 'consumer', unit(
-        'ConsumerHotPathBenchmarks', 'ConsumerBatchInterceptorBenchmarks', 'FetchResponseParsingBenchmarks', 'FetchRequestBuildBenchmarks',
+        'ConsumerHotPathBenchmarks', 'ConsumerBatchInterceptorBenchmarks', 'ConsumerFollowerOffsetRetryBenchmarks', 'FetchResponseParsingBenchmarks', 'FetchRequestBuildBenchmarks',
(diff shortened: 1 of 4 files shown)


thomhurst:
Accepted the benchmark-tooling scope at 0ef25af46ed53fc1360a08ba91283f6394b010c8. All checks are terminal and green, with current-head Claude CLEAR, completed Codex review, no inline findings or review threads, and no outstanding review-body concerns. CI: https://github.com/thomhurst/Dekaf/actions/runs/34517825888

The identical maintained fixture/helper validates all 10 cases against baseline and #3086's implementation; selector tests pass (42). Dry runs establish fixture correctness, not product performance acceptance. The controlled two-response case excludes retry routing and offset lookup; #3086 still needs its applicable performance and loaded recovery evidence. The duplicate response-pool parameter case is harmless and does not alter that boundary.

Raw fixture validation reports, binaries, sources and logs are retained outside the repository under C:/git/Dekaf-evidence/pr-3228/0ef25af46ed53fc1360a08ba91283f6394b010c8 and the paired pr-3086/compatible-fixture archives. The final baseline archive's 10 files have been rechecked against their SHA-256 manifest. No generated evidence is committed.

Issue #3232, opened 2026-09-10
perf(outbox): remove SaveChanges notification overhead


thomhurst:
## Current status — 2026-09-11

PR #3234 merged the duplicate-change-detection correction: the observer examines Added-state entries and listens to EF's existing detection-completed event. Its tests cover navigation-discovered inserts, pooled contexts, disabled detection, cancellation and reentrant saves. The [automatic performance screen](https://github.com/thomhurst/Dekaf/actions/runs/34525987924) now succeeds.

This does not complete the issue. #3234 explicitly excludes parity with the pre-notification baseline and resolution of the previously recorded loaded idle-CPU regression. The original regression numbers below describe #3171 before this correction; they are not measurements of current main.

## Remaining work

- Measure the corrected notification path against both current main for further changes and the pre-feature baseline using the maintained OutboxSaveChangesBenchmarks.
- Remove or explain remaining overhead on unaffected saves and large contexts, including the per-save allocation difference from the pre-feature baseline.
- Investigate and resolve the loaded idle-CPU regression using existing stress tooling and the repository's acceptance rules.
- Preserve immediate post-commit notification, rollback, autocommit, explicit transactions, cancellation, polling fallback and reliable change tracking.
- Record exact-SHA comparisons and applicable loaded acceptance in GitHub job output/artifacts. The duplicate scan is fixed; do not repeat that implementation as the remaining task.

## Original report and full acceptance criteria

Historical problem statements are retained for context; the original acceptance requirements remain in force.

The SaveChanges notification path merged in #3171 has a demonstrated performance regression in the maintained BenchmarkDotNet fixture. The exact-SHA, same-VM A1/B/A2 run reports 15 REGRESSION cases and one INCONCLUSIVE case (candidate warmup shortfall): https://github.com/thomhurst/Dekaf/actions/runs/34511656032

Large tracked contexts lose roughly 52–133% against both controls in the affected cases, with approximately 205–207 KB more allocations per save. Even no-outbox saves with zero tracked business rows allocate 152 B versus 32 B in both controls. The extra ChangeTracker.Entries scan triggers redundant change detection. Earlier local event/interceptor prototypes did not meet the allocation contract and were discarded; no passing replacement has been demonstrated. The previously recorded loaded idle-CPU regression also remains unresolved.

[...]

PR #3234, opened 2026-09-10
perf(outbox): avoid duplicate change detection during saves

thomhurst:
Outbox notification interception currently invokes EF change detection before EF's normal SaveChanges pass, scanning and allocating for unchanged business entities twice. This change inspects only the Added state map and observes EF's existing detection-completed event, preserving navigation-discovered inserts and notification ordering before user SavedChanges handlers. A context-scoped cached state manager avoids repeated service resolution, and transaction tracking reuses an immutable marker.

Validation: all 113 outbox tests pass on net10.0 and net8.0, including new single-detection, navigation discovery, pooled context, disabled detection, cancellation and reentrant save coverage. The product builds for both frameworks without warnings. EF's internal state-map access is deliberately narrow and covered against EF 8 and 10; unfamiliar collection shapes retain an enumeration fallback.

All 16 cases in the maintained OutboxSaveChangesBenchmarks validate in local BenchmarkDotNet Short runs. Against the local current-main control, unchanged saves improve approximately 21–25% at zero business rows and 59–61% at 1024 rows; allocations decrease from 152 to 144 B and from 409960 to 204968 B respectively. These are diagnostic results, not hosted performance acceptance. Fresh exact-head CI and A1/B/A2 remain required. EF/SQLite allocations are per-save costs, not Kafka per-message allocation evidence.

Related: #3232. This fixes duplicate change detection; it does not claim the original notification feature has achieved parity with its pre-feature baseline or resolve its previously recorded loaded idle-CPU regression. Evidence stays outside Git under C:/git/Dekaf-evidence/issue-3232; hosted job output/artifacts will supply acceptance. Only product code and correctness tests are committed.

### DIFF
--- a/src/Dekaf.Outbox.EntityFrameworkCore/OutboxCommitObserver.cs
+++ b/src/Dekaf.Outbox.EntityFrameworkCore/OutboxCommitObserver.cs
@@ -2,7 +2,10 @@ using System.Data.Common;
 using System.Runtime.CompilerServices;
 using System.Transactions;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.ChangeTracking;
+using Microsoft.EntityFrameworkCore.ChangeTracking.Internal;
 using Microsoft.EntityFrameworkCore.Diagnostics;
+using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Storage;
 
 namespace Dekaf.Outbox.EntityFrameworkCore;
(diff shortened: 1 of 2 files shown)


thomhurst:
The PR was merged at a22e8b40a while its exact-head performance comparison remains in progress: https://github.com/thomhurst/Dekaf/actions/runs/34525987924. The merge itself does not establish performance acceptance; no hosted PASS has been reviewed or recorded here yet. Issue #3232 remains open for the outstanding notification-cost scope and historical baseline comparison.

PR #3270, opened 2026-09-11
test(consumer): establish prefetch before timeout assertion

### LINKED ISSUE #3263: test(consumer): establish prefetch lifetime before timeout assertion
`ConsumeOneFastPathTests.ConsumeOneAsync_TimeoutDoesNotCancelPrefetchLoop` can fail before it establishes the prefetch lifetime it intends to test.

Observed during #3258 validation on .NET 10, Windows x64: 1,416 consumer tests passed and this test failed at `Assert.That(prefetchCts).IsNotNull()`. The test uses a 10 ms poll timeout and assumes `StartPrefetch()` has already run. `ConsumeOneWithTimeoutAsync` arms the timeout before entering `ConsumeOneCoreAsync`; the latter intentionally skips `StartPrefetch()` when the poll token is already canceled. A scheduling or JIT delay can therefore leave `_prefetchCts` null without canceling an existing prefetch loop. This consumer path is unchanged by #3258.

Make the test establish and observe the prefetch lifetime before the timed poll, then assert that the same lifetime remains uncanceled after timeout. Keep the separate timeout-returns-null test. Do not increase a timing allowance, weaken the assertion, or change consumer cancellation behavior to satisfy this fixture.

Evidence: `C:/git/Dekaf-evidence/issue-3258/scalar-storage/unit-net10-inlined.log`; exact failing unit binaries, source and patch are retained with a SHA-256 inventory in `candidate-validation/` under that directory. The observed failure establishes the missing prefetch state; the cancellation-before-start ordering is the source-based diagnosis, not an instrumented scheduler trace. No rerun was used to hide the failure.

thomhurst:
The prefetch-lifetime test can time out before `ConsumeOneCoreAsync` starts prefetch, then incorrectly fail because no cancellation source exists. Start the real prefetch loop before arming the existing 10 ms poll timeout, capture its cancellation source, and verify the same source remains uncanceled afterward. The separate timeout-returns-null test remains in place.

This is a test-only synchronization fix. Setup invokes the actual private `StartPrefetch` method through the fixture's existing reflection approach; it does not substitute a fake lifetime or extend a deadline. Consumer behavior is unchanged.

Validation: all 50 `ConsumeOneFastPathTests` cases passed on net10.0 and net8.0 in Release. The repository artifact check and final diff review passed. The original intermittent failure is retained under the evidence linked in #3263; no repeated baseline run was used to manufacture a failure or green result.

Final source, loaded test binaries, reports and SHA-256 inventory are archived outside removable worktrees at `C:/git/Dekaf-evidence/issue-3263/prefetch-lifetime/`.

Closes #3263

### DIFF
--- a/tests/Dekaf.Tests.Unit/Consumer/ConsumeOneFastPathTests.cs
+++ b/tests/Dekaf.Tests.Unit/Consumer/ConsumeOneFastPathTests.cs
@@ -1031,12 +1031,21 @@ public sealed class ConsumeOneFastPathTests
     {
         await using var consumer = CreateInitializedConsumer(queuedMinMessages: 2, fetchMaxWaitMs: 1);
 
-        var result = await consumer.ConsumeOneAsync(TimeSpan.FromMilliseconds(10), CancellationToken.None);
-
-        await Assert.That(result).IsNull();
+        // Establish the real prefetch lifetime before arming the short poll timeout.
+        // A poll canceled before ConsumeOneCoreAsync starts need not create a loop.
+        var startPrefetch = typeof(KafkaConsumer<string, string>)
+            .GetMethod("StartPrefetch", BindingFlags.NonPublic | BindingFlags.Instance)
+            ?? throw new InvalidOperationException("StartPrefetch method not found.");
+        startPrefetch.Invoke(consumer, null);
         var prefetchCts = GetPrefetchCts(consumer);
         await Assert.That(prefetchCts).IsNotNull();
         await Assert.That(prefetchCts!.IsCancellationRequested).IsFalse();
+
+        var result = await consumer.ConsumeOneAsync(TimeSpan.FromMilliseconds(10), CancellationToken.None);
+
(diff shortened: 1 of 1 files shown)


thomhurst:
Reviewed current head 05f9a14c1aa7845e8aade98b7775589e0bc3e6d8 and all review surfaces. The real prefetch lifetime is established before the unchanged 10 ms poll deadline, and the test verifies cancellation-source identity and state afterward. No outstanding finding remains; the current-head review/CI cycle completed.

Reverified all 1,285 files against the retained SHA-256 inventory at C:/git/Dekaf-evidence/issue-3263/prefetch-lifetime/final-validation. The archived fixture matches the PR head, its working patch is empty, and retained logs report 50 passing cases on each runtime. No tests were rerun for this review. The merge gate reports 26 green checks, MERGEABLE/CLEAN and no unresolved threads. Main remains the PR's base 5ad345184.

PR #3321, opened 2026-09-14
fix(ci): preserve required runtimes during automated updates

thomhurst:
Renovate treated the .NET 8 runtime installations as SDK upgrades and replaced them with .NET 10, even though net8.0 tests and consumers still require .NET 8. Restore those installations and constrain only GitHub Actions entries currently on the 8.x runtime line. SDK updates continue through global.json.

Remove the mutation-report test that duplicates exact runtime and SDK versions from workflow configuration. Mutation report behavior tests remain.

Validation: Python tooling suite passed (366 tests, 16 platform-dependent skips); JSON parsing and git diff --check passed.


thomhurst:
### Edited/Blocked Notification

Renovate will not automatically rebase this PR, because it does not recognize the last commit author and assumes somebody else may have edited the PR.

You can manually request rebase by checking the rebase/retry box above.

 ⚠️ **Warning**: custom changes will be lost.

PR #3346, opened 2026-09-17
fix(producer): restart every partition's sequences lazily under a bumped epoch

### LINKED ISSUE #3342: Idempotent epoch bump leaves unaffected partitions with non-zero sequences the broker rejects under the new epoch
## Summary

`KafkaProducer.BumpEpochForRecoveryAsync` restarts sequence counters only for the partitions that triggered `OutOfOrderSequenceNumber`, `InvalidProducerEpoch` or `UnknownProducerId`. Partitions that did not fail keep their counters and continue them under the new epoch. The broker does not accept that: `ProducerAppendInfo.checkSequence` rejects the first batch of a new producer epoch unless its base sequence is 0 whenever the broker already holds state for that producer ID on the partition (`Invalid sequence number for new epoch of producer ...`). The integration test `IdempotentProducer_EpochSpaceExhausted_ReplacesProducerIdAndKeepsDelivering` added in #3341 relies on exactly this rejection to trigger the producer ID replacement, so the behaviour is confirmed against the Testcontainers broker.

## Effect

Take two active partitions P1 and P2 at epoch N.

1. P1 is rejected. The epoch is bumped to N+1 and only P1's counter restarts at 0.
2. P2's next batch goes out under N+1 with its old non-zero sequence and is rejected with `OutOfOrderSequenceNumber`. The epoch is bumped to N+2 and P2's counter restarts.
3. P1's next batch under N+2 carries a non-zero sequence and is rejected in turn.

Every bump re-stales every other active partition, so one transient error on a multi-partition idempotent producer cascades into a chain of bumps. Each step costs a round trip plus the retry backoff for that partition, and the chain burns through the epoch space of the producer ID. Delivery stays exactly-once and ordered because each rejected batch is re-stamped from the restarted counter in carry-over order, but latency and epoch consumption suffer, and the epoch exhaustion that #3341 handles becomes reachable in practice.

No test covers an epoch bump on a multi-partition topic. `IdempotentProducer_ConcurrentProduction_AllSucceed` uses three partitions but never induces an error.

## Java behaviour

The Java client tracks the producer ID and epoch per partition (`TopicPartitionBookkeeper`). On a bump it re-sequences the triggering partitions from 0 immediately (`startSequencesAtBeginning`). For every other partition, `TransactionManager.maybeUpdateProducerIdAndEpoch` restarts the counter at 0 lazily, the first time the partition drains under the new epoch and has no in-flight batches under the old one (`RecordAccumulator.drainBatchesForOneNode`, `shouldStopDrainBatchesForPartition`). The in-flight gate prevents reordering when an old-epoch retry and a new-epoch batch for the same partition overlap.

## Proposed fix

Extend the per-partition producer state stamp that #3341 added to `RecordAccumulator` (`PartitionSequence.ResetProducerId` / `ResetEpoch`) into the send-time sequence assignment: when a partition's counter was last restarted under an older producer state and the partition has no in-flight batches under that state, restart it at 0 before assigning the batch's sequence. The in-flight check needs the `InflightBatchTracker`, and the partition must stay held (muted) while old-epoch batches are outstanding, as Java does.

Cover it with an integration test that induces `OutOfOrderSequenceNumber` on one partition of a multi-partition topic while another partition keeps producing, asserting a single epoch bump and ordered exactly-once delivery on both partitions.

The comments in `RecordAccumulator.ResetSequencesForPartitions` and `BumpEpochForRecoveryAsync` that claimed the broker carries per-partition sequence state across epoch bumps were corrected in #3341; the behaviour itself is unchanged.

Related: #3341.

thomhurst:
## Summary

Fixes #3342.

`KafkaProducer.BumpEpochForRecoveryAsync` restarted the sequence counters only of the partitions that triggered `OutOfOrderSequenceNumber`, `InvalidProducerEpoch` or `UnknownProducerId`. The broker (`ProducerAppendInfo.checkSequence`) requires sequence 0 for the first batch of a new epoch on every partition it already holds state for, so every other active partition continued its old counter under the new epoch, was rejected with "Invalid sequence number for new epoch", and bumped again. One transient error on a multi-partition idempotent producer cascaded into a bump per active partition, each costing a round trip plus retry backoff, and burned through the epoch space that #3341 now has to recover from.

## Change

Mirrors the Java client (`TransactionManager.maybeUpdateProducerIdAndEpoch` and `RecordAccumulator.shouldStopDrainBatchesForPartition`):

- `RecordAccumulator` keeps, per partition, the producer state (`ProducerIdAndEpoch` reference) under which the counter was last restarted, plus the state the producer published last. `GetAndIncrementSequence(topicPartition, recordCount, state, out restarted)` restarts the counter at 0 the first time a batch is stamped under the published state; a state that is not the published one never restarts anything, so a send loop that still holds the previous snapshot cannot zero sequences already assigned under the current state. `HasStaleSequenceState` answers whether a partition still has to restart.
- `KafkaProducer.PublishIdempotentProducerId` publishes the new state to the accumulator first (a replaced producer ID still clears every counter, since the broker holds no state for the new ID). `BumpEpochForRecoveryAsync` no longer takes partitions and no longer touches any counter; the eager per-partition reset (`ResetSequencesForPartitions`, the sender's `_partitionsNeedingSequenceReset`) is removed, as the lazy restart subsumes it, including for the partition that triggered the bump.
- `BrokerSender` reads one producer-state snapshot per loop iteration (step 4c) and uses it for both coalescing and stamping. When the snapshot changes, it collects the partitions carried by pending requests whose wire stamps name another state, and `CoalesceBatch` holds those partitions (retries included, since Dekaf re-stamps them) until those requests are answered. Sequence 0 of the new epoch therefore never overtakes an old-epoch batch that may still be retried.

## Tests

- Unit: `BrokerSenderEpochRecoveryTests.UnaffectedPartition_RestartsAtZeroUnderBumpedEpoch_WithoutSecondBump` (partition 1 continues after partition 0's bump with sequence 0 and a single bump request) and `UnaffectedPartition_WithRequestPendingUnderOldEpoch_IsHeldUntilItIsAnswered` (the partition 1 batch arriving with the bump is held while partition 1's epoch 5 request is pending, then sent as epoch 6 sequence 0). `EpochBumpRecoveryTests` covers the accumulator contract (restart once per published state, never for an unpublished state, other partitions untouched, full reset on a replaced ID). `IdempotentInitializationTests` updated for the bump that publishes only.
- Integration: `IdempotentProducer_EpochBumpOnOnePartition_OtherPartitionsContinueWithoutAnotherBump` produces to a two-partition topic, induces `OutOfOrderSequenceNumber` on partition 0 through a sequence gap, and asserts exactly one epoch bump plus ordered exactly-once delivery on both partitions. On `main` it fails with `Expected to be equal to 1` (the epoch reaches 2). `IdempotentProducer_EpochSpaceExhausted_ReplacesProducerIdAndKeepsDelivering` now establishes epoch `short.MaxValue` with a lazily restarted sequence 0 and induces the rejection through a sequence gap, since the previous version relied on the bug.
- Full unit suite: 10909 passed (`net10.0`). `IdempotentProducerTests`: 7 passed against the Testcontainers broker.

## Performance

- Per batch (amortized over its records): one extra reference compare in `GetAndIncrementSequence`, and one `bool` read in `CoalesceBatch`. `InflightTrackingBenchmarks.GetAndIncrementSequence_CurrentProducerState` (new, `[MemoryDiagnoser]`) measures the state-aware assignment at 10.9 ns versus 10.6 ns for the stateless overload, 0 B allocated.
- Per send-loop iteration: one delegate call and one reference compare (`RefreshIterationProducerState`).
- Recovery only (from a bump until the requests sent under the previous state are answered): a scan of the pending requests' live batches into a reusable `HashSet<TopicPartition>`, one `HashSet.Contains` and one `ConcurrentDictionary.TryGetValue` per coalesced batch, and the restart itself under the accumulator's restart lock.
- `PendingResponse` is unchanged; the old-state check reads the wire stamps the pending batches already carry.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved idempotent producer recovery after sequence-number errors and producer epoch changes.
  - Prevented duplicate or out-of-order message delivery when multiple partitions are active during recovery.
  - Ensured affected partitions restart sequence numbering safely without unnecessary additional epoch changes.
  - Improved handling of in-flight requests during recovery, preserving message ordering and delivery reliability.
  - Added safeguards for producer ID replacement when the producer epoch is exhausted.


thomhurst:
Stress lane dispatched for the send-loop change (CLAUDE.md loaded-evidence rule): https://github.com/thomhurst/Dekaf/actions/runs/35251290268 — lane `producer-idempotent-1b`, default dispatch shape (`cheap`: 1 sample, no 3conn), `duration_minutes=10`, exact-SHA A1/B/A2 with `baseline_sha=6a886de9e` (fresh `main`, the PR's base after rebase) against candidate `765f80b95`. Verdict to follow.

PR #3354, opened 2026-09-18
fix(ci): use matching runner defaults for NativeAOT

thomhurst:
Use ubuntu-latest and the runner's default clang compiler for NativeAOT. The integration jobs that execute these native binaries already use ubuntu-latest, so the publisher must use the same OS baseline to avoid newer glibc requirements. Remove the redundant Ubuntu and clang-18 pins that break automated runner upgrades.

Remove the wall-clock slow-publish cluster test: it asserts zero duplicates and refused writes despite a three-second lease on a loaded runner. Slow-publish renewal and peer exclusion remain covered by the deterministic OutboxLeaseRenewalTests, with DynamoDB renewal behavior covered by OutboxDynamoDbLeaseTests. The remaining cluster failure/recovery scenarios stay intact.

Validation: all 13 lease-renewal unit tests and all 29 DynamoDB lease integration tests passed locally against Docker; git diff --check passed. Linux NativeAOT publishing is validated by the updated CI jobs.


thomhurst:
### Edited/Blocked Notification

Renovate will not automatically rebase this PR, because it does not recognize the last commit author and assumes somebody else may have edited the PR.

You can manually request rebase by checking the rebase/retry box above.

 ⚠️ **Warning**: custom changes will be lost.

PR #3371, opened 2026-09-21
test: fix the net8.0 DynamoDB outbox harness and a group coordinator flake

thomhurst:
Fixes the two test problems that failed release matrix run https://github.com/thomhurst/Dekaf/actions/runs/35566386338 (5 red legs, all net8.0). Test code only; nothing under `src/` changes.

## 1. Broken: every DynamoDB outbox harness test on net8.0 (34 tests × 4 messaging legs)

```
System.TypeLoadException: Signature of the body and declaration in a method implementation do not match.
  Type: 'generatedProxy_5'. Assembly: 'ProxyBuilder'
   at System.Reflection.DispatchProxy.Create[T,TProxy]()
   at Dekaf.Tests.Integration.OutboxDynamoDbCallInterceptor.Wrap(...)
```

`OutboxDynamoDbCallInterceptor` (added in #3370) was a `DispatchProxy` over `IAmazonDynamoDB`. The AWS SDK's net8.0 build gives `IAmazonService` static abstract members (`CreateDefaultClientConfig`, `CreateDefaultServiceClient`), and the .NET 8 proxy generator emits a type the runtime refuses to load for such an interface. .NET 9 fixed the generator, so net10.0 passed, and PR CI runs integration tests on net10.0 only, which is why #3370 went green.

The interceptor is now a `PipelineHandler`, added outermost to the request pipeline of a client of its own (`DynamoDbLocalContainer.CreateClient(beforeCall)`):

- no runtime code generation, so it behaves the same on every target framework;
- it still sees every operation, and an injected fault still reaches the store as thrown: outermost means outside the SDK's retry handler and its exception event, as the proxy was;
- callbacks match on the request type (`request is QueryRequest`) instead of a method name string;
- the fleet and the cluster count refused conditional writes on both clients, so `RefusedWrites` keeps its meaning.

## 2. Flaky: `GroupCoordinatorCrash_SendOffsetsRetriesThroughStaleMetadataAndCommits` (produce, net8.0, Kafka 4.1.2)

```
System.InvalidOperationException: No candidate group id was coordinated by a broker other than 1.
```

The test needs a group whose coordinator is not the transaction coordinator. Sibling tests SIGKILL brokers, which piles `__consumer_offsets` leadership onto one broker, so the search elected preferred leaders **once**, on the first pass, then waited nine seconds and gave up.

A local probe (kill broker 2, restart, kill broker 3, restart, then search with broker 1 excluded) showed the state the search starts from:

```
partition 0: leader 1, replicas [1,3,2], isr [1,2,3]
partition 1: leader 1, replicas [2,1,3], isr [1,2,3]
partition 2: leader 1, replicas [1,3,2], isr [1,2,3]
```

The preferred replicas name only brokers 1 and 2. With broker 1 to avoid, the one way out is partition 1 going back to broker 2, and a preferred election moves a partition only when that replica is in the ISR. When broker 2 is the broker the previous test has just killed and restarted, it is registered but not yet in sync, the single election moves nothing, and nothing asks again.

The search now elects on every pass that finds no group (up to 60 passes), scoped to `__consumer_offsets` so the transaction coordinator stays where the test found it. If it still gives up, the exception reports each partition's leader, replicas, ISR and election result.

## Verification

Local, Docker, Release:

| Filter | Framework | Result |
| --- | --- | --- |
| `/*/*/OutboxDynamoDb*/*` | net8.0 | 63 passed, 0 failed (was 34 failed in CI) |
| `/*/*/OutboxDynamoDb*/*` | net10.0 | 63 passed, 0 failed |
| `/*/*/TransactionCoordinatorCrashIntegrationTests/*` | net8.0 | 4 passed, 0 failed |

The passing set includes the assertions that prove the interceptor is live rather than a no-op: `faults.Throttled > 0`, `faults.LostAnswers > 0`, the straggling claim landing from `BeforeStoreCall`, and `RefusedWrites == 1` for a refusal raised on the store's client. The probe above found a group on broker 2 one pass after the election, 1.2 s in.

PR CI covers net10.0 only, so a manual non-publish `CI` dispatch on this branch covers the net8.0 integration legs.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved integration-test fault simulation for DynamoDB operations, including more accurate request handling and error timing.
  - Improved reliability when locating consumer groups during broker crash tests by retrying more extensively and providing additional diagnostic details.

- **Tests**
  - Updated DynamoDB integration scenarios to use dedicated clients for store operations and cleaner request interception.
  - Preserved existing fault behaviors while improving test setup and cleanup.


thomhurst:
net8.0 coverage: manual non-publish `CI` dispatch on this branch, https://github.com/thomhurst/Dekaf/actions/runs/35570410369, is green on all 9 `Integration (net8.0)` legs and all 9 `Integration (net10.0)` legs (Kafka 4.3.1). That includes `messaging (net8.0)`, which failed 34 tests in run 35566386338, and `produce (net8.0)`, which holds the group coordinator crash test.

